PR #89, opened 2018-10-11
Refactor/review date

AdamChrimes:
## Description

Refactored the review date component to meet the component criteria, also added examples, documentation and tests for the component. Add visual regression tests for all the components, so we have a baseline before refactoring them. Uses the screen resolutions from our testing policy.

## Breaking change

Review date format has changed from 01/01/2016 to 1 January 2016. Code won't break but this is an important change.

## Component checklist

- [x] SCSS
- [x] SCSS lint
- [x] HTML template
- [x] HTML validate & lint
- [x] Nunjucks macro
- [x] A standalone example
- [x] README/Documentation
- [x] Pseudocode tests
- [x] Visual tests 
- [x] Print stylesheet considered
- [x] CHANGELOG


packages/components/review-date/template.html
```diff
@@ -0,0 +1,6 @@
+<div class="nhsuk-c-review-date">
+  <p class="nhsuk-body-s">
+    Page last reviewed: 12/02/2016<br>
```
mcheung-nhs: @AdamChrimes - What do you think about surrounding the dates in a `<time>` element? For semantics.

AdamChrimes: Resolved

packages/components/review-date/README.md
```diff
@@ -10,8 +10,8 @@ Find out more about the review date component and when to use it in the [NHS dig
 
     <div class="nhsuk-c-review-date">
       <p class="nhsuk-body-s">
-        Page last reviewed: 12/02/2016<br>
-        Next review due: 01/02/2019
+        Page last reviewed: <time>12/02/2016</time><br>
```
mcheung-nhs: @AdamChrimes, @davidhunter08  - Do you think we should go further and add the `datetime` attribute?
`<time datetime="2016-02-12">12/02/2016</time>`?
VoiceOver reads "Page last reviewed twelve slash zero two slash two thousand and sixteen, the twelfth of February two thousand and sixteen"
JAWS reads "Page last reviewed colon twelve slash zero two slash twenty sixteen" (ignores the datetime attribute)
NVDA reads "Page last reviewed twelve slash zero two slash two thousand and sixteen" (ignores the datetime attribute)

AdamChrimes: I was looking at the `datetime` attribute but could get annoying having to write or output the date twice in 2 different formats. I wonder if the date would be better (and more accessible to everyone, not just screen readers) written as '12th January 2016', maybe a question for the standards team to see if there is a standard for date format?

mcheung-nhs: Agreed. Also, I was having an issue with my Windows machine not having the right locale (it was defaulted to American) so when I used the in-built speech reader, it read the date as "2nd December 2016" rather than "12th February 2016" !
I shall speak to the standards team.

Issue #194, opened 2018-11-13
Scss comments being outputted in compiled CSS


AdamChrimes:
Comments intended for use in the scss files only are being outputted in the compiled CSS file. We should use the correct scss comment syntax so these comments are not outputted. 

http://sass-lang.com/documentation/file.SCSS_FOR_SASS_USERS.html#Comments

`//` for scss only comments
`/*` for scss and css comments

Need to decide what comments we do want outputting CSS, if any?

mcheung-nhs:
Changes done and merged. Closing this issue.

PR #195, opened 2018-11-13
[refs #194] Use css comments for component line comments

AdamChrimes:
We were using scss only comments for line comments in components,
these weren't getting outputted when compiling to CSS but are required.


packages/core/utilities/_spacing.scss
```diff
@@ -1,14 +1,10 @@
...
 
-/* Directions for spacing
-   ========================================================================== */
-
-/**
- * Original code taken from GDS (Government Digital Service)
- * https://github.com/alphagov/govuk-frontend
- */
+// Directions for spacing
+//  Original code taken from GDS (Government Digital Service)
```
mcheung-nhs: Extra space

packages/core/styles/_icons.scss
```diff
@@ -75,8 +73,7 @@
   }
 }
 
-/* Icon size adjustments
-   ========================================================================== */
+/* Icon size adjustments */
```
mcheung-nhs: Should this be // like the 2 other comments above for defaults, or should they both be /* like this one?

packages/core/utilities/_spacing.scss
AdamChrimes: Updated!

packages/core/styles/_icons.scss
AdamChrimes: Updated!

PR #230, opened 2018-11-20
Build Documentation with metalsmith

mikemonteith:
## Description

Compile markdown documentation to http://localhost:3000/docs/ with syntax highlighting and breadcrumbs.

### Organisation
I've taken files that are not natively readable on github out of ./docs and into ./site.
The aim is that someone looking at our github repo could read the markdown directly in docs.
./site should contain documentation intended for rendering on the github.io site only. (e.g component examples)

### Metalsmith
Gulp is appropriate for transforming files, but not the right choice for building sites that depend on other files (e.g a breadcrumb, navigation etc.)
I have included a [metalsmith](https://metalsmith.io/) build script to transform the markdown (using marked) and nunjucks into html.

### Relative links
Make sure we're using relative URLs in our docs and site, so that it can be hosted under the github root of `/nhsuk-frontend/`.

### TODO
There is still no link to the http://localhost:300/docs page from the homepage


AdamChrimes:
Closing this PR as we are moving this functionality into the service manual.

PR #249, opened 2018-11-26
Fix care-card template

mikemonteith:
The visually-hidden text was always "Non-urgent advice:" due to
incorrect usage of template parameters.

See here for bug: https://nhsuk.github.io/nhsuk-frontend/components/care-card-emergency.html


mikemonteith:
I'll close this MR then - awaiting the `review/carecards` branch.

PR #258, opened 2018-11-27
Implement ES6 modules JavaScript with webpack

AdamChrimes:
Use modules to expose components individual logic and compile them into a single JavaScript file using webpack and ES6. Refactor emergency banner JavaScript as an example. Still a work in progress, need to refactor the configuration and further components.


mikemonteith:
Are you going to integrate the webpack setup into gulp? (https://pawelgrzybek.com/using-webpack-with-gulpjs/)
Might help keep all of the build-related stuff in one place.

mikemonteith:
`skip-link.js`:
The skip-link code used to be inside a 
```
document.addEventListener('DOMContentLoaded', function() {
...
})
```
and now it's not.

Is that going to be a problem? If someone puts our JS in the <head>, the skipLink selector will fail.

mikemonteith:
`const` is not supported in IE10. https://caniuse.com/#search=const
If we want to start using some more modern JS, maybe have a look at using babel?

AdamChrimes:
@mikemonteith 

`document.addEventListener('DOMContentLoaded', function() {` won't be a problem as we have `if (skipLinkElement && headingElement) {` before trying to add a click event listener to the skip link element.

Added the babel present environments to polyfill our JavaScript for IE10>

AdamChrimes:
Just realised the `details.js` JavaScript was not getting included in the main bundle. This will need rewriting in ES6.

AdamChrimes:
Picking this back up tomorrow, to do:

- Update to the latest `details.js` code in ES6 format
- Tidy up the `typeahead.js` when its compiled into the `nhsuk.min.js` bundle
- Test the bundle

mikemonteith:
> @mikemonteith
> 
> `document.addEventListener('DOMContentLoaded', function() {` won't be a problem as we have `if (skipLinkElement && headingElement) {` before trying to add a click event listener to the skip link element.
> 
> Added the babel present environments to polyfill our JavaScript for IE10>

My point about this is: what if the javascript executes before the skiplink element has been parsed? The skiplink listeners are not going to get attached?

mcheung-nhs:
Have added a `DOMContentLoaded` when initialising components, which will deal with @mikemonteith's concerns above regarding the skiplink element not loaded before the js.

Issue #253, opened 2018-11-27
Change focus styles


mcheung-nhs:
Need to change focus style on links from using box shadow to outline and background colour so links are noticeable when in focus in high contrast mode.

mcheung-nhs:
Focus styles added, closing issue.

PR #294, opened 2018-12-12
[refs #253] Amended focus styles

davidhunter08:
All links now have a 4px transparent outline on `:focus`, so that the `:focus` state is visible for users using high contrast mode


packages/core/tools/_links.scss
```diff
@@ -83,6 +86,8 @@
   &:focus {
     color: $nhsuk-link-focus-color;
     text-decoration: underline;
+    outline: $nhsuk-focus-width solid transparent;
```
mcheung-nhs: @davidhunter08 - Outline stuff needs to come before the text-decoration property as it's causing a linting issue (alphabetical order)

mcheung-nhs: Have fixed.

Issue #96, opened 2018-10-15
Components JavaScript dependencies


AdamChrimes:
Review each of the components required JavaScript, make sure they are compiled in the gulp scripts and update the documentation on how to use them.

- [x] Use IDs instead of class names in JavaScript.
- [x] eslint
- [x] Fix JavaScript comment format, and comment for each component
- [x] Compile into 1 file using Gulp and minify that file.
- [x] Document each components JavaScript and how to include
- [x] Write coding standards for JavaScript
- [x] Header JavaScript: Only try to run the jQuery code if jQuery is included to stop it blocking any other scripts below.
- [ ] Test overall JavaScript

### jQuery Discussion

Remove jQuery? The only component that uses jQuery is the auto complete. GDS have an accessible auto complete component (JavaScript only) that we could use.

https://githubengineering.com/removing-jquery-from-github-frontend/

- No jQuery dependency required in projects.

AdamChrimes:
@sabi98 is currently working on replace our search autocomplete with the GOVUK accessible auto complete. Thus we can delete the jQuery dependency from the project.

PR #304, opened 2018-12-18
[refs #96] Revert event-stream dev dependency version

mcheung-nhs:
Revert to previous version due to security vulnerability

## Checklist

- [ ] SCSS
- [ ] Nunjucks macro
- [ ] A standalone example
- [ ] CHANGELOG entry


AdamChrimes:
I think we may be able to just install it now @mcheung-nhs

PR #329, opened 2019-01-04
GOV.UK autocomplete and removing jQuery dependency

davidhunter08:
## Description

The Header component search autocomplete is now using the GOV.UK accessible autocomplete so that we can remove the Typeahead script which requires jQuery.

This is the only component within the Frontend library that requires jQuery, so we can now move away from using jQuery. This will improve performance as we won't have to load in a (large) third party library.

JavaScript bundle before: 53kb + 87 KB jQuery
JavaScript bundle after: 40kb

## Checklist

- [ ] SCSS
- [ ] Nunjucks macro
- [ ] A standalone example
- [ ] CHANGELOG entry


mikemonteith:
I think I've fixed the git history. I've pushed to mmonteith/autocomplete (diff with master here https://github.com/nhsuk/nhsuk-frontend/compare/mmonteith/autocomplete)
@mcheung-nhs could you take a look at my branch and make sure nothing is missing?

AdamChrimes:
superseded by PR342

PR #342, opened 2019-01-16
Mmonteith/autocomplete

mcheung-nhs:
Cleaned up version of feature/autocomplete branch

## Description

The Header component search autocomplete is now using the GOV.UK accessible autocomplete so that we can remove the Typeahead script which requires jQuery.

This is the only component within the Frontend library that requires jQuery, so we can now move away from using jQuery. This will improve performance as we won't have to load in a (large) third party library.

JavaScript bundle before: 53kb + 87 KB jQuery
JavaScript bundle after: 40kb


packages/components/header/autocomplete.js
```diff
@@ -0,0 +1,121 @@
...
+
+    return svgIcon + resultsHref;
+  }
+
+  const defaultConfig = {
+    element: document.querySelector('#autocomplete-container'),
+    id: id,
+    minLength: 2,
+    placeholder: fallbackInputElement.placeholder,
+    // TO DO:
```
AdamChrimes: @mikemonteith there are a couple of 'TO DO' comments in this JavaScript file, do you know if we need to actually do anything or is it fine without? cc: @sabi98

mikemonteith: I think in this case we should just delete these 3 commented lines.

sabi98: Can remove lines 61 to 63;
  ```
// TO DO:
// cssNamespace: ??
// displayMenu: 'overlay', this need enable but currently breaks.
```

mcheung-nhs: Comments removed,

PR #351, opened 2019-01-17
Use rollup.js instead of webpack to bundle our ES6 JavaScript 

AdamChrimes:
## Description
- remove unused gulp dependencies and add gulp-scss errors
- use rollup.js with Babel to bundle JavaScript
- fix broken inset-text macro from https://github.com/nhsuk/nhsuk-frontend/pull/349 (need to escape `"` in nunjuck macros). Updated all the uses of the macro in the examples, also updated the BackstopJS reference screenshots for these examples.


AdamChrimes:
We're going to pick this up in the future as an enhancement. The current build tools do the job and with the rollup.js IE8 limitations we're going to have to investigate it further. Thanks @nickcolley for the heads up.

Issue #162, opened 2018-11-09
Details


AdamChrimes:
# Details

Use this issue to discuss the **Details** component.

[Preview the details component](https://nhsuk.github.io/nhsuk-frontend/components/details.html)

Find out more about the action link component and when to use it in the [NHS Digital service manual](https://beta.nhs.uk/service-manual/patterns/).

See our [contributing guidelines](https://github.com/nhsuk/nhsuk-frontend/tree/master/docs/contributing)

beseven:
We've used the details component to update the reveal and Q&A wagtail components on the nhsuk-live as we discussed previously. 
As these particular components allow a markdown component, we had to target the styling of multiple HTML elements to make them fit in with the rest of the styling. As this just a small step prior to integrating the FE library with the project, we wanted to share our learning. Please take a look at the PR and let us know if this is of use to you: https://github.com/nhsuk/nhsuk-frontend/pull/363

AdamChrimes:
Closing these individual placeholder issues. 

Please use the [NHS digital service manual backlog](https://github.com/nhsuk/nhsuk-service-manual-backlog/issues) to discuss the usage of this component. If you're having trouble with the code, or have found a bug, please [raise an issue on the NHS.UK frontend backlog](https://github.com/nhsuk/nhsuk-frontend/issues/new)

PR #363, opened 2019-01-28
:nail_care: extra details styling

beseven:
## Description
Proposed changes to the details content text as presented here: https://github.com/nhsuk/nhsuk-frontend/issues/162.
N.B. It's not 100% done, as we experienced some spacing issues with unordered lists, but will give an idea of it's usage.

If the discussion states this is needed, we can take a look at the other bits from the checklist too.

## Checklist

- [x] SCSS
- [ ] Nunjucks macro
- [ ] A standalone example
- [ ] CHANGELOG entry

### DIFF
--- a/packages/components/details/_details.scss
+++ b/packages/components/details/_details.scss
@@ -86,9 +86,15 @@
   padding-left: nhsuk-spacing(4);
   @include top-and-bottom();
 
-  p {
+  h1, h2, h3, h4, h5, h6, p, hr, ul, ol, li, a, figure, img, span, div, table {
     margin-bottom: nhsuk-spacing(4);
-    margin-top: 0;
+    &:first-child {
+      margin-top: 0;
+    }
+    &:last-child {
+      margin-bottom: nhsuk-spacing(2);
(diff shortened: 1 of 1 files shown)


mikemonteith:
Thanks @beseven. I've created an issue for this https://github.com/nhsuk/nhsuk-frontend/issues/364 and I'll bring it to the attention of the team.
Feel free to correct my issue description if I haven't summarised the issue properly.

AdamChrimes:
I'm not sure we should add Wagtail specific fixes to the frontend library, otherwise we could end up with X amount of CMS fixes, X amount of JavaScript framework fixes etc and we will end up with a frontend library full of hotfixes.

So the removal of the margin for the first and last child seems to be a Wagtail specific fix.

However, the ability to support more than just a `<p>` element within a details element is something we should look at. Whether its this implementation or using `*` or `:first-child`. 

@mikemonteith @mcheung-nhs

mcheung-nhs:
@AdamChrimes - agreed. #364 has been created to look at this, so happy for this to be closed.

PR #365, opened 2019-01-30
Backstopjs docker

suryapulakhandam:
## Checklist

- [ ] SCSS
- [ ] Nunjucks macro
- [ ] A standalone example
- [ ] CHANGELOG entry


package.json
```diff
@@ -20,7 +20,11 @@
   },
   "dependencies": {
     "accessible-autocomplete": "^1.6.2",
-    "sass-mq": "^4.0.2"
+    "node-sass": "^4.11.0",
```
AdamChrimes: Howcome we have installed `node-sass`, `npm` and `start` as dependencies? `dependencies` are for production dependencies and `devDependencies` are for development dependencies (so they will get installed within peoples projects when they `npm install nhsuk-frontend`

If these are required only for BackstopJS to run, can we install these only on the Docker container?

package.json
```diff
@@ -48,6 +52,7 @@
     "htmlhint": "^0.10.1",
     "nunjucks": "^3.1.6",
     "sass-lint": "^1.12.1",
+    "start-server-and-test": "^1.7.11",
```
AdamChrimes: Where does this dependency get used? I can't see it run anywhere in the Docker image or within any `package.json` scripts.

AdamChrimes:
@suryapulakhandam it seems the build is not passing because some of the BackstopJS tests are breaking.

package.json
```diff
@@ -20,7 +20,11 @@
   },
   "dependencies": {
     "accessible-autocomplete": "^1.6.2",
-    "sass-mq": "^4.0.2"
+    "node-sass": "^4.11.0",
+    "npm": "^6.7.0",
```
AdamChrimes: Where does this dependency get used? I can't see it run anywhere in the Docker image or within any `package.json` scripts.

There is also a medium vulnerability within this package. https://app.snyk.io/vuln/npm:mem:20180117

package.json
```diff
@@ -20,7 +20,11 @@
...
+    "rebuild": "^0.1.2",
```
AdamChrimes: Where does this dependency get used? I can't see it run anywhere in the Docker image or within any `package.json` scripts.

package.json
suryapulakhandam: removed   "start-server-and-test": "^1.7.11",

package.json
```diff
@@ -20,7 +20,11 @@
...
+    "sass-mq": "^4.0.2",
+    "start": "^5.1.0"
```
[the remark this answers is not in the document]
suryapulakhandam: updated the dependencies

mikemonteith:
These 4 commits aren't very ["atomic"](https://www.freshconsulting.com/atomic-commits/). We should probably look to squash them into 1 change

mikemonteith:
We only need docker for running backstop so maybe we should put all of the backstop-related files (.dockerignore, Dockerfile) inside the `tests/backstop/` directory

mikemonteith:
I wonder if we can get the node app to run inside the travis running rather than inside docker. That way we could remove the slow build step caused by the Dockerfile

suryapulakhandam:
> We only need docker for running backstop so maybe we should put all of the docker-related files (.dockerignore, Dockerfile, docker-compose.yml) inside the `tests/backstop/` directory

I have tried moving the files into backstop directory, but I keeping getting error messages. This reads as   "Can't find a suitable configuration file in this directory or any parent. Are you in the right directory?"

AdamChrimes:
@suryapulakhandam do you need to pass the custom file path in docker-compose? something like:

`docker-compose -f path/to/docker-compose-tests.yml`

AdamChrimes:
This has been picked up in the following PR https://github.com/nhsuk/nhsuk-frontend/pull/390

Issue #398, opened 2019-02-20
Toggle Menu not working in Angular


puravupadhyay:
# Bug Report

## What is the issue?

I'm trying to use this library in an Angular project. I installed nhsuk-frontend via npm fine.
I managed to import the styles and can see all the component rendering but I'm having problems with the javascript files. I imported all the modules like suggested but not able to make the Menu Toggle work. Can someone please help?

## What steps are required to reproduce the issue?
1. Create an Angular app, install nhsuk-frontend using NPM.
2. Include the NHS nav header in the app.component.html
3. Import the javascript modules in the app.component.ts
4. Run the app, keep reducing the screen size until the Menu button appears. Click on it and nothing happens.

## What was the environment where this issue occurred?

* Device: PC / MAC
* Operating System: Windows 10, macOS
* Browser: All
* Browser version: -
* NHS.UK frontend package version: 1.0.0
* Node version: 11.10.0
* npm version: 6.7.0

## Is there anything else you think would be useful in recreating the issue?

This is the sample code of app.component.ts file:

``` import { Component } from '@angular/core';
import nhsuk_feedbackBanner from 'node_modules/nhsuk-frontend/packages/components/feedback-banner/feedback-banner';
import nhsuk_header from 'node_modules/nhsuk-frontend/packages/components/header/header';
import nhsuk_skipLink from 'node_modules/nhsuk-frontend/packages/components/skip-link/skip-link';
import autocomplete from 'node_modules/nhsuk-frontend/packages/components/header/autocomplete';

@Component({
  selector: 'app-root',
  templateUrl: './app.component.html',
  styleUrls: ['./app.component.scss']
})
export class AppComponent {

  title = 'NHSFrontEnd';


  ngOnInit() {
    nhsuk_feedbackBanner(3000);
    nhsuk_header();
    nhsuk_skipLink();
    autocomplete();
  }

}
```
<img width="753" alt="nhs header" src="https://user-images.githubusercontent.com/17790271/53112886-3418ec00-3538-11e9-9c40-e3222a9dd49f.png">

AdamChrimes:
Are there any errors in the browser console?

puravupadhyay:
No errors in the console

<img width="930" alt="console" src="https://user-images.githubusercontent.com/17790271/53113769-409e4400-353a-11e9-9b2b-ccd46dd59ad1.png">

AdamChrimes:
[...]

PR #402, opened 2019-02-22
JavaScript Variable Scope and Hoisting

AdamChrimes:
## Description

This is a potential fix for https://github.com/nhsuk/nhsuk-frontend/issues/398

Globally scoped variables were causing issues with JavaScript frameworks such as Angular, so they have been moved to be locally scoped within functions. However, this could be improved as we have to duplicate variables between the multiple functions.

There was also an issue with the autocomplete load event listener with Angular, The wrap-search element was not being found as the page load event listener was coming after the search for the element which caused issues in JavaScript frameworks such as Angular.

Currently testing this out with Angular and ES5/Webpack, also looking for better ways to do this and potentially look for advice on how we write our JavaScript (ES6).

### Further reading

https://www.sitepoint.com/demystifying-javascript-variable-scope-hoisting/

### To Do:

- [x] Tested in Angular using the [NhsukFrontendAngular](https://github.com/puravupadhyay/nhsuk-frontend-angular) repo provided by @puravupadhyay
- [x] Tested in Browsers (to IE8) with ES5 (using the nhsuk-frontend repo pages)
- [ ] Tested in third party application (non JS framework) which uses the `nhsuk-frontend` library as a npm package
- [ ] Tested in third party application (non JS framework) which uses the `nhsuk-frontend` library from compiled files
- [ ] Investigate better ways to write this JavaScript to avoid duplication of variables, also is there a better way to check the element exists eg `if (element)` or an alternative not to have to do this?

The above also needs to be tested by someone else to confirm.

### DIFF
--- a/packages/components/header/header.js
+++ b/packages/components/header/header.js
@@ -27,7 +27,6 @@ function toggleSearch(e) {
   const menuSearchContainer = document.querySelector('#content-header');
 
   e.preventDefault();
-  e.stopImmediatePropagation();
 
   if (searchToggleButton.hasAttribute('aria-expanded')) {
     searchToggleButton.removeAttribute('aria-expanded');
@@ -59,7 +58,6 @@ function toggleMenu(e) {
   const nav = document.querySelector('#header-navigation');
 
   e.preventDefault();
-  e.stopImmediatePropagation();
 
   if (menuToggleButton.hasAttribute('aria-expanded')) {
     menuToggleButton.removeAttribute('aria-expanded');
--- a/packages/components/header/header.js
+++ b/packages/components/header/header.js
@@ -21,25 +21,29 @@ function toggleClass(ele, class1) {
 }
 
 /* Search toggle */
-
-const searchToggleButton = document.querySelector('#toggle-search');
-const searchClose = document.querySelector('#close-search');
-const searchContainer = document.querySelector('#wrap-search');
-const menuSearchContainer = document.querySelector('#content-header');
-
 function toggleSearch(e) {
+  const searchToggleButton = document.querySelector('#toggle-search');
+  const searchContainer = document.querySelector('#wrap-search');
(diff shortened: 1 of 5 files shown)


packages/components/header/header.js
```diff
@@ -21,25 +21,29 @@ function toggleClass(ele, class1) {
...
-const searchContainer = document.querySelector('#wrap-search');
-const menuSearchContainer = document.querySelector('#content-header');
-
 function toggleSearch(e) {
+  const searchToggleButton = document.querySelector('#toggle-search');
+  const searchContainer = document.querySelector('#wrap-search');
+  const menuSearchContainer = document.querySelector('#content-header');
+
   e.preventDefault();
+  e.stopImmediatePropagation();
```
mikemonteith: Need to take these out (1)

packages/components/header/header.js
```diff
@@ -50,22 +54,27 @@ function handleSearchToggle() {
...
-const menuToggleButton = document.querySelector('#toggle-menu');
-const menuClose = document.querySelector('#close-menu');
-const nav = document.querySelector('#header-navigation');
-
 function toggleMenu(e) {
+  const menuToggleButton = document.querySelector('#toggle-menu');
+  const nav = document.querySelector('#header-navigation');
+
   e.preventDefault();
+  e.stopImmediatePropagation();
```
mikemonteith: (2)

AdamChrimes:
@mikemonteith removed the stopImmediatePropogation calls.

package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "nhsuk-frontend",
-  "version": "1.0.0",
+  "version": "1.1.0",
```
mikemonteith: Does this need to be a minor version increase? It's a bugfix rather than new feature so I'd make it a patch update.

Issue #384, opened 2019-02-06
Create a summary list component


GrilloPress:
## What is it?

The [Summary list component](https://design-system.service.gov.uk/components/summary-list/) is an established GOV.UK component in use in various GOV services as well as live in the [National data opt-out service](https://your-data-matters.service.nhs.uk) service.

It is also in use in an alpha service for the "Check your personal details" part of the audit and transparency foundation service/s.

The summary list component is a key element of the [GDS pattern to help users to "check answers"](https://design-system.service.gov.uk/patterns/check-answers/).

## What problem does this thing solve?

Many transactional services need to have users check data before submission, or check data against a question or criteria, and the summary list is an established component for doing this.

Having the summary list component and related macro would:

- speed up development
- ensure services don't have to recreate their own versions/interpretations of this component
- make it easier for services to build key/value `<dl>` lists
- make it easier to make `<dl>` lists accessible and well structured

## Where has this been tested

Across gov. This component has been tested across numerous GOV services and internal case working systems.

In the NHS this has been tested in the national data opt-out service:

- with 20+ users in various labs
- as part of remote unmoderated A/B testing with 50 users
- as part of the live service since October 2018 (thousands of users etc.)

It has also been tested in the new frontend in the "Check your personal details" service:

- with 2 users in a pop-up
- 5 users in a lab

*Issues reported:* None

## Other links

For the audit and transparency set of services I have created a proof of concept fork of the GDS component:

- [scss for the component](https://github.com/GrilloPress/audit/blob/master/public/stylesheets/prototype/_summary-list.scss)
- [macro for the component](https://github.com/GrilloPress/audit/blob/master/views/_macros/_summary-list.html)
- [list in example page](https://nhs-audit.herokuapp.com/examples/check-details)
- [list in alpha service](https://nhs-audit.herokuapp.com/view-your-pds/v1/your-details/1-inline)

AdamChrimes:
Looks like we having something along the lines of the Summary list component within the [NHS.UK prototype kit - Check your answers page](https://nhsuk-prototype-kit.azurewebsites.net/templates/check-your-answers). 

Do we just need to move from the prototype kit into the frontend @davidhunter08?

mcheung-nhs:
Done and merged into master - https://github.com/nhsuk/nhsuk-frontend/pull/404

PR #404, opened 2019-02-25
[refs #384] Create summary list component from GOV.UK

AdamChrimes:
## Description

Summary list component styles, markup, macro and documentation taken
from the GOV.UK Frontend.

### To do
- [x] Design sign off (@davidhunter08)
- [x] BackstopJS tests
- [x] Screen resolution, browser and accessibility testing


AdamChrimes:
@davidhunter08 would you be able to have a look at the design (spacing etc) and let me know if you're happy with it.

davidhunter08:
@AdamChrimes - Design (spacing etc.) looks ok.

AdamChrimes:
Thanks @davidhunter08, @mcheung-nhs please could you have a look when you are back. 🌤

GrilloPress:
Looks good generally. Big fan of the links going to the left-hand margin on mobile.

Two thoughts:

1.  Word break attribute seems to be causing some issues with the contact information label
image

2. Wrapping `<dl>` in `<div>`

Just working through some issues with GDS about this. Technically it may not be valid HTML to wrap the `<dt>` and `<dd>` inside a `<div>`.

Axe interprets the rule as a strict `<dt>` must have a parent of `<dl>`

I'm interpreting https://www.w3.org/TR/html5/grouping-content.html#the-dl-element as meaning that we can have `<div>` children of `<dl>` but waiting for GDS to get back to me.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl#Wrapping_name-value_groups_in_%3Cdiv%3E_elements

GrilloPress:
Dave House from GDS confirmed my interpretation of the HTML spec. From GDS point of view.

So wrapping`<div>` is fine. But will trigger some accessibility tools to raise an error. DAC also raised it to us but I'm challenging them on it.

Something to be aware of.

joelanman:
Related work in Axe: https://github.com/dequelabs/axe-core/pull/1284 (unreleased I think)

mcheung-nhs:
I would also interpret the W3 specs mentioned that a `<div>` is allowed in a `<dl>`.
Also, running the code through the W3 HTML5 validator doesn't bring up any issues.

Looking at further discussions with the axe-core library (https://github.com/dequelabs/axe-core/issues/262) it seems there are no issues from a screen reader point of view.

Therefore I'm happy for this to be merged and we'll have to put any exceptions if/when we implement axe in our CI.

PR #405, opened 2019-02-25
Stop compiling markdown files to HTML within the docs

AdamChrimes:
## Description

- We explored the idea of compiling markdown files, such as the README.md, into HTML files so that they could be viewed within the browser. We're not really using the READMEs in HTML and the service manual is being used for documentation instead of the GitHub pages. 

- Remove any files or gulp packages related to the compiling of markdown files.

- The paths for pages in the `lint:html` script were incorrect, so the HTML linting wasn't actually running in the pipeline; the paths have been fixed. The HTML linter also fails on inline page styles `<style></style>` so a `<section>` has been used to change the background of the reverse button page background colour.


GrilloPress:
Looks good generally. Big fan of the links going to the left-hand margin on mobile.

Two thoughts:

1.  Word break attribute seems to be causing some issues with the contact information label
image

2. Wrapping `<dl>` in `<div>`

Just working through some issues with GDS about this. Technically it may not be valid HTML to wrap the `<dt>` and `<dd>` inside a `<div>`.

Axe interprets the rule as a strict `<dt>` must have a parent of `<dl>`

I'm interpreting https://www.w3.org/TR/html5/grouping-content.html#the-dl-element as meaning that we can have `<div>` children of `<dl>` but waiting for GDS to get back to me.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl#Wrapping_name-value_groups_in_%3Cdiv%3E_elements

PR #410, opened 2019-03-01
Made distinction between heading and caption

markasrobinson:
## Description

Changed heading to "Conditions similar to impetigo", caption to "Other possible causes of your symptoms" and first column to "Symptoms".

Previously, screenreaders will have read out the same line "Skin symptoms and possible causes" twice (as heading and caption).

### DIFF
--- a/app/components/tables/tables-panel.njk
+++ b/app/components/tables/tables-panel.njk
@@ -12,12 +12,12 @@
         <div class="nhsuk-grid-column-two-thirds">
           {{ table({
             panel: true,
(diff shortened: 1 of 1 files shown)


AdamChrimes:
Hi @markasrobinson looks good.

Would you be able to update the relevant examples in the `packages/components/tables/README.md`, there should be an HTML snippet and the nunjucks macro again in the documentation for the [Table panel variant](https://github.com/nhsuk/nhsuk-frontend/tree/master/packages/components/tables#table-panel).

Give me a shout if you need any help.

Issue #416, opened 2019-03-14
Full width grid container


AdamChrimes:
We've seen a need for new designs to support a full-width page design, spanning the entire width of the viewport. For example to fit content such as large data tables on the page without restricting it to a 'letterbox' design.

The current `nhsuk-width-container` is restricted to a max width of 960px. This could be modified to allow the 960px to be customisable via variables or an additional `nhsuk-width-container` class to give a full width layout with the relevant spacing and guttering. Similar to `container` and `container-fluid` classes used in the [Bootstrap framework](https://getbootstrap.com/docs/4.3/layout/overview/#containers).

AdamChrimes:
The documentation for the Layout/Grid will also need information adding about the full width container and the service manual documentation for Grid could do with a bit of re-writing, as there is no logical order to the page for introducing the parts of the layout and also some classes and content is wrong. https://beta.nhs.uk/service-manual/styles-components-patterns/grid

AdamChrimes:
Released in version `2.1.0`.

PR #418, opened 2019-03-18
[refs #416] Extend the page layout to include a fluid-width container.

AdamChrimes:
## Description

The current nhsuk-width-container is restricted to a max width of 960px. A new container has been added (nhsuk-width-container-fluid), which which spans the entire width of the viewport and allows for full-screen layouts.

**To Do:**
Just in process of writing up the documentation for the service manual, should be ready before we release 2.10.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # NHS.UK frontend Changelog
 
+## 2.1.0 - TBC, 2019
+
+:new: **New features**
+
+- Fluid width container - Extend the page layout to include a fluid-width container, which spans the entire width of the viewport.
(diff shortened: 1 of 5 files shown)


AdamChrimes:
@davidhunter08 can you check the spacing of the guttering? It's 16px same as the guttering on the normal container, but not sure if its too tight to the edge.

mcheung-nhs:
Am happy with code. Personally I'd prefer 32px guttering when going full width, but happy to hear what @davidhunter08 thinks.

davidhunter08:
I agree with @mcheung-nhs, I think it should be a 32px guttering for `nhsuk-width-container-fluid` from the desktop breakpoint (990px).

davidhunter08:
Looking good, @AdamChrimes

PR #419, opened 2019-03-19
Prefix error messages with hidden "Error:"

AdamChrimes:
## Description

Prefix error messages with a visually hidden "Error:", to make it clearer to users of assistive technologies.

### DIFF
--- a/packages/components/error-message/template.njk
+++ b/packages/components/error-message/template.njk
@@ -1,6 +1,9 @@
+{% set visuallyHiddenText = params.visuallyHiddenText | default("Error") -%}
+
 <span class="nhsuk-error-message
 {%- if params.classes %} {{ params.classes }}{% endif %}"
 {%- if params.id %} id="{{ params.id }}"{% endif %}
 {%- for attribute, value in params.attributes %} {{ attribute }}="{{ value }}"{% endfor %}>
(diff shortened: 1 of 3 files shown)


packages/components/error-message/template.njk
```diff
@@ -1,6 +1,9 @@
+{% set visuallyHiddenText = params.visuallyHiddenText | default("Error") -%}
+
 <span class="nhsuk-error-message
 {%- if params.classes %} {{ params.classes }}{% endif %}"
 {%- if params.id %} id="{{ params.id }}"{% endif %}
 {%- for attribute, value in params.attributes %} {{ attribute }}="{{ value }}"{% endfor %}>
+  {% if visuallyHiddenText %}<span class="nhsuk-u-visually-hidden">{{ visuallyHiddenText }}:</span> {% endif -%}
```
mcheung-nhs: @AdamChrimes - Do you think we should hardcode the ":" or allow/rely on the user to put any punctuation they want?

AdamChrimes: @mcheung-nhs I have a feeling people will miss it out and will end up with no punctuation. Can you think of an alternative punctuation people will use than `:`?

mcheung-nhs: @AdamChrimes - fair enough. Just don't want people to put in any extra punctuation at the end of the text so having a colon there doesn't make sense, although this is probably highly unlikely!

PR #441, opened 2019-04-26
[refs #432] Full width of hero on small screens

### LINKED ISSUE #432: Hero image with image and content, longer words escaping container
When on a smaller mobile device with text enlarged, longer words escape the blue container.

Screenshot: 
Screenshot_2019-04-03-12-10-38

A potential fix could be to make the blue container fill the full-screen width on mobile devices. However, this will need to be tested further

Screenshot:
Screenshot_2019-04-03-15-16-51

adrhumphreys:
## Description
Fixes #432 with the idea that full width makes the most sense only on small devices. This should work for words such as "transformation".

This only effects devices under 375px which are devices small than e.g. an iPhone 6 such as the iPhone 5/SE

This omits the idea of using `overflow-wrap: break-word;` as it doesn't clearly break words which could lead to confusion for users. For example you could have the following with it:
```
Driving forward the digital transformati
on of health and care.
```

## Checklist

- [x] SCSS
- [ ] Nunjucks macro
- [ ] A standalone example
- [x] Visual tests ([BackstopJS](https://github.com/nhsuk/nhsuk-frontend/blob/master/docs/contributing/automated-testing.md))
- [ ] CHANGELOG entry


AdamChrimes:
Hi @adrhumphreys thanks for having a look at this.

We try to keep the number of breakpoints we have to a minimum and try to avoid having custom units that aren't mapped to our sass-mq breakpoint variables.

```
$mq-breakpoints: (
  mobile:  320px,
  tablet:  641px,
  desktop: 769px,
  large-desktop: 990px
);
```

Would this work using the `mobile` `320px` breakpoint?

adrhumphreys:
If we were to go with the `320px` then we would end up losing the partial banner background for all mobile devices where as this solution allows us to keep how it looks for most devices (iPhone 6 screen size and up) but only changes how it looks for phones smaller than that size.

TLDR: We could but it would come at the cost of making the mobile background image redundant. 

Thoughts?

AdamChrimes:
Yeah I see what you mean - let me speak to our designers and I'll get back to you. Sorry for the delayed response, we've had lack of resources recently (summer holidays, other commitments etc.

adrhumphreys:
Just wanted to point out that this is probably not a great experience for people looking to engage with y'all and contribute

davidhunter08:
Apologies for this @adrhumphreys. We recently cleaned up all our issues and PR's. This PR must of been overlooked. I'll reopen it and review it this week.

davidhunter08:
Hi @adrhumphreys - the fix looks good.

Could you please rebase with master and add an entry into the `CHANGELOG.md`?

We're also planning to review the breakpoints for mobile - so that we can address @AdamChrimes 's above concerns.

AdamChrimes:
Just needs rebasing with master to fix merge conflicts and this can be released as part of 2.3.1

AdamChrimes:
Hey @adrhumphreys would you be able to rebase this with master with a change log entry added and we can include this in the next release?

AdamChrimes:
Due to this being a fork, I couldn't update the branch easily. I've made a new pull request https://github.com/nhsuk/nhsuk-frontend/pull/549 crediting you as Co-Author so you still get the commit 👍 

Thanks for looking at this and apologies again for the delay and confusion.

PR #442, opened 2019-04-30
Expander group spacing

### LINKED ISSUE #439: Expander group spacing
When using the expander group component above another component, with no top margin - such as details or action link, the expander group is very tight to that component due to no margin bottom on the expander group.

**Example:** 
Screenshot 2019-04-16 at 10 25 04
Screenshot 2019-04-16 at 10 25 20

Another example is when you use an expander group, followed by a single expander, visually it looks like all 3 are in a single group due to the spacing.

The 3rd expander in the example is actually a single expander. The first 2 are in a group.

**Example:**
Screenshot 2019-04-16 at 10 26 40

xLasercut:
## Description

resolves #439 

## Checklist

- [x] SCSS
- [ ] Nunjucks macro
- [ ] A standalone example
- [x] Visual tests ([BackstopJS](https://github.com/nhsuk/nhsuk-frontend/blob/master/docs/contributing/automated-testing.md))
- [x] CHANGELOG entry


AdamChrimes:
Hi @xLasercut, we have a responsive spacing Sasss mixin which is mapped to our defined spacing scale. We try to avoid hard coding units for spacing, there also needs to be more spacing on larger screens.

`@include nhsuk-responsive-margin(4, 'bottom');` would give us a margin-bottom of 16px on smaller screens and 24px on larger screens. You can read more about how the spacing works on the [NHS digital service manual spacing page](https://beta.nhs.uk/service-manual/styles-components-patterns/spacing)

We may also be able to just add the margin to the main `.nhsuk-expander-group` container, which wraps around the group of expanders. This would reduce the specificity a little bit.

xLasercut:
changed scss to use the nhsuk-responsive-margin  and moved the margin to the expander group as suggested

AdamChrimes:
Cheers @xLasercut looks good, just looping in @davidhunter08 that the spacing size is correct from a design point of view.

In the meantime, could you add a `CHANGELOG` entry. The version will be `2.1.1` with the date.

PR #444, opened 2019-05-03
fixes backstop test not working on linux or windows

xLasercut:
## Checklist

- [ ] SCSS
- [ ] Nunjucks macro
- [ ] A standalone example
- [ ] Visual tests ([BackstopJS](https://github.com/nhsuk/nhsuk-frontend/blob/master/docs/contributing/automated-testing.md))
- [ ] CHANGELOG entry

Fixes backstop tests failing on linux and windows by adding a script to run `/sbin/ip route` inside the container and retrieve the docker bridge IP.

### DIFF
--- a/tests/backstop/backstop.js
+++ b/tests/backstop/backstop.js
@@ -1,6 +1,26 @@
 // Get the --testhost=... argument from the backstop command
 const arguments = require('minimist')(process.argv.slice(2));
-const TEST_HOST = arguments.testhost || "http://host.docker.internal:3000"
+var TEST_HOST = arguments.testhost || "http://host.docker.internal:3000"
+
+const childProcess = require('child_process')
+var output = ""
+try {
+  output = childProcess.execFileSync("/sbin/ip", ["route"], {encoding: "utf-8"})
+}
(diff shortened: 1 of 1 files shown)


AdamChrimes:
@mikemonteith are you OK to have a look at this? Our Travis CI pipeline is Linux so also thought it was working there and I'm sure we had it running on Suryas machine (Windows)

mikemonteith:
@xLasercut can you explain this a bit?
What command are you trying to run?
I think we added the `testhost` argument so that you could pass the correct URL for your specific environment. If we have some code to figure out the ip address, it will need to be tested on a range of hosts.

xLasercut:
@mikemonteith 
The command is `/sbin/ip route`

which will give us the docker bridge ip in the line with:
default via <IP> dev eth0

This will be run inside the container (which will always be linux). https://stackoverflow.com/questions/22944631/how-to-get-the-ip-address-of-the-docker-host-from-inside-a-docker-container

AdamChrimes:
There is no straight forward fix for this, so we're going to close this PR and investigate it further.

PR #460, opened 2019-05-31
Remove light version of the font

mcheung-nhs:
The Frutiger light font was only being referenced in the header
autocomplete dropdown, so we've decided to remove it and save a request
to get it.

## Checklist

- [ ] SCSS
- [ ] Nunjucks macro
- [ ] A standalone example
- [ ] Visual tests ([BackstopJS](https://github.com/nhsuk/nhsuk-frontend/blob/master/docs/contributing/automated-testing.md))
- [x] CHANGELOG entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 # NHS.UK frontend Changelog
 
-## 2.1.1 - Unreleased
+## 2.2.0 - Unreleased
+
+:boom: **Breaking changes**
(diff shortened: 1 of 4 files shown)


AdamChrimes:
Wary of doing a breaking change - are we able to set $nhsuk-font-light to 400 instead of removing it all together? If its a breaking change the first number needs to change so it will be `3.0.0`

PR #510, opened 2019-08-05
add copyright override parameter and docs

### LINKED ISSUE #485: Allow override for copyright in footer macro
Reviewing some work with @samanthasaw and we noticed you couldn't override the copyright notice in the footer macro.

Not every service will be crown copyright. Either at all, or may have more complicated copyright issues.

Probably should be able to either toggle the visibility and use a different copyright

sabi98:
## Description
Resolves #485

### DIFF
--- a/packages/components/footer/template.njk
+++ b/packages/components/footer/template.njk
@@ -11,7 +11,7 @@
         {%- endfor %}
       </ul>
       {% endif %}
-      <p class="nhsuk-footer__copyright">&copy; Crown copyright</p>
+      <p class="nhsuk-footer__copyright">{% if params.copyright %}{{ params.copyright | safe }}{% else %}&copy; Crown copyright{% endif %}</p>
     </div>
   </div>
 </footer>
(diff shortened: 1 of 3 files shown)


AdamChrimes:
Thanks @sabi98, mostly this looks OK. However, I can't use the Copyright symbol HTML entity (&copy;) within the macro because HTML is escaped. You can use the `| safe` pipe within nunjucks to output HTML, but we may need to investigate if there's any potential issues with doing this.

`{{ params.copyright | safe }}`

AdamChrimes:
Screenshot 2019-08-07 at 10 38 22

AdamChrimes:
I've fixed an issue with Travis build failing in https://github.com/nhsuk/nhsuk-frontend/pull/512 if you can rebase this branch with `master` it should start passing again.

packages/components/footer/template.njk
```diff
@@ -11,7 +11,7 @@
         {%- endfor %}
       </ul>
       {% endif %}
-      <p class="nhsuk-footer__copyright">&copy; Crown copyright</p>
+      <p class="nhsuk-footer__copyright">{% if params.copyright %}{{ params.copyright | safe}}{% else %}&copy; Crown copyright{% endif %}</p>
```
AdamChrimes: Needs a space after the word `safe` 🗡

PR #518, opened 2019-08-23
add eslint and sass lint configuration path for code editors

AdamChrimes:
## Description

Code editors weren't automatically picking up the configuration file for the eslint, added to package.json so that they know where to look.


AdamChrimes:
I've added the sass lint configuration too, @tomdoughty

PR #526, opened 2019-09-05
[refs #525] Hide search results downdown on page scroll

### LINKED ISSUE #525: Header search results dropdown doesn't disappear when you scroll down the page.
# Bug Report

## What is the issue?

When scrolling down with search results dropdown the expected behaviour is that it should disappear. However, this doesn't happen and it stays open regardless of how much you scroll. Only disappears by clicking.

* Device:
* Operating System:
* Browser:
* Browser version:
* NHS.UK frontend package version:
* Node version:
* npm version:

sabi98:
## Description

Resolves #525

## Checklist

- [x] CHANGELOG entry


AdamChrimes:
Need to check the behavior of this on mobile devices. The dropdown disappears on scroll but the search box remains open and you have clear the text input field before you can get the autocomplete again.

AdamChrimes:
Spoke to @sabi98 and we've decided that is a non-issue, the behavior should be left as it is to avoid causing complications on different devices/assistive technology.

PR #548, opened 2019-10-21
Update package-lock to build on new Node LTS

tomdoughty:
## Description
Create new package-lock.json using Node.js v12.13.0 as it is the LTS release as of 21/10/2019.
Without this fix the package doesn't install the node-sass binaries when using Node v12.x

I will bump my changelog and versions after discussing the planned order of release with other PRs.

## Checklist

- [ ] CHANGELOG entry

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3221,7 +3221,7 @@
     },
     "es6-promise": {
       "version": "3.3.1",
-      "resolved": "https://registry.npmjs.org/es6-promise/-/es6-promise-3.3.1.tgz",
+      "resolved": "http://registry.npmjs.org/es6-promise/-/es6-promise-3.3.1.tgz",
       "integrity": "sha1-oIzd6EzNvzTQJ6FFG8kdS80ophM=",
       "dev": true
     },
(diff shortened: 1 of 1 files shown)


AdamChrimes:
This can go in whenever, as it doesn't affect the package - just the development environment.

Will the `package-lock` changes get overwritten next time? Do we need to install `node-sass` as a devDependency until `gulp-sass` updates to the latest version?

AdamChrimes:
https://github.com/dlmanning/gulp-sass/pull/748

tomdoughty:
@AdamChrimes It will only be an issue if a package-lock.json generated using lower that Node 12.x gets committed. We should be making sure when we do reviews that this doesn't happen.

AdamChrimes:
Good shout @tomdoughty I probably should update my Node to 12.x!

AdamChrimes:
Will it be the same fix for the prototype kit? https://github.com/nhsuk/nhsuk-prototype-kit

tomdoughty:
Looks that way... it is currently locked to node-sass 4.11.0, while the minimum to work with Node 12 is 4.12.x

PR #555, opened 2019-11-01
Focus styles update

davidhunter08:
## Description

Update active state style for the main nav links. It was conflicting with the new focus styles.

Update hover state style for details and expander component. It now uses the new link hover colour.


AdamChrimes:
@davidhunter08 please could you rebase this with `develop`. There are also some Backstop tests that need updating.

AdamChrimes:
I think we may just need to look at the Back link hover state, the chevron changes to the correct hover colour but the text doesn't.

PR #587, opened 2020-02-17
fix issue with Travis CI by opting into v2 deployment tooling

AdamChrimes:
## Description

Travis CI was having issues deploying to npm due to recently version changes in tooling on their end. This change opts in to using the latest version and relevant configuration for the new token name for API keys.

https://docs.travis-ci.com/user/deployment-v2
https://travis-ci.community/t/missing-api-key-when-deploying-to-github-releases/5761/14

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,7 +49,8 @@ jobs:
       script: npm run build-gh-release
       deploy:
         provider: releases
-        api_key: $GITHUB_TOKEN
+        edge: true
+        api_token: $GITHUB_TOKEN
         file_glob: true
         file: ./dist/nhsuk-frontend-*.zip
         skip_cleanup: true
@@ -60,6 +61,7 @@ jobs:
       name: publish to npm
       deploy:
         provider: npm
+        edge: true
         email: $NPM_EMAIL
         api_key: $NPM_API_KEY
         on:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -49,7 +49,8 @@ jobs:
       script: npm run build-gh-release
       deploy:
         provider: releases
-        api_key: $GITHUB_TOKEN
+        edge: true
+        api_token: $GITHUB_TOKEN
```
st3v3nhunt: I think `api_token` needs to be just `token` https://docs.travis-ci.com/user/deployment-v2/providers/releases/#known-options for the `releases` provider

AdamChrimes: Good spot, changed that to `token`

.travis.yml
```diff
@@ -60,8 +61,9 @@ jobs:
       name: publish to npm
       deploy:
         provider: npm
+        edge: true
         email: $NPM_EMAIL
-        api_key: $NPM_API_KEY
+        token: $NPM_API_KEY
```
st3v3nhunt: The npm provider needs `api_token` - https://docs.travis-ci.com/user/deployment-v2/providers/npm/#known-options
You think they would be the same across all providers but apparently not!

AdamChrimes: That would make too much sense, updated.

PR #607, opened 2020-05-01
Responsive accessible table

Fenwick17:
## Description
Component for a responsive and accessible table

## How it works
When viewed on a destkop the table component will behave like any other table. However, when viewed on a mobile the table collapses into what appears to be a group list style component.
The markup remains the same as it did originally, so when using a screen reader it will still be read out as if it was a normal table just with a different visual representation.
To do this the markup includes numerous extra attributes and elements paired with CSS to modify the layout of the table. 

- CSS is used to modify the layout to use flex and block
- Roles are added to all table elements which will then mean screenreader behaviour remains the same and despite having a different visual representation, it will still behave as a normal table would
- `aria-hidden` is used to provide visual context of the heading which is no longer visible, this will be the same content used in the appropriate table heading

## When to use
This should only be used to show tabular data when your table has 3 columns or more. 

Further documentation to follow.

## Checklist

- [X] CHANGELOG entry


packages/components/tables/_tables.scss
```diff
@@ -71,3 +71,82 @@
...
+  font-weight: 700;
+  padding-right: 1em;
+  text-align: left;
+
+  @media (min-width: $mobile) {
+    display: none;
+  }
+}
+
+.table-foot {
```
AdamChrimes, line 145: Does this need to be `.nhsuk-table__footer`

packages/components/tables/_tables.scss
```diff
@@ -71,3 +71,82 @@
 .nhsuk-table__cell--numeric {
   text-align: right;
 }
+
+$mobile: 768px;
```
AdamChrimes, line 75: To avoid added another breakpoint could this not use the `desktop` breakpoint which is 769px?

Fenwick17:
Closing due to https://github.com/nhsuk/nhsuk-frontend/pull/635

PR #633, opened 2020-09-22
Update warning callout template.njk

davidhunter08:
## Description
Change the `{%- if params.heading != "Important" %}` to `{%- if "Important" in params.heading %}` so the hidden text of 'Important' does not show if a custom header includes the word important, for example 'Important Coronavirus update'

For example:

image

### DIFF
--- a/packages/components/warning-callout/template.njk
+++ b/packages/components/warning-callout/template.njk
@@ -3,7 +3,7 @@
 {%- if params.classes %} {{ params.classes }}{% endif %}"
 {%- for attribute, value in params.attributes %} {{attribute}}="{{value}}"{% endfor %}>
   <h{{ headingLevel }} class="nhsuk-warning-callout__label">
-    {%- if "Important" in params.heading %}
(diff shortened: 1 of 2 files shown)


packages/components/warning-callout/template.njk
```diff
@@ -3,13 +3,13 @@
 {%- if params.classes %} {{ params.classes }}{% endif %}"
 {%- for attribute, value in params.attributes %} {{attribute}}="{{value}}"{% endfor %}>
   <h{{ headingLevel }} class="nhsuk-warning-callout__label">
-    {%- if params.heading != "Important" %}
+    {%- if "Important" or "important" in params.heading %}
```
mcheung-nhs, line 6: I couldn't get this to work when having "Important" or "important" in the heading.
I had to use this:
`{%- if "Important" in params.heading or "important" in params.heading %}`
Could you confirm this works for you too?

mcheung-nhs (changes requested):

mcheung-nhs (approved):

PR #661, opened 2020-11-24
Fix Install, Examples and About footer link URLs

henocookie:
## Description
Fix the URLs that "Install", "Examples" and "About" footer links direct to.

## Checklist

- [ ] CHANGELOG entry


tomdoughty (changes requested):
This leaves the links broken on dev... it might be best to configure a base-path that we can pass to the GitLab pages build which will append to the links when necessary

AdamChrimes (approved):

tomdoughty (approved):

PR #715, opened 2021-03-11
stylelint

magicaljellybeans:
## Description
Removed sass-lint and replaced it with [stylelint](https://stylelint.io/). This also comes with some plugins for working with SCSS specifically. This is because sass-lint was no longer maintained and stylelint is, as well as being pretty extensible.

I have kept all the same linting rules in pretty much every case, although there are some changes. For example, stylelint will allow duplicate properties if they are consecutive, like when `display: flex` comes after `display: block`. I've tried to ensure that running the old sass-lint config and running this one have the same effect.

I have also tried to not change the SCSS as much as possible. I have noticed that some of the vendor prefix properties have been moved around because of an alphabetical property-order rule. If the order is important I can rectify this and add any necessary exceptions.

Other changes are mostly altering the exceptions to stylelint format and obviously writing the config.

## Checklist

- [x ] Tested against our [testing policy](https://github.com/nhsuk/nhsuk-frontend/blob/master/docs/contributing/testing.md) (Resolution, Browser & Accessibility)
- [ x] Follows our [coding standards and style guide](https://github.com/nhsuk/nhsuk-frontend/blob/master/docs/contributing/coding-standards.md)
- [ x] CHANGELOG entry


magicaljellybeans:
@chrimesdev cool, rebased it now

chrimesdev (approved):

chrimesdev:
Thanks @magicaljellybeans, merged!

PR #723, opened 2021-03-25
replace hardcoded page templates with link to the service manual

chrimesdev:
## Description

Rather than having a hardcoded HTML page template in the using compiled files documentation, link to the guidance in the NHS digital service manual.

Add `script` tags to the JavaScript example


davidhunter08 (approved):
LGTM

Issue #728, opened 2021-04-14
Implications of setting font-family on html selector


windy1:
# Bug Report

## What is the issue?

Because of [how cache partitioning works in most modern browsers](https://www.zdnet.com/article/chromes-new-cache-partitioning-system-impacts-google-fonts-performance/), it's now recommended that websites self-host their fonts. Unfortunately, `nhsuk-frontend` does not make this easy to achieve. The culprit is [setting the `font-family` on an `html` selector](https://github.com/nhsuk/nhsuk-frontend/blob/master/packages/core/elements/_page.scss#L19). Even when overriding with an `!important` flag, Chrome will still attempt to load the external font because it will still be set on the original stylesheet.

I would recommend that the `font-family` not be set explicitly on the `html` selector and instead leave it to the downstream dependents to set.

## What steps are required to reproduce the issue?

1. Create an application that overrides the `font-family` in the `html` selector
2. Observe that the `nhsuk-frontend` font is still loaded

## What was the environment where this issue occurred?

* Operating System: Windows 10 Pro
* Browser: Google Chrome
* Browser version: 89
* NHS.UK frontend package version: 4.1.0
* Node version: 14.x
* ~npm~ yarn version: 1.22.10

AdamChrimes:
Hey @windy1 thanks for opening this issue, the team will get together and have a chat about this. If you want to open a pull request with your proposed idea or solution that is something we'd be grateful for.

windy1:
Hi @AdamChrimes, thanks for the timely response. I've submitted a PR https://github.com/nhsuk/nhsuk-frontend/pull/732

Thanks!

andymantell:
This was fixed in #732

PR #732, opened 2021-04-22
Conditionally load fonts to avoid performance issues

windy1:
## Description

Related: https://github.com/nhsuk/nhsuk-frontend/issues/728

This PR fixes the above issue by separating font loading into a separate stylesheet; disabling the font by default. If a dependent wishes to load fonts from `assets.nhs.uk` they will need to include the new `nhsuk-fonts.css` stylesheet. I've incremented the MAJOR version in `package.json` to signify this as a breaking change for those who are using `nhsuk-frontend` already.

Further reading: https://developers.google.com/web/updates/2020/10/http-cache-partitioning

## Checklist

- [x] CHANGELOG entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@
   ```
 
   Note: You may need to change the path to `node_modules` depending on your project structure.
+- Added `$nhsuk-fonts-path` and `$nhsuk-include-font-face` variables to allow for self-hosted fonts or
+disabling font altogether
 
 :wrench: **Fixes**
 
--- a/packages/core/README.md
+++ b/packages/core/README.md
@@ -229,6 +229,14 @@ Hide elements visually but keep it in the DOM, useful for screen readers.
 <p class="nhsuk-lede-text">Advice, tips and tools to help you make the best choices about your health and wellbeing.</p>
 ```
 
+### Font
+
+The default `@font-face`, "Frutiger", is loaded from `https://assets.nhs.uk`. The host for the fonts can be
+overridden or disabled entirely.
+
+- `$nhsuk-fonts-path`: base URL to load fonts from (e.g. `/fonts/`; trailing slash required)
+- `$nhsuk-include-font-face`: set to false to disable the inclusion of the `@font-face` definition entirely
+
 ## Breakpoints
 
 ```
--- a/packages/core/elements/_page.scss
+++ b/packages/core/elements/_page.scss
@@ -16,8 +16,12 @@
 
 html {
   background-color: $color_nhsuk-grey-4;
-  font-family: $nhsuk-font, $nhsuk-font-fallback;
   overflow-y: scroll; /* [1] */
+
+  @if $nhsuk-include-font-face {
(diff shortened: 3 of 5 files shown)


andymantell:
It would be simpler if we were to follow a an approach more similar to govuk-frontend. They have a `$govuk-include-default-font-face` variable which is set to `true` by default like this:

```$govuk-include-default-font-face: (
  $govuk-font-family == $govuk-font-family-gds-transport
) !default;
```

So if developers do nothing, they will get the fonts, because they are loaded like this:

```
@if ($govuk-include-default-font-face) {
    @include _govuk-font-face-gds-transport;
  }
```

There are then two ways that this can be overriden. Either by setting `$govuk-font-family` to something else entirely. Or by explicitly setting `$govuk-include-default-font-face` to `false`.

In our case, you would be wanting to leave `$nhsuk-font` set to Fruitger, so you would fall down the second option of manually setting `$nhsuk-include-default-font-face` to `false`.

This approach has the advantage of requiring no breaking changes because anyone doing an upgrade will still receive the default setup. As well as being slightly more consistent with the govuk-frontend approach (Not something we have to be bound by, but where we can steer close to their approaches I think it is beneficial).

windy1:
@andymantell Thanks for the feedback. I did consider something like this but this would require those who depend on the package to compile the SCSS themselves instead of using the pre-compiled CSS in the `dist` directory, no? I suppose this would be OK, but would complicate the dependency management a bit. Please let me know if I've understood what you mean correctly. Thanks.

andymantell:
@windy1 I think I would see hosting the font in a different way to the default as enough of a departure that I'd probably expect the person to be compiling the scss themselves, and I would anticipate a large proportion of the people using this package are probably already doing so.

It may be in the (near?) future that we turn this on it's head and that the font files come back in to be served directly rather than via assets.nhs.uk, but there's conversations we need to have around licencing that complicate this slightly so we may not be quite ready to do this just yet.

In the interim, the ideal solution from our perspective is something that enables you to do what you need to do (which I agree is the right thing), but presents the smallest change to anyone already consuming this package. I think the solution which treads closely to the scss variables from govuk-frontend probably satisfies this.

Sorry if that means a bit of rework of this pull request. If you're still happy to pick it up we'd be very grateful for the contribution.

andymantell:
Hi @windy1, thanks for this. I've got a busy day ahead so will review this properly next week. Initial scan looks good though, thankyou.

windy1:
Any update on this @andymantell ?

andymantell:
Sorry yes, I've had a look through and it looks good to me. 👍🏻 

It does need some documentation if you're up for doing that? Probably just a little something in here documenting the new variables and how to use them:

https://github.com/nhsuk/nhsuk-frontend/blob/master/packages/core/README.md

Once that's done I think we're happy to merge. There's a few things queued up for release so hopefully we could then get an actual release out for you soonish.

andymantell:
Also a changelog entry would be good please. Thanks!

windy1:
Done!

andymantell (approved):

PR #742, opened 2021-05-11
update README to remove specific contributors

chrimesdev:
## Description

Over the years this library has grown and has been contributed to by a number of people (both through code and non code contributions), these specific people helped us start the library, but since there have been many more who have helped build and grow it further. So it would be unfair to highlight only a select few. 

Credit to the Government Digital Service remains in code and specific documentation.


davidhunter08 (approved):
LGTM

PR #750, opened 2021-06-25
Tabs

andymantell:
## Description

Porting the tabs component from govuk-frontend.
See https://github.com/nhsuk/nhsuk-service-manual-community-backlog/issues/279

## Checklist

- [ ] CHANGELOG entry


andymantell:
@tomdoughty @chrimesdev Couple of things I'd like to discuss with you before I go any further...

1) govuk-frontend uses `data-module` attributes to attach JavaScript behaviours to but this doesn't seem to be a thing in nhsuk-frontend. Would you prefer that the JS is bound to the tabs class which is used for the styling? This would mean people wouldn't be able to do "serverside tabs" with a full page load on each tab, since the JS would get in the way. Or would you prefer the JS and the tab-like presentation be separable as in the govuk-frontend case?

2) I've pretty much just brought the govuk-frontend tabs JS in as it stands with a couple of minor tweaks to make it "fit in" with the way NHS components are initialised. I don't know whether you want me to do any further refactoring to make it fit some notion of a "house style" however? The JS doesn't _look_ like the rest of our JS (use of prototype etc etc). The upside of this of course is that if govuk make any fixes at their end then they're fairly easy to bring in. (I'm aware the linting rules don't currently pass - I will address that)

wa-rren-dev:
Closing here as tabs in this form are now merged into [alpha v7](https://github.com/nhsuk/nhsuk-frontend/pull/831)

PR #755, opened 2021-07-16
Update header readme

mcheung-nhs:
## Description
In the nunjucks table of the header readme (https://github.com/nhsuk/nhsuk-frontend/tree/master/packages/components/header#nunjucks-arguments), the description of the 'service' argument is incorrect - there is no 'longName' option in the macro, so this content should be removed.
Now, because there is only one parameter to the service object, there's no need for it to be an object. However, changing the macro would be a breaking change, so probably best update the readme first until the header component is refactored.

## Checklist

- [ ] CHANGELOG entry

### DIFF
--- a/packages/components/header/README.md
+++ b/packages/components/header/README.md
@@ -806,7 +806,7 @@ compiled JavaScript for all components `nhsuk.min.js` or the individual componen
 
 [Preview the header organisational with white header and navigation component](https://nhsuk.github.io/nhsuk-frontend/components/header/header-org-white-nav.html)
 
-#### HTML markup 
+#### HTML markup
 
 ```html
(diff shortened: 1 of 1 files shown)


chrimesdev:
This will also need updating on the [NHS digital service manual Header page](https://service-manual.nhs.uk/design-system/components/header)

mcheung-nhs:
I can't find any reference to the 'service' argument in any examples on the NHS digital service manual header page.

chrimesdev:
I'll take a look at why it is missing from the NHS digital service manual and see if it needs updating

chrimesdev:
Resolved by https://github.com/nhsuk/nhsuk-frontend/pull/757

PR #756, opened 2021-07-26
Checkboxes - none of these JS behaviour

### LINKED ISSUE #609: radiobutton/checkboxed in nested fieldsets inadvertently setting conditional to hidden
# Bug Report

## What is the issue?

In the example shown in the screenshot below, there are radio buttons shown only when the user chooses the first option("I am employed or self employed").
When I click in any of the radio buttons below the conditional content will then get closed and the option I just clicked on disappears.

So, on click, it's running `addClass(input, 'nhsuk-radios__conditional--hidden')` to all `conditionalInputs` in the page.

One of two things I think should happen:
- either the `addEventListener` is added only to radiobuttons/checkboxes  that could affect the aria-expanded;
- find the `.nhsuk-radios--conditional` or `nhsuk-checkboxes--conditional` which is an immediate parent of the current node, and apply the `toggleConditionalInput` only to that;

image

andymantell:
## Description

Add JavaScript behaviour to a set of checkboxes which unsets other checkboxes in a given group when the "None" option is checked, as in the gif below:

ezgif-2-6af8cacf2150

Also Fixes #609 

## Checklist

(*Tested in all except Samsung internet as browserstack wasn't working for it*)

- [x] CHANGELOG entry


app/components/checkboxes/conditional.njk
```diff
@@ -44,43 +44,45 @@
...
-            "hint": {
-              "text": "Select all options that are relevant to you."
-            },
-            "items": [
-              {
-                "value": "email",
-                "text": "Email",
-                "conditional": {
-                  "html": emailHtml
+          <form>
```
andymantell, line 47: Diff looks huge but it's just the addition of the <form> tag

andymantell:
@tomdoughty @chrimesdev I think this is ready for review now. Just about to do some browser testing but I think it's complete otherwise - happy to jump on a call and discuss if that is easiest - it might be good to highlight some places where I have had to deviate from the govuk-frontend implementation

app/components/checkboxes/none-of-the-above.njk
```diff
@@ -0,0 +1,108 @@
+{% set html_style = 'background-color: #f0f4f5;' %}
+{% set title = 'Checkboxes with conditional content' %}
```
chrimesdev, line 2: Very minor but page title just needs to be `Checkboxes with "none of the above" option`

packages/components/checkboxes/template.njk
```diff
@@ -56,6 +56,8 @@
       <input class="nhsuk-checkboxes__input" id="{{ id }}" name="{{ name }}" type="checkbox" value="{{ item.value }}"
```
chrimesdev, line 56: HTML validator is complaining that the `input is missing required attribute role.`

The HTML validation seems to pass without errors on the GOV.UK example.

app/components/radios/nested-conditionals.njk
```diff
@@ -0,0 +1,128 @@
...
+
+{%- set nestedConditional %}
+    {{ radios({
+      "idPrefix": "inner",
+      "name": "inner",
+      "fieldset": {
+        "legend": {
+          "text": "How would you prefer to be contacted?",
+          "classes": "nhsuk-fieldset__legend--l",
+          "isPageHeading": "true"
```
chrimesdev, line 49: Should we remove this `isPageHeading` to prevent there being multiple `<h1>` on the page?

packages/components/checkboxes/template.njk
andymantell: As discussed on Slack, this is a known issue at the moment and will go away if all the proposed drafts and changes to the aria spec land. It's not an issue introduced by this pull request, and is present in gov-frontend also (Where they are monitoring the situation ongoing).

See also:

https://github.com/validator/validator/issues/782
https://github.com/w3c/aria/issues/1404

tomdoughty (approved):
LGTM 👍

chrimesdev:
Cheers Andy!

Issue #720, opened 2021-03-17
Transactional header long service names not wrapping correctly 


chrimesdev:
## Description
 
Long services names on the [Header transactional with service name](https://nhsuk.github.io/nhsuk-frontend/components/header/header-transactional-service-name.html) aren't wrapping correctly on smaller mobile devices. 

There is a modifier class to fix this `nhsuk-header__transactional-service-name--long`, however the default behaviour should be that the service name always wraps below without the need for the modifier class.

We should remove the `nhsuk-header__transactional-service-name--long` modifier class and make the default behaviour wrap underneath for service names of any length. This behaviour would also match the [Header with a service name](https://nhsuk.github.io/nhsuk-frontend/components/header/header-service-name.html).

Example of the current default:

![Screenshot 2021-03-17 at 14 21 14](https://user-images.githubusercontent.com/11615501/111483661-ea0d8f00-872c-11eb-967b-332d723f48a5.png)

Example when using the modifier class (ideal default)

![Screenshot 2021-03-17 at 14 28 56](https://user-images.githubusercontent.com/11615501/111483877-20e3a500-872d-11eb-8e2b-4659e5a481a0.png)

AdamChrimes:
Related comment https://github.com/nhsuk/nhsuk-service-manual-backlog/issues/16#issuecomment-621918905

PR #757, opened 2021-07-26
Make transactional headers wrap by default

andrewbarnesweb:
Resolves: #720 

The transactional header did not wrap correctly on smaller devices. The long name modifier was removed, and instead merged into the base class. This seemed to take precedent of the tablet spacing.

I assumed that since the modifier was to be removed, and made the default behaviour, it didn't make sense to have an example for the longer name. So the transactional header long name example has been removed.

## Checklist

- [x] CHANGELOG entry


chrimesdev:
Thanks for opening this pull request @andrewbarnesweb - I'll give it look through tomorrow.

andrewbarnesweb:
Awesome. I'm getting a number of what I'd call false positives in Backstop locally (as I need to approve the header changes), so I don't want to approve those, as I expect they'll fail in ci. Any ideas how to deal with these? Or if there might be something in my environment that is awry?

image

chrimesdev:
@andrewbarnesweb I've checked the GitHub actions CLI logs and the only tests failing are the 3 related to the `Header_transactional_with_service_name`, which can be approved with the changes in this pull request.

When running the BackstopJS tests locally are you running them using Docker? We use Docker to run our tests within a consistent environment, this prevents cross-platform rendering issues (such as font rendering and aliasing) - which I think may be happening for you here. 

Theres some documentation here about running the BackstopJS tests in Docker: https://github.com/nhsuk/nhsuk-frontend/blob/master/docs/contributing/automated-testing.md#test-commands

chrimesdev:
I think the version in the `CHANGELOG` can be `5.1.1`, following [SEMVER](https://semver.org/), this would just be a fix rather than a breaking change

chrimesdev:
I think once the `CHANGELOG` and BackstopJS reference screenshots are updated, this looks good to go. Thank you for this @andrewbarnesweb.

andrewbarnesweb:
> @andrewbarnesweb I've checked the GitHub actions CLI logs and the only tests failing are the 3 related to the Header_transactional_with_service_name, which can be approved with the changes in this pull request.
>
> When running the BackstopJS tests locally are you running them using Docker? We use Docker to run our tests within a consistent environment, this prevents cross-platform rendering issues (such as font rendering and aliasing) - which I think may be happening for you here.
>
> Theres some documentation here about running the BackstopJS tests in Docker: https://github.com/nhsuk/nhsuk-frontend/blob/master/docs/contributing/automated-testing.md#test-commands

Thanks for getting back to me!

Yeah, spotted the obvious ones in CI. I'm running it in Docker, will give my environment a kick to see if I can get it working OK locally! 😄 When I get that working, will amend the SEMVER version also (sorry about that!).

andrewbarnesweb:
Spent half an hour on it, without much luck. If I run in Docker, I get 16 issues (including the transactional ones). If I run outside of Docker, I get 312 issues (as expected I guess!). But as you suggest, Docker is designed to do away with this sort of issue. Not sure what I'm doing wrong at this stage!

I'm running:

- MacOS Big Sur
- BackstopJS v5.3.5
- Docker Desktop for Mac (Intel) v3.5.2
- Node v15.14.0

chrimesdev:
Are you able to push the 3 transactional header reference screenshots to get the CI to pass? I will investigate as to why you are getting errors with your Docker locally in the meantime - seems a bit strange as I'm running a similar setup as you but seems OK for me and the GitHub actions pipeline.

chrimesdev:
It looks like the Docker image on the CI and my local machine is using BackstopJS `v5.0.8` (inherited from `package.json`) when upgrading to the latest version of BackstopJS `v5.3.5` the tests are failing (16 fails same as you @andrewbarnesweb - attached the report).

[backstopjs_html_report.zip](https://github.com/nhsuk/nhsuk-frontend/files/6893511/backstopjs_html_report.zip)

It's quite difficult to see what has changed in the [BackstopJS releases](https://github.com/garris/BackstopJS/releases) between `v5.0.8` to `v5.3.5` due to the lack of `CHANGELOG` on the BackstopJS project (https://github.com/garris/BackstopJS/issues/330) 

We may need to keep the version pinned to v5.0.8 until we can investigate further.

andrewbarnesweb:
Nice one. Thanks for debugging that for me! Was banging my head against it. When I get a moment later, I'll downgrade and try again. 👍🏻

chrimesdev:
Thanks @andrewbarnesweb this all looks good now, I've created an issue to look further into the BackstopJS issue we were getting. Appreciate the contribution ⭐

chrimesdev (approved):

PR #761, opened 2021-08-03
Fix presentation of nhs logo when css is disabled

### LINKED ISSUE #747: Consider the appearance of components when CSS is not loaded
# Bug Report

## What is the issue?

When viewing an nhsuk-frontend based site without CSS there are some aspects that become very large (the back arrow) or fail to convey their purpose (the NHS logo). This doesn't matter most of the time - but this morning the CSS was failing to load on one of the covid related services in production. The service still worked (Hooray for progressive enhancement!) but the NHS logo and the back arrow marred what was otherwise still quite a usable service.

We should investigate ways to make these (and any of the other components) more usable without CSS in case this happens again. Off the top of my head - putting some fallback colours natively into the NHS SVG, and giving it and the back button SVG some inherent dimensions would be a good start.

### Design system homepage without CSS:

image

---

### And one of the covid service whilst it had _partial_ CSS

(For some reason it wasn't _all_ missing, but the bug still rears it's head)
image

andymantell:
## Description

Fixes #747

## Checklist

- [ ] CHANGELOG entry

### DIFF
--- a/packages/components/header/template.njk
+++ b/packages/components/header/template.njk
@@ -24,9 +24,9 @@
       </a>
       {%- else -%}
       <a class="nhsuk-header__link{% if params.service %} nhsuk-header__link--service {% endif %}" href="{% if params.homeHref %}{{ params.homeHref }}{% else %}/{% endif %}" aria-label="{% if params.ariaLabel %}{{ params.ariaLabel }}{% else %}NHS homepage{% endif %}">
-        <svg class="nhsuk-logo" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 40 16">
-          <path class="nhsuk-logo__background" d="M0 0h40v16H0z"></path>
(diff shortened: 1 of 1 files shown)


chrimesdev:
This looks good to me, we just need to update the HTML markup in the individual component `README` files eg. https://github.com/nhsuk/nhsuk-frontend/blob/master/packages/components/action-link/README.md

We're planning on moving the HTML markup out of the `README` and just have the single source of HTML/Nunjucks truth on the NHS digital service manual - but for now we best just update it in there too.

Also I think we might need to update the SVG asset files incase anyone uses them for includes found in the `packages/assets/icons` folder eg. https://github.com/nhsuk/nhsuk-frontend/blob/master/packages/assets/icons/icon-arrow-left.svg?short_path=e101404

andymantell:
@chrimesdev For the heights/widths inside the SVG files themselves - some of the SVGs are used in multiple places with different dimensions, so it's not possible for me to set the "correct" dimensions in each file for every case. I have therefore defaulted the SVG files to 34px x 34px which is the same as the default `$nhsuk-icon-size` from the SCSS. If people are including them in this manner, this should still be sufficient to tidy up the default output.

andymantell:
@chrimesdev I have addressed your comments and this one is ready for review again. Cheers!

chrimesdev (approved):
Looks good thanks @andymantell

PR #762, opened 2021-08-03
Rework clickable cards using CSS invisible overlay rather than JS event handler

### LINKED ISSUE #749: Ctrl + Click on card doesn't open new tab
# Bug Report

Ctrl + Click not respected on card components

## What is the issue?

When ctrl + clicking on the body of a clickable card, the link navigates insside the current tab.
It is expected that ctrl + click creates a new tab in most browsers.

Ctrl + Click on a card. (examples here https://www.nhs.uk/pregnancy/)

N.B ctrl + click on the title of the card opens a new tab _and_ navigates the current tab.

### LINKED ISSUE #758: Clickable Cards open two new tabs when anchor target is set to "_blank"
# Bug Report

## What is the issue?

When using Clickable Cards for links to external sites (and thus expecting links to open as new tabs), clicking the link itself causes two new tabs to open, instead of one. Clicking anywhere else in the card only causes one new tab to open, so this is likely an event propagation issue.

**Note**: Initial tests suggest this issue affects Firefox, but not Chrome or Safari.

To reproduce, recreate the HTML example from the [NHS Digital Service Manual](https://service-manual.nhs.uk/design-system/components/card#clickable-card) but with the anchor's `target` attribute set to `_blank`:

```html
<div class="nhsuk-card nhsuk-card--clickable">
  <div class="nhsuk-card__content">
    <h2 class="nhsuk-card__heading nhsuk-heading-m">
      <a class="nhsuk-card__link" href="#" target="_blank">Introduction to care and support</a>
    </h2>
    <p class="nhsuk-card__description">A quick guide for people who have care and support needs and their carers</p>
  </div>
</div>
```

Open the page in Firefox and click on the link inside the card.

Alternatively the issue can be observed by navigating to the bottom of https://www.nhs.uk/, inspecting one of the cards, adding the `target="_blank"` attribute and then clicking on the link.

* Device: Apple Macbook Pro (13-inch, 2019)
* Operating System: macOS Catalina
* Browser: Mozilla Firefox
* Browser version: 90.0.1

andymantell:
## Description

Rework clickable cards using CSS invisible overlay rather than JS event handler

This avoids problems with using Ctrl-click, middle click, right click to open new tabs since the link just behaves as it would natively

Fixes #758
Fixes #749

## Checklist

- [x] CHANGELOG entry

## Issues observed during browser testing

- ~IE10 - image is not clickable in "card with an image" example. Ignoring, because IE10 is not supported and the experience is ok.~
  Fixed. Don't know whether we want IE10 fixes lingering around, but it's relatively inoffensive, so might not hurt to have it for a while? Easily removed if we don't want it...


packages/components/card/card.scss
```diff
@@ -54,20 +53,18 @@ $card-border-hover-color: $color_nhsuk-grey-3;
...
-    .nhsuk-card__heading a,
-    .nhsuk-card__link {
-      color: $nhsuk-link-hover-color;
-      text-decoration: none;
-
-      &:focus {
-        color: $nhsuk-focus-text-color;
-      }
+  .nhsuk-card__heading a,
+  .nhsuk-card__link {
```
andymantell, line 57: @chrimesdev I've put both these selectors in here in order to be consistent with what came before. Is this still necessary? I assume it's backwards compatibility before the `.nhsuk-card__link` class was added

packages/components/card/card.scss
```diff
@@ -54,20 +53,18 @@ $card-border-hover-color: $color_nhsuk-grey-3;
...
+    &::before {
+      background-color: rgba(255, 255, 255, 0); /* [8] */
```
andymantell, line 59: @chrimesdev Don't know whether we want to carry IE10 baggage or not? It's a relatively innocuous little hack to force IE10 and 9 to render the clickable overlay. But we don't "officially support" IE10.

Your call...

chrimesdev:
@Fenwick17 as you did the original work with the clickable Card, are you OK to take a look at this? I'll also give it a review

Fenwick17:
Spotted one issue that occurs with MacOS Hover Text. 
The large-text view does not work for the content of the card due to the overlay, hovering the heading will display the large-text view, but not the body content.

Everything else appears to be working as you would expect.

andymantell:
Thanks @Fenwick17 - do you mean these ones?
image

andymantell:
Sorry, I see what you mean now. Using https://support.apple.com/en-gb/guide/mac-help/mchlb203bc78/mac

Let me have a think about this 🤔

Fenwick17:
Added a gif of the issue, if it can't be resolved we will have to make a decision on whether it is acceptable as it is. 
I would image any Windows software that does this would have a similar behaviour

Hover text not display on hover

andymantell:
yeah makes sense. I'm on windows so I'll see if there's anything similar here. And will have a think if there's anything sensible can be done

andymantell:
@chrimesdev @Fenwick17 I'm not sure what to do about this. Surely anything that alters what HoverText "sees" would also affect what a screenreader sees. So if we start changing things for HoverText, we would negatively affect other tools.

I don't know if a perfect solution exists for this then. I think we're stuck choosing between two slightly imperfect solutions:

1) Go back to the JS way of doing this, and try and enhance it for more ways to open new tabs (Ctrl-Click, middle click etc). This will never catch all the ways though (right click / long press on mobile to open new tab from the context menu for example)

2) Go with the invisible overlay, and accept the hovertext deficiency.

Any thoughts? Have I missed any other options?

packages/components/card/card.scss
chrimesdev: Happy to include this - 1 line of code to make it work in a whole browser isn't bad!

Fenwick17:
I just performed some Browserstack testing and didn't spot any issues with the CSS only solution.
So the only issues we seem to have are not being able to copy text and no hover text on the paragraph.

For both of these I personally don't think they are show stoppers.   
The text is meant to just be a little bit of extra context so there shouldn't really be a need to copy it.  
The link itself ideally should be clear enough for a user to know what it means without needing the additional text and standard zoom behaviour will still work as a fallback.

The only thing that comes to mind at the moment is if we add visually hidden content to the link itself, so that it will show up with Hover Text. 
```
<a class="nhsuk-card__link" href="./pages/examples.html">
  Examples<span class="nhsuk-u-visually-hidden">: Page layouts, text pages and components</span>
</a>
```
Then you would have to use `aria-hidden` on the `<p>`.
This is probably wildly overkill, but would _solve_ that issue (could introduce others though). Certainly won't work if theres a lot of content in the card.

What do you think?

chrimesdev:
@andymantell do you think this needs a little more information around removing the JavaScript? Incase anyone is importing and initialising the Card JavaScript and we have now removed it, would it be a breaking change?

DomBaker (approved):
Reviewed and approved. ready for v7.0.0 of frontend library, date for this is unconfirmed

wa-rren-dev (approved):

Issue #768, opened 2021-08-23
Search input: focus state style (desktop)


idiosyn:
The search input focus state style on desktop has a thin black border. It’s different from the [standard focus state styles](https://service-manual.nhs.uk/design-system/styles/focus-state). Is there any reason for that?

![image](https://user-images.githubusercontent.com/38734779/130381615-3fe4e9a2-858d-42dd-85e2-da1c8898ff01.png)

If not, I'd like to suggest these changes:

https://github.com/nhsuk/nhsuk-frontend/blob/5ecf5c953c023f20f020a99206159248b432966b/packages/components/header/_header.scss#L355

``` scss
      border: 4px solid $nhsuk-focus-text-color;
```

https://github.com/nhsuk/nhsuk-frontend/blob/5ecf5c953c023f20f020a99206159248b432966b/packages/components/header/_header.scss#L359

``` scss
      padding: 0 9px; 
```

So the final result is:

![image](https://user-images.githubusercontent.com/38734779/130381969-1a12ba54-f063-464e-966a-b666d88f996b.png)

chrimesdev:
Hey @idiosyn I will look into this and see if theres a reason it is this way. I may need to check with a Designer about it.

karlgoldstraw:
Seems to work as expected in mobile breakpoints, but not on desktop. Will take a closer look. 

<img width="581" alt="Screenshot 2021-08-23 at 12 49 11" src="https://user-images.githubusercontent.com/3852457/130442313-f4cee746-1f08-48bd-9ee0-5dfa68b9004f.png">

karlgoldstraw:
@chrimesdev I've had a look around at this and can find absolutely nothing stating that this was done for a reason, so please can we set it up as a bug to be fixed as suggested by @idiosyn above? 

Or @idiosyn if you'd like to make pull request of the changes?

PR #771, opened 2021-08-25
Search box tweak

idiosyn:
Search input focus state style (desktop): fix border width.

Closes https://github.com/nhsuk/nhsuk-frontend/issues/768.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 - Transactional header defaults to wrapping underneath the logo without the need for a modifier ([Issue 720](https://github.com/nhsuk/nhsuk-frontend/issues/720)).
 - Add width, height, stroke and fill attributes to inline SVGs in order that they render at appropriate sizes when viewed with disabled/broken/missing CSS  ([PR 761](https://github.com/nhsuk/nhsuk-frontend/pull/761)).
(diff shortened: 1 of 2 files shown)


chrimesdev:
If these styles now match those on breakpoints lower than tablet breakpoint, we probably don't need most or all of these `&:focus` styles block for > tablet breakpoint.

Earlier in the code we already have these styles : https://github.com/nhsuk/nhsuk-frontend/blob/07d4c5c19bc8a5fa79251ec10b9284395c857892/packages/components/header/_header.scss#L25 so now the only difference now would be the `padding: 0 13px;` compared to the `padding: 0 13px;` for < tablet breakpoint. But they could probably even be both be 13px? So we might be able to get rid of this full block now.

karlgoldstraw (approved):
Looks good to me

karlgoldstraw:
> If these styles now match those on breakpoints lower than tablet, we can remove some of the duplication by removing most of the styles for `&:focus` above tablet breakpoint.
> 
> Earlier in the code we already have these `focus` styles:
> 
> https://github.com/nhsuk/nhsuk-frontend/blob/07d4c5c19bc8a5fa79251ec10b9284395c857892/packages/components/header/_header.scss#L253
> 
> so now the only difference now would be the `padding: 0 9px;` compared to the `padding: 0 13px;` for the lower than tablet breakpoint. We might even be able to keep 13px and then be able to remove the full block here.

I've approved the PR, but will leave these amends to a dev to look at as tech debt. Perhaps @DomBaker might be able to help?

Issue #721, opened 2021-03-18
Menu toggle on header doesn't work with Javascript disabled


StephenHill-NHSBSA:
# Bug Report

## What is the issue?

When Javascript is disabled and a user is on a mobile device, the <MENU> toggle no longer works to reveal the different menu items.

**What happens:**

![image](https://user-images.githubusercontent.com/45825845/111646809-e09e2880-87f9-11eb-82a7-b2011642683d.png)

**Expected:**

![image](https://user-images.githubusercontent.com/45825845/111647781-ada86480-87fa-11eb-867e-9ce37c9cc9f0.png)

## What steps are required to reproduce the issue?

1. Disable Javascript on your browser / mobile device.
2. Resize the browser window to mobile size / view the site on mobile
3. Try clicking on the <MENU> toggle

## Is there anything else you think would be useful in recreating the issue?

This problem exists across a number of services, but the most prominent one is NHS.UK.

idiosyn:
This happens with the search toggle too.

AdamChrimes:
Thanks for raising this issue @StephenHill-NHSBSA, we're going to doing some work on the Header in the near future which will see a few design changes to both the Menu and the Search. This is something we'll factor into that piece of work.

nitin094-hub:
Can you plz assign this task to me?

PR #777, opened 2021-10-01
[refs #721] Add menu and search toggles when JS is disabled

jtuds:
This commit adds toggle functionality for the menu and search header buttons
when the user has JS disabled, via input checkbox toggles.
See https://github.com/nhsuk/nhsuk-frontend/issues/721

## Checklist

- [ ] CHANGELOG entry

### DIFF
--- a/packages/components/header/_header.scss
+++ b/packages/components/header/_header.scss
@@ -36,6 +36,8 @@
  * 15. Remove random top margin in Safari
  * 16. Align close icon with nav item arrow icons
  * 17. Add nhsuk-spacing(9) to align right and left main nav with header
+ * 18.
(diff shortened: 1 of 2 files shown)


chrimesdev:
Hey @jtuds thanks for opening this pull request, the team will put some time in to review this soon.

jtuds:
No problem thanks for the feedback :) I completely agree the simpler approach to this is to just show the navigation and search without any interaction, but I suppose that just depends on the desired functionality with JS disabled which I don't have enough context for.

Happy to wait for more thinking around this. If it is decided that this approach is the right one I would perhaps suggest raising the accessibility issues as a separate ticket in order to get a fix out for the non-Javascript functionality - though the importance/urgency of this fix depends on the stats on users coming to the site without JS enabled.

PR #798, opened 2021-12-02
Fix charities spelling mistake

mcheung-nhs:
## Description
Fix spelling mistake - 'charitiest' => 'charities'

## Checklist

- [ ] CHANGELOG entry


andymantell:
the nvmrc file in the root says version 12...

andymantell:
Michael - what's the status of this one - are you happy with the rebuilt package-lock? I'm trying to run a build on a different pull request and am getting failures that would be fixed by what you've done here.

What node version did you build it with? We should update the `.nvmrc` file in the root to reflect this - and maybe document it. If people use different node versions all the time to update dependencies we will get a nasty amount of churn in the package-lock. Previously it _was_ 12 just _because_, but that is ancient history. I think it's time to get up to date...

andymantell:
We might also consider using a similar approach to govuk-frontend's check-nvmrc script - which prevents people using nodejs versions other than the one specified in the repository, to avoid package-lock churn:

https://github.com/alphagov/govuk-frontend/blob/main/bin/check-nvmrc.js

mcheung-nhs:
Hi @andymantell - I've created a separate [PR](https://github.com/nhsuk/nhsuk-frontend/pull/803) with just the dependency updates and have also updated the .nvmrc file to 16 which seems to have built successfully. What do you reckon?

PR #816, opened 2022-05-10
Use fallback font if not including NHS font face

jacksonj04:
## Description

There is a configuration option `$nhsuk-include-font-face` which controls if the NHS corporate font is included in the compiled CSS. In the current implementation, if this option is false then `font-face` is not set at all, which leads to unstyled text.

This change makes it so that if the fonts are not included, a `font-face` declaration is still made using only the fallback fonts (as set in `$nhsuk-font-fallback`).

## Checklist

- [x] CHANGELOG entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # NHS.UK frontend Changelog
 
+## Unreleased
+
+:wrench: **Fixes**
+
+- `font-family` is now being set (to the fallback font) even when not including NHS typefaces
+
 ## 6.1.0 - 12 January 2022
 
 :new: **New features**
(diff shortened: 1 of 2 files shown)


andymantell:
Not passing judgement on this either negatively or positively, but just wanted to link to this which has some background to why it is like it is right now, i.e. mimicing the way it works in govuk-frontend. And to note that whatever path is taken here needs to be compatible with the needs from this thread - where someone is self hosting the nhs fonts.

https://github.com/nhsuk/nhsuk-frontend/pull/732

colinrotherham:
I'm going to close this based on https://github.com/nhsuk/nhsuk-frontend/pull/816#issuecomment-1122263886 above

Any further improvements that came in GOV.UK Frontend will be picked up in https://github.com/nhsuk/nhsuk-frontend/issues/952

Thanks all

Issue #754, opened 2021-07-09
Details component announces its default and expanded state 1 time but not subsequent times using NVDA


mcheung-nhs:
# Bug Report

## What is the issue?
When using the details component with NVDA, expanding the summary text announces the 'expanded' state and when closing it, the 'collapsed' state is announced. However, subsequent opening and closing does not get announced.

## What steps are required to reproduce the issue?
1. Go to a page with the details component, eg. https://www.nhs.uk/medicines/paracetamol-for-children/
2. Go to section 10, Common Questions
3. Enable NVDA and listen when expanding and collapsing any of the common questions details/summary components.
4. Observe that the first time you expand and collapse, it is announced. However, on subsequent time, it is not announced.

## What was the environment where this issue occurred?

* Device: Windows Surface Pro 4
* Operating System: Windows 10
* Browser:Firefox
* Browser version: 89.0.2
* NVDA version 2020.4

## Is there anything else you think would be useful in recreating the issue?
No

andymantell:
I can reproduce this, on NVDA 2020.4 and 2021.1 in Firefox.
It doesn't occur in Chrome.
And crucially, it doesn't occur when I made a codepen with a really basic details element, so there's something uniquely wrong with the nhsuk-frontend one.

Will investigate...

andymantell:
It appears that the JS polyfill that we are using and applying _regardless_ of the presence of native details/summary support is actually making it worse in Firefox. When I disable this JS, Firefox starts behaving properly.

Speaking to @tomdoughty - this polyfill was put in at a time when even browsers that _did_ support it, didn't support it well, and screenreaders were not announcing the state correctly.

However, it appears time has moved on, and support for the native element and assistive tech is better than it was. I feel therefore that we should only initialise the polyfill when support for the native element is not present.

This is backed up by this commit from govuk-frontend:
https://github.com/alphagov/govuk-frontend/commit/f5cfa5f39a8a0816203c1c69f61c8611e0b4f82f

At the moment, we do still need the polyfill for IE11 however so we cannot yet remove it entirely.

andymantell:
Weirdly I seem to no longer be able to reproduce this issue! I could do this morning, but NVDA has updated itself since and now I can't get back to a broken state.

Nevertheless, I think the fix in the attached pull request is sensible anyway.

PR #821, opened 2022-05-27
Add missing exclamation mark to details component browser compatibility check

bjss-colin:
## Description

This corrects an outstanding bug in the details component which was introduced [with the following commit](https://github.com/nhsuk/nhsuk-frontend/commit/b96defbac870f7de758923038454e14503462d30).

It's a straight forward change as it's an exclamation mark that is missing to negate the condition.

Currently, the bug manifests itself on pages that use details components where the components aren't attributed correctly.

## Checklist

- [X] CHANGELOG entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # NHS.UK frontend Changelog
 
+## Unreleased
+
+:wrench: **Fixes**
+
+- Correctly assess whether the details component is supported on browsers
+
 ## 6.1.0 - 12 January 2022
 
 :new: **New features**
--- a/packages/components/details/details.js
+++ b/packages/components/details/details.js
@@ -8,7 +8,7 @@ import { toggleAttribute } from '../../common';
 export default () => {
   // Does the browser support details component
   const nativeSupport = typeof document.createElement('details').open === 'boolean';
-  if (nativeSupport) {
+  if (!nativeSupport) {
     return;
   }
 


andymantell:
Not sure about this. The code is meant to be without the negation?

```
export default () => {
  // Does the browser support details component
  const nativeSupport = typeof document.createElement('details').open === 'boolean';
  if(nativeSupport) {
    return;
  }
```

In that if we have native support, we want to return early, because we don't need to polyfill anything.

Can you clarify exactly what you're expecting to see, that you aren't seeing?

andymantell:
You can see the same logic here in govuk-frontend:

https://github.com/alphagov/govuk-frontend/blob/main/src/govuk/components/details/details.js#L26

andymantell:
Worth noting that if we have native support, we do not want any of the aria attributes being added. Is that what you are expecting to see? These are only to be included if the browser does not support the details element. Including them on browsers that do support the details element actually makes accessibility worse (Based on UR performed in the 111 online team - See #754 )

bjss-colin:
Hi Andy, thanks for the sanity check.

The issue we're seeing is that some of the more important attributes, such as aria which screen readers will use to determine whether they should be reading the hidden/expanded content, aren't applied to the components since all of this logic is located within the `initDetails` function.

However, having referred to the [Mozilla developer guide](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA), I can see the warning at the top of the page which mentions the migration from ARIA to native support in HTML5.

I'll mark this closed in that case, I thought the logic would/should have been the other way but I can see now that was down to a knowledge gap with HTML5.

andymantell:
👍🏻

PR #838, opened 2022-08-10
Allow back links to be buttons

andymantell:
## Description

On 111 online we (unfortunately) use buttons in some places where links might more normally be used. This is down to the way the application state is managed in post data - we can't use a link and a GET request because the state would be lost.

Therefore, we need back links to be buttons. I currently have these style overrides kept in our `shame.scss` but it would be nice to keep this file as clean as possible by allowing button back links in nhsuk-frontend natively.

I've documented it here in the package - I don't know whether we want to document it on the service manual as well or leave it undocumented to discourage people from doing it if possible. I'm aware it's not ideal from an accessibility perspective - a link should be a link really.

## Checklist

- [x] CHANGELOG entry


wa-rren-dev:
I'm not _personally_ convinced that something should traverse from 'shame.css' into the frontend library, but I'm no gatekeeper! In the absence of a designer on the service manual team, I'd be happy to proceed either way with input from @mcheung-nhs @Fenwick17 @pflynny

andymantell:
it's only in my shame file because I considered it temporary, something I didn't want to live forever in our codebase. Don't get hung up the naming of that file! :-)

pflynny:
LGTM 👍

mcheung-nhs (approved):
All looks good.

PR #840, opened 2022-08-12
Add role=button spacebar activation and button debouncing JS

### LINKED ISSUE #796: Add submit debounce functionality
# Bug Report/Issue report

Not really a bug. On NBS we have had an issue with users double clicking and submitting forms twice.  This has led some users to an error page, even though they have booked a vaccine.  They then proceed to book another vaccine as they are told to.  It seems some users routinely double click and cause extra load on the servers by doing so.

The solution is to add a debounce function similar to what the gov.uk site has  https://github.com/alphagov/govuk-frontend/blob/main/src/govuk/components/button/button.js

I used a solution based on the approach that the gov.uk site uses.

This is a small slack thread discussing the issue 
https://nhsuk.slack.com/archives/CCPLQ9YAJ/p1637328178007900

andymantell:
## Description

Links which are styled to appear visually like buttons can cause problems for users who might expect the button-like element to be clickable with a spacebar since this is how _actual_ buttons behave. This PR adds some JS brought over from govuk-frontend to shim this behaviour in on top - so that links that look like buttons now also _behave_ like buttons.

I have also brought over the button debouncing functionality from govuk while I was here. Which fixes #796 

## Checklist

- [x] CHANGELOG entry


Fenwick17 (commented):
Tested this on a few browsers and within my service locally and all seem to be working as intended.

Fenwick17:
@andymantell Once theres some tests this could go into the version 6 release rather than version 7 I think

andymantell:
@Fenwick17 Agreed. It's on my radar - will put some in when I next get a moment.

mcheung-nhs (approved):

PR #858, opened 2023-02-17
Tell git to keep `lf` line endings in js files to avoid eslint errors

wozzo:
## Description

When the repository is downloaded to Windows if the `core.autocrlf` setting is set to `true` in the git global config then line endings are converted to `crlf` or `\r\n`.
When pushed back to the repo they are automatically converted back to `lf`, but on the Windows machine the eslint `linebreak-style` error (below) will be triggered on every line in every `*.js` file.

```
Expected linebreaks to be 'LF' but found 'CRLF'. eslint([linebreak-style](https://eslint.org/docs/rules/linebreak-style))
```

This is very noisy in editors that automatically show eslint errors, e.g.
image

Add a `.gitattributes` file which instructs git to keep the `lf` line endings even on windows.

## Checklist

- [ ] CHANGELOG entry

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+.js text eol=lf


wozzo:
Prematurely opened PR to the main repo. Needs testing in my fork first.

PR #859, opened 2023-03-02
Update breadcrumbs style

pflynny:
## Description
Add the updated breadcrumb that has been beta-tested on the live site; we've seen positive use on the new hub pages, and we're happy to apply it to the remainder of the site.

Some context for the update: 
- https://github.com/nhsuk/nhsuk-service-manual-community-backlog/issues/6

## Checklist

- [x] CHANGELOG entry


packages/components/breadcrumb/_breadcrumb.scss
```diff
@@ -15,11 +15,10 @@
  */
 
 .nhsuk-breadcrumb {
-  @include print-hide();
+  @include print-hide(); /* [1] */
```
andymantell, line 18: This [1] comment here would seem to relate to the bespoke spacing, and the [2] down below next to the padding relates  to the print hiding. i.e. the wrong way round.

I'm not entirely sure why we even do comments this way - it seems to completely separate them from the code they're referencing and make them really hard to read. Is it time to break away from this habit?

Also, the mixin is literally called `print-hide` - I don't even think a comment adds anything at all, we could remove that one.

andymantell: Taking the padding example - I see that the [1] can be put on two lines, and the comment relating to those lines can be "applied" to both - I guess that's the argument for this comment style? But if you were to just put that comment on one of the padding lines, or even above it like this:

```
// Bespoke spacing numbers used as there is no 12px spacing mapped in settings/spacing.
padding-bottom: nhsuk-spacing(3);
padding-top: 20px;
```

that would seem just as clear, and keeps the comment near the code?

Sorry to nitpick on this. It may be a change for another day. It was the numbers being the wrong way round that drew my attention to it! Maybe just sort the numbers for this change and we can bikeshed about the comments separately?

andymantell (changes requested):

andymantell (changes requested):
Needs a changelog entry

andymantell:
Couple of comments above. Otherwise seems reasonable.

fofr:
To summarise this change, is it:

- removing the white background from the navigation area
- removing the word "Back" from back links

?

packages/components/breadcrumb/_breadcrumb.scss
mcheung-nhs: Agree that [1] and [2] should be swapped.

Also, [1] message should read "Bespoke spacing number used as there is no 20px spacing mapped in settings/spacing."
Also, [6] and [7] comments need to be removed as there is no reference to them anymore.

And agree that discussions about comments should be left for another day ;-)

mcheung-nhs (changes requested):

packages/components/breadcrumb/_breadcrumb.scss
pflynny: Good catch, cheers @andymantell and @mcheung-nhs. I've updated the comments and the corresponding references.

I agree about the detachment from the comment and the code; it can be easy to overlook and sometimes confusing. Ha, the `print-hide` is overkill. Maybe we address these issues at the next frontend catchup?

pflynny:
> 

@fofr there is a reduction in the space between the breadcrumbs and the title of the page, but alongside what you stated, that is it.

andymantell (approved):

andymantell:
Approved @pflynny - but there's a merge conflict in the readme now. Sorry, that was me fiddling in the other PR 😬

packages/components/breadcrumb/template.njk
```diff
@@ -9,6 +9,6 @@
   {% endfor %}
       <li class="nhsuk-breadcrumb__item"><a class="nhsuk-breadcrumb__link" href="{{ params.href }}"{% for attribute, value in item.attributes %} {{attribute}}="{{value}}"{% endfor %}>{{ params.text}}</a></li>
     </ol>
-    <p class="nhsuk-breadcrumb__back"><a class="nhsuk-breadcrumb__backlink" href="{{ params.href }}">Back to {{ params.text }}</a></p>
+    <p class="nhsuk-breadcrumb__back"><a class="nhsuk-breadcrumb__backlink" href="{{ params.href }}"><span class="nhsuk-u-visually-hidden">Back to </span>{{ params.text }}</a></p>
```
mcheung-nhs, line 12: Do we need to have this as the same format as in the readme and changelog, rather than being on one line?
Also, missing nbsp in the hidden span?

mcheung-nhs (changes requested):

mcheung-nhs (approved):

PR #877, opened 2023-07-19
Add gitleaks pre-commit

the-nathan-smith:
## Description
Add gitleaks as a pre-commit to scan commits for secrets before they are commited. Resolves [TM-2721](https://nhsd-jira.digital.nhs.uk/browse/TM-2721).

## Checklist
- [ ] CHANGELOG entry

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "7.0.0",
   "description": "NHS.UK frontend contains the code you need to start building user interfaces for NHS websites and services.",
   "scripts": {
-    "prepare": "husky install",
+    "prepare": "gulp bundle",
     "build": "gulp build",
     "start": "gulp",
     "test": "npm run lint && npm run jest:ci",
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "7.0.0",
   "description": "NHS.UK frontend contains the code you need to start building user interfaces for NHS websites and services.",
   "scripts": {
-    "prepare": "gulp bundle",
+    "prepare": "husky install",
     "build": "gulp build",
     "start": "gulp",
     "test": "npm run lint && npm run jest:ci",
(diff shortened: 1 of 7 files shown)


roshaanbajwa (approved):
looks good to me

package.json
```diff
@@ -3,7 +3,7 @@
   "version": "7.0.0",
   "description": "NHS.UK frontend contains the code you need to start building user interfaces for NHS websites and services.",
   "scripts": {
-    "prepare": "gulp bundle",
+    "prepare": "husky install",
```
chrimesdev, line 6: If I remember how this works correctly, this will break the publishing of the nhsuk-frontend library package to the npm regsitry.

I think you're looking for `pre-push` further down`package.json`, not `prepare` for what you're trying to do.

the-nathan-smith: Hey @chrimesdev , thanks for spotting this.
I'll re-add the gulp-bundle.
Further down the file as you mention there is this snippet:
`  "husky": {
    "hooks": {
      "pre-push": "npm test"
    }
  },`
Does this want leaving as is or should I move the "husky install" in place of "npm test"?

chrimesdev: You don't need this `husky install` at all, the package `husky` will get installed by `npm install` as its listed as a dependency.

It seems you want to run `gitleaks` as part of your `pre-push`: https://github.com/gitleaks/gitleaks#pre-commit

I'm not sure why a user needs to install an `.exe` as documented in [README-gitleaks.md](https://github.com/nhsuk/nhsuk-frontend/pull/877/files#diff-8f0950db1689c6436897a8adc752c2139c48921d0cd3f57a8dbe0e6a8db28d67)?

Might be worth looking into this again.

the-nathan-smith: You're right that dockerfile is only for Gitpod workdspaces to download gitleaks. I presume this is okay?

I opted to use husky instead of pre-commit as pre-commit only works with the pip package manager. This way of running gitleaks as a precommit means the user only has to install gitleaks once and then it should work fine.

I'll add a line in the README about how to skip the pre-commit incase someone outside of the NHS doesn't want to install gitleaks on their machine?

Let me know what you think.

Issue #862, opened 2023-03-22
Action link example links to removed external page


will-dyson-nhsd:
# Bug Report

## What is the issue?

The **Action link** example links to an nhs.uk page that has been removed. Ideally I think this should be changed to use `href='#'`

## What steps are required to reproduce the issue?

Navigate to http://localhost:3000/components/action-link/index.html and click on the link.

PR #882, opened 2023-08-15
Remove dead link in action link example

roshaanbajwa:
## Description

Removed dead link in Action link example. (See https://github.com/nhsuk/nhsuk-frontend/issues/862)


sarawilcox:
Hi @roshaanbajwa, in fact, I think we have the same issue in the design system. Was it this link? https://service-manual.nhs.uk/design-system/components/action-link? If it is, would you mind doing the same for the design system, please? (Or is it linked to the frontend and automatically sorted when the frontend release goes out?)

the-nathan-smith (approved):

PR #898, opened 2023-10-06
Header drop down menu covers the top of the page contents

roshaanbajwa:
## Description

The header drop down menu covers the top of the content, as opposed to moving the content to start after the drop down menu.


the-nathan-smith (approved):

pflynny (approved):
LGTM!

Issue #901, opened 2023-11-08
Header container responds differently to width container


paulrobertlloyd:
# Bug Report

`.nhsuk-header__container` introduces different inline margin sizes at different breakpoints from that of `.nhsuk-width-container`. This means that at certain breakpoints, elements in the header don’t align with other elements on the page. 

## What is the issue?

Here are a few snapshots taken at different viewport widths

* Up to 640px, the containers share the same inline margin size (16px):
  
  <img alt="NHS header on 640px wide screen." src="https://github.com/nhsuk/nhsuk-frontend/assets/813383/c7cb983f-b7e8-4de5-848b-11cdaf2ae094" width="320">

* Between 641px and 768px, the header container has an inline margin of 32px while the width container sticks with 16px:

  <img alt="NHS header on 768px wide screen." src="https://github.com/nhsuk/nhsuk-frontend/assets/813383/aa244a55-a349-423e-ab62-d1f5d9103f6f" width="320">

* Between 769px and 989px, the contains again share the same inline margin size (32px):

  <img alt="NHS header on 800px wide screen." src="https://github.com/nhsuk/nhsuk-frontend/assets/813383/f53b3719-9c12-4bed-a19f-e71607558dca" width="320">

* Above 990px, the header container uses an inline margin of `auto` while the width container remains sticks with 32px:  

  <img alt="NHS header on 990px wide screen." src="https://github.com/nhsuk/nhsuk-frontend/assets/813383/6cdcae32-89e2-4df3-ac3e-b18e5bd5315d" width="320">

* Above 1024px, max-widths kick in and the containers are aligned again.

## What was the environment where this issue occurred?

* Device: Mac
* Operating System: macOS Sonoma
* Browser: Safari
* Browser version: 17.1
* NHS.UK frontend package version: 8.0.2

PR #902, opened 2023-11-13
Header and width containers margin

diana-nita:
## Checklist

- [ ] CHANGELOG entry


the-nathan-smith (approved):

andymantell (changes requested):
I assume this is in reference to #901? Is this the right way to fix this? Feels like maybe the bug should be fixed in the header component, and not by modifying the generic width container? Happy to be wrong - dismiss my review if this has already been discussed - just concerned that it feels like quite a sweeping change across the board, to fix a specific issue in one component.

andymantell (approved):

pflynny (approved):
LGTM 🙌

paulrobertlloyd:
Thanks all!

Issue #918, opened 2024-01-19
Footer copyright is not right aligned


ibrahimqasim-abtrace:
The copyright element of footers is no longer right aligned, and so does not align with the search bar, etc. of the header.

From looking into this, I believe this was removed in #881, but I can't find a changelog entry, and all the page templates still have the old style.

farwahkazmi1-nhs:
Issue resolved in: https://github.com/nhsuk/nhsuk-frontend/pull/919. Copyright footer now aligns to right in line with header.

PR #919, opened 2024-01-23
Align copyright in footer to the right

farwahkazmi1-nhs:
## Description
- Align the footer copyright to the right so it aligns with the search bar of the header.
- Allow links in footer to remain in a row when on a tablet (similar to desktop), instead of collapsing into a column (like on mobile).

Issue: https://github.com/nhsuk/nhsuk-frontend/issues/918


roshaanbajwa:
The backstop reference images will need to be approved for this one 
```
npm run backstop:test
npm run backstop:approve
```

diana-nita (approved):

PR #915, opened 2024-01-09
Update font and copyright guidance

roshaanbajwa:
## Description

Added font licensing guidance to the relevant package READMEs and updated the copyright guidance to include NHS England.

## Checklist

- [x] Tested against our [testing policy](https://github.com/nhsuk/nhsuk-frontend/blob/master/docs/contributing/testing.md) (Resolution, Browser & Accessibility)
- [x] Follows our [coding standards and style guide](https://github.com/nhsuk/nhsuk-frontend/blob/master/docs/contributing/coding-standards.md)
- [x] CHANGELOG entry

PR #940, opened 2024-04-27
Update footer copyright statement

frankieroberto:
The copyright guidance was changed in #915 to be copyright of NHS England instead of Crown Copyright, but the text in the footer was only partially updated.

This updates the second instance of Crown Copyright in the footer to also be NHS England.  (The text is in 2 places in the template, and which one is used depends on how many columns of links are included).

This updates the [footer with navigation links example](https://service-manual.nhs.uk/design-system/components/footer#footer-with-navigation-links) to match the guidance on the page.


sarawilcox:
Hi @frankieroberto, in this case Crown copyright is correct. It's because it's the NHS website example and the NHS website is Crown copyright, unlike most other NHS products.

frankieroberto:
> Hi @frankieroberto, in this case Crown copyright is correct. It's because it's the NHS website example and the NHS website is Crown copyright, unlike most other NHS products.

Hi @sarawilcox! Thanks. I should have been clearer - I think the issue is that in the footer component in nhs-frontend, whether it displays "Crown Copyright" or "Copyright NHS England" seems to depend on the number of columns used in the footer. Not sure if this is deliberate or an oversight, but it confused me!

roshaanbajwa:
> > Hi @frankieroberto, in this case Crown copyright is correct. It's because it's the NHS website example and the NHS website is Crown copyright, unlike most other NHS products.
> 
> Hi @sarawilcox! Thanks. I should have been clearer - I think the issue is that in the footer component in nhs-frontend, whether it displays "Crown Copyright" or "Copyright NHS England" seems to depend on the number of columns used in the footer. Not sure if this is deliberate or an oversight, but it confused me!

Hi Frankie, 

The copyright doesn't depend on the number of columns, the component "Footer (columns)" is the implementation that is included in the live site, so we chose to use it as the example of footer with columns with the copyright included.

edwardhorsford:
I agree this looks like a bug to me. Copyright text is hardcoded to *either* NHS England or Crown copyright, and the thing that decides between them is whether you've provided `linksColumn2`.  I don't see why providing `linksColumn2` should make it Crown copyright, yet NHS England without.

Footer with a single link:
Screenshot 2024-06-12 at 12 03 24

Footer with a single linkColumn2 in addition:
Screenshot 2024-06-12 at 12 07 29



Aside: ideally these messages would be configurable, not hardcoded. I think this component could be more flexible.

paulrobertlloyd:
> Copyright text is hardcoded to either NHS England or Crown copyright, and the thing that decides between them is whether you've provided `linksColumn2`. 

Ooooh, this is why the copyright changed in my prototype; it was saying NHS England, then it changed to Crown Copyright. When I first looked at this codebase, I only spotted one of the 2 lines of code where a copyright statement is provided.

Until this issue is resolved, in my prototype I’ve added this fix:

```njk
{{ footer({
  …
}) | replace("&copy; Crown copyright", "&copy; NHS England") }}
```

I agree that it should be configurable; perhaps with the default/fallback value being `© Crown Copyright`.

Reading the Footer guidance in the Service Manual, it looks like this was either once possible, or was intended to be possible but not implemented:

* Nunjucks macro options include the following option:

  > | Name | Type | Required | Description |
  > | - | - | - | - |
  > | copyright | string | false | Optional text for the copyright notice in the footer. |

* Guidance later on the page states:

  > ### Copyright
  >
  > Your copyright statement must reflect the ownership of your website or service. The [NHS website](https://www.nhs.uk/) is (c) Crown copyright but the NHS generally does not have Crown status. The service manual footer, for example, says (c) NHS England.

frankieroberto:
@paulrobertlloyd well spotted! Looks like the documentation for this component on the NHS Service Manual doesn’t quite match the Nunjucks macro in nhsuk-frontend.

I’ve opened a quick PR to fix this: #975

frankieroberto:
Closing this now, as #975 makes it redundant.

PR #946, opened 2024-05-13
Update label margins to match fieldsets and align with GOV.UK

### LINKED ISSUE #947: Label bottom margins are too large
# Bug Report

The bottom margins on the label modifiers are too large. On GOV.UK they're set to align with the fieldset legend bottom margins - but on the NHS design system they use much larger responsive margins. The result is that the label looks too visually distant from the input or hint.

You can see the [discrepancy on this page](https://service-manual.nhs.uk/design-system/patterns/ask-users-for-their-nhs-number) when you compare fieldset legends to labels - they should all look approximately the same distance from the form fields.

**XL label**
xl-before

**L label**
l-before

**M label**
m-before

**S label**
s-before

[Here's GOV.UK's code for labels.](https://github.com/alphagov/govuk-frontend/blob/6c9d8d7191edb18e8a46d2a383e311f397367946/packages/govuk-frontend/src/govuk/components/label/_index.scss)

edwardhorsford:
## Description
This updates the bottom margins of labels to better match those of legends. I've followed a bit of how GOV.UK structures these.

At the moment I think the label margins are overly large - too separated from the hint or inputs they relate to. GOV.UK doesn't have responsive margins for these - the label margins are all fixed and the same as the fieldset legends.

The only label that doesn't now match fieldset legends is the small size - where it follows the no modifier size 1.

**XL label before**
xl-before

**XL label after**
xl-after

**L label before**
l-before

**L label after**
l-after

**M label before**
m-before

**M label after**
m-after

**S label before**
s-before

**S label after**
s-after

**No modifier label before**
no-modifier-before

**No modifier label after**
no-modifier-after

Fixes #947 

## Checklist

- [x] CHANGELOG entry


anandamaryon1 (approved):
Sensible change, helps to group labels with their inputs and now matches fieldset legends. Looks good thank you.

anandamaryon1 (approved):
Good catch with the backstop reference images

PR #951, opened 2024-05-17
Fix inset-text not supporting html param

### LINKED ISSUE #950: Bug: inset text requires html option to be uppercase
# Bug Report

The HTML option for inset text requires it to be in uppercase. Lowercase `html` does not render. I think this is a mistake as all other Nunjucks arguments I have seen are lowercase.

Screenshot 2024-05-17 at 14 19 43

PR incoming.

edwardhorsford:
## Description

Fixes #950. I think we expect all Nunjucks arguments to be lowercase - but here is is uppercase.

I think it would be a breaking change to swap it over, so I've added support for `params.html` in preference whilst still accepting `params.HTML`. It could potentially be considered cleaning up for a future major version release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - Removing classes from icon card that are not doing anything
 - Align label bottom margins with fieldset legend bottom margins ([PR 946](https://github.com/nhsuk/nhsuk-frontend/pull/946)).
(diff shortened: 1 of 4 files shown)


frankieroberto (dismissed):
Nice.

As a follow-up, is it worth having a doc somewhere outlining naming conventions for the Nunjucks params (if it doesn't already exist)?

ie use cameCase for multiple words, and lowercase for abbreviations?

PR #954, opened 2024-05-24
Ensure that headers are aligned to expected standards

roshaanbajwa:
## Description
https://nhsd-jira.digital.nhs.uk/browse/TM-3235

## Checklist

### DIFF
--- a/packages/components/header/template.njk
+++ b/packages/components/header/template.njk
@@ -78,7 +78,7 @@
     </div>{# close nhsuk-header__container #}
     <div class="nhsuk-navigation-container">
       <nav class="nhsuk-navigation" id="header-navigation" role="navigation" aria-label="Primary navigation">
-        <ul class="{{ "nhsuk-header__navigation-list nhsuk-header__navigation-list--left-aligned" if params.primaryLinks.length < 4 else "nhsuk-header__navigation-list" }}">
+        <ul class="nhsuk-header__navigation-list {%- if params.primaryLinks.length < 4 %} nhsuk-header__navigation-list--left-aligned{% endif %}">
           {%- for item in params.primaryLinks %}
           <li class="nhsuk-header__navigation-item">
             <a class="nhsuk-header__navigation-link"  href="{{item.url}}">
@@ -136,7 +136,7 @@
     
     <div class="nhsuk-navigation-container">
       <nav class="nhsuk-navigation" id="header-navigation" role="navigation" aria-label="Primary navigation">
-        <ul class="nhsuk-header__navigation-list{{ ' nhsuk-header__navigation-list--left-aligned' if params.primaryLinks.length < 4 }}">
+        <ul class="nhsuk-header__navigation-list {%- if params.primaryLinks.length < 4 %} nhsuk-header__navigation-list--left-aligned{% endif %}">
           {%- for item in params.primaryLinks %}
           <li class="nhsuk-header__navigation-item">
             <a class="nhsuk-header__navigation-link"  href="{{item.url}}">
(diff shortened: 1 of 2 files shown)


packages/components/header/template.njk
```diff
@@ -78,7 +78,7 @@
     </div>{# close nhsuk-header__container #}
     <div class="nhsuk-navigation-container">
       <nav class="nhsuk-navigation" id="header-navigation" role="navigation" aria-label="Primary navigation">
-        <ul class="nhsuk-header__navigation-list">
+        <ul class="{{ "nhsuk-header__navigation-list nhsuk-header__navigation-list-fewer-links" if params.primaryLinks.length < 4 else "nhsuk-header__navigation-list" }}">
```
frankieroberto, line 81: Rather than adding this class automatically based on number of links, maybe it’d make sense to have it as a class that can be optionally added? (Or make `—-justify` the optional one?)

I can see there being some scenarios where you might want left-alignment even with 4 or 5 links? 🤔

edwardhorsford: Might be worth a discussion on best approach?

If doing it automatically I'd be tempted to apply a class like `nhsuk-header_navigation-list--justified` when there are over x links. So the class sets the appearance, it doesn't describe the links.

roshaanbajwa: Hi Frankie, I’ve spoken to service manual interaction designers who agree with this method of having a hard coded limit of navigation links as it's an issue that's been raised a few times in the design system issues page - could you add your suggestion as a separate issue in this repo? It would be good to feed it back to the designers

packages/components/header/template.njk
```diff
@@ -136,7 +136,7 @@
     
     <div class="nhsuk-navigation-container">
       <nav class="nhsuk-navigation" id="header-navigation" role="navigation" aria-label="Primary navigation">
-        <ul class="nhsuk-header__navigation-list">
+        <ul class="{{ "nhsuk-header__navigation-list nhsuk-header__navigation-list--left-aligned" if params.primaryLinks.length < 4 else "nhsuk-header__navigation-list" }}">
```
edwardhorsford, line 139: I wonder if this could be refactored so `nhsuk-header__navigation-list` was always included, and it's just the second class that's conditional?

edwardhorsford (changes requested):
One thing to make consistent and two nitpicks.

packages/components/header/template.njk
```diff
@@ -78,7 +78,7 @@
     </div>{# close nhsuk-header__container #}
     <div class="nhsuk-navigation-container">
       <nav class="nhsuk-navigation" id="header-navigation" role="navigation" aria-label="Primary navigation">
-        <ul class="nhsuk-header__navigation-list">
+        <ul class="{{ "nhsuk-header__navigation-list nhsuk-header__navigation-list--left-aligned" if params.primaryLinks.length < 4 else "nhsuk-header__navigation-list" }}">
```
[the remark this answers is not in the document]
roshaanbajwa: Resolved with last commit

packages/components/header/template.njk
```diff
@@ -136,7 +136,7 @@
     
     <div class="nhsuk-navigation-container">
       <nav class="nhsuk-navigation" id="header-navigation" role="navigation" aria-label="Primary navigation">
-        <ul class="nhsuk-header__navigation-list">
+        <ul class="nhsuk-header__navigation-list{{ ' nhsuk-header__navigation-list--left-aligned' if params.primaryLinks.length < 4 }}">
```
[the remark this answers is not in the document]
roshaanbajwa: Resolved with last commit

edwardhorsford (approved):

PR #960, opened 2024-06-08
Announce whitespace in screen reader announcements of visually hidden text

frankieroberto:
This fixes a bug in the nhsuk-u-visually-hidden helper class where any whitespace surrounding visually hidden text isn't announced by screen readers.

This fix was added to `govuk-frontend` in [pull request #3868](https://github.com/alphagov/govuk-frontend/pull/3836) and released in [v4.7.0](https://github.com/alphagov/govuk-frontend/releases/tag/v4.7.0) and has been re-applied here.

Thanks to @hannalaakso for the work upstream. 🙌 

I’ve also followed the approach of the GOV.UK Design System team of adding a Typography page with examples of using visually-hidden text, to make testing easier.

### DIFF
--- a/packages/core/tools/_mixins.scss
+++ b/packages/core/tools/_mixins.scss
@@ -38,6 +38,18 @@
 //
 
 @mixin visually-hidden() {
+  // Absolute positioning has the unintended consequence of removing any
+  // whitespace surrounding visually hidden text from the accessibility tree.
+  // Insert a space character before and after visually hidden text to separate
+  // it from any visible text surrounding it.
+  &::before {
+    content: "\00a0";
+  }
+
+  &::after {
+    content: "\00a0";
+  }
+
   border: 0;
   clip: rect(0 0 0 0);
   -webkit-clip-path: inset(50%);
--- a/packages/components/tables/_tables.scss
+++ b/packages/components/tables/_tables.scss
@@ -99,6 +99,11 @@
 
     @include mq($from: desktop) {
       @include visually-shown(table-header-group); /* [2] */
+
+      &:before,
+      &:after {
+        content: "";
+      }
     }
   }
 


packages/core/tools/_mixins.scss
```diff
@@ -38,6 +38,20 @@
 //
 
 @mixin visually-hidden() {
+  position: absolute;
```
mcheung-nhs, line 41: A minor pendantic point but to follow our usual convention, you might want to do is move this to underneath the padding declaration so it's all in alphabetical order and then move the comment outside the mixin.

mcheung-nhs (dismissed):

packages/core/tools/_mixins.scss
```diff
@@ -38,6 +38,20 @@
...
+
+  // Absolute positioning has the unintended consequence of removing any
+  // whitespace surrounding visually hidden text from the accessibility tree.
+  // Insert a space character before and after visually hidden text to separate
+  // it from any visible text surrounding it.
```
edwardhorsford, line 46: Does this mean that the examples / usage don't need any leading or trialing spaces in the markup?

frankieroberto: I guess it’t no longer technically needed, but probably still makes sense to leave it in as it makes the HTML more readable?

eg 

```html
<span class="nhsuk-u-visually-hidden">Countries beginning with</span>A
``` 

is not super clear vs 

```html
<span class="nhsuk-u-visually-hidden">Countries beginning with</span> A
```

?

edwardhorsford: My loosely held thought is that the latter implies it's _necessary_, whereas it now seems like it isn't - but then perhaps something to address every time this class is used.

frankieroberto: It’s possibly still good practice to include the space anyway actually, for any modes where the CSS isn’t applied, like possibly some 'Reader' modes or print layouts?

anandamaryon1:
Tested on VoiceOver and I can hear it working when we have a space inside a visually hidden span 👍 

Would like to remove the 'Styles' example testing pages before merging though.

app/styles/typography.njk
anandamaryon1, line 1: Remove before merge

app/pages/examples.njk
```diff
@@ -132,6 +132,7 @@
 
   <h2>Styles</h2>
```
anandamaryon1, line 133: Remove before merge

app/styles/typography.njk
frankieroberto: Will do. I copied this from the govuk example page, but maybe it's overkill here?

app/pages/examples.njk
frankieroberto: Done

frankieroberto:
@anandamaryon1 @edwardhorsford I’ve removed the test pages from the examples so this is now just the CSS changes (and I’ve squashed all the commits into 1).

Think this can now be merged?!

anandamaryon1:
Backstop to the rescue… caught something… 

Visually hidden change has affected the table head styling, showing extra white space elements:

image

anandamaryon1 (approved):

PR #971, opened 2024-06-18
Let page width be overridden

edwardhorsford:
## Description
Updates `$nhsuk-page-width` to be a SASS default, which would let teams override it. Nearly every other item in `globals.scss` can be overridden already. This works the same way as GOV.UK for setting page width.

## Checklist
- [x] CHANGELOG entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # NHS.UK frontend Changelog
 
+## Unreleased
+
+:new: **New features**
+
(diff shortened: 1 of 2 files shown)


frankieroberto (approved):
Seems sensible to me. I’ve tended to use modifier classes like `app-width-container--wide` to override the page width, but that’s partly because I forget the sass variables exist.

Modifier classes are still useful if you have 2 different page widths in your service (rare, but can happen!).

PR #974, opened 2024-06-21
Draft: Move breadcrumb inside nhs-width-container

frankieroberto:
This is a quick proof-of-concept showing how the Breadcrumb component might move inside the `<div class="nhs-width-container">` to avoid having to define its own, and for consistency with the Back component.

This would be a breaking change as the HTML is different, and teams would have to update where the breadcrumb appears.

I've also switched the padding to be margin, for consistency with the Back link. The actual spacing values we use are tbc.

I've also added the `outerContent` block to the template used by the test app in the repo.

See [Review back link and breadcrumb spacing and positioning ](https://github.com/nhsuk/nhsuk-service-manual-community-backlog/issues/526)

### DIFF
--- a/app/_templates/page.njk
+++ b/app/_templates/page.njk
@@ -48,9 +48,8 @@
 
 {% block body %}
 
-  {% block breadcrumb %}{% endblock %}
-
   <div class="nhsuk-width-container">
+    {% block outerContent %}{% endblock %}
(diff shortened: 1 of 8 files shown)


frankieroberto:
Closing as this has been merged into #1002

PR #975, opened 2024-06-26
Make copyright statement settable via params

frankieroberto:
This de-duplicates 2 areas where the copyright statement is set (depending on the number of columns of links), and also allows the copyright text to be set using `params.copyright` on the Nunjucks macro.

This was previously documented as an option on the [footer component](https://service-manual.nhs.uk/design-system/components/footer) but wasn’t implemented in the actual nunjucks macro.

> Screenshot showing the text: 'copyright string false Optional text for the copyright notice in the footer'

I’ve set the default to "NHS England" as the guidance suggests that’s what most services should use, but the NHS.UK website, or any local NHS trusts or other organisations, could override this.


frankieroberto:
Possibly this should follow the pattern of having `copyright.text` and `copyright.html` instead, but the labels for the links in the footer don’t seem to do that so I wasn’t sure how rigidly we wanted to follow the pattern? 🤷‍♂️

roshaanbajwa (approved):

PR #976, opened 2024-06-27
Warning button variant

paulrobertlloyd:
## Description

On our staff facing service, we use interruption screens that warn a user before they proceed to undertake an action that is destructive and/or cannot be undone. On these pages, we use also a red warning button to further indicate that a destructive action will take place:

Screenshot of a screen asking if a user wants to delete a vaccine.

This follows [a similar pattern from the GOV.UK Design System](https://design-system.service.gov.uk/components/button/#warning-buttons):

> ### Warning buttons
>
> Warning buttons are designed to make users think carefully before they use them. They only work if used very sparingly. Most services should not need one.

This PR proposes adding an equivalent `--warning` variant to `nhsuk-button`. In the corresponding Service Manual guidance for this component, we can start by using the same guidance as that used in the GOV.UK Design System.

## Technical note

The SCSS for each button variant currently includes a lot of repeated properties that shouldn’t need redeclaring given each modifier inherits from the base class. That’s a separate issue, so for this PR, I’ve maintained the same implementation and redeclared properties for this variant.

## Checklist

- [x] CHANGELOG entry


frankieroberto (approved):
Looks good to me, and this would be useful in our service too.

anandamaryon1:
Thanks @paulrobertlloyd this helpful. We'll need to hold this until we can add the guidance to the design system: https://github.com/nhsuk/nhsuk-service-manual/issues/1995

Should be able to [use the gov.uk guidance](https://design-system.service.gov.uk/components/button#warning-buttons), I don't think there's anything else specific we'd need to add/amend?

frankieroberto:
Guidance added in this PR: https://github.com/nhsuk/nhsuk-service-manual/pull/1999

@anandamaryon1 I think this is un-blocked now?

anandamaryon1:
Ah, one small thing, do you mind if I/we change the example button label to align with the design system documentation? 
@paulrobertlloyd @frankieroberto 

(It's something we want to do moving forward, to ensure both are aligned to avoid confusion).

frankieroberto:
@anandamaryon1 ah do you mean change from "Delete account" to "Yes, delete this vaccine"?

app/components/button/warning.njk
```diff
@@ -0,0 +1,22 @@
...
+
+{% block body %}
+
+  <div class="nhsuk-width-container">
+    <main class="nhsuk-main-wrapper" id="maincontent">
+
+      <div class="nhsuk-grid-row">
+        <div class="nhsuk-grid-column-two-thirds">
+          {{ button({
+            "text": "Delete account",
```
frankieroberto, line 14:
```suggestion
            "text": "Yes, delete this vaccine",
```

packages/components/button/README.md
```diff
@@ -130,6 +130,31 @@ Find out more about the button component and when to use it in the [NHS digital
...
+</button>
+```
+
+#### Nunjucks macro
+
+```
+{% from 'components/button/macro.njk' import button %}
+
+{{ button({
+  "text": "Delete account",
```
frankieroberto, line 151:
```suggestion
  "text": "Yes, delete this vaccine",
```

anandamaryon1 (approved):

frankieroberto (approved):

frankieroberto (approved):

frankieroberto:
Merged, thanks @paulrobertlloyd and @anandamaryon1!

PR #978, opened 2024-07-05
Support classes and attributes in header primary links

frankieroberto:
## Description

This adds support for adding optional custom CSS classes or other HTML attributes to the list item tags of the primary links within the Header component.

This gives the header navigation a bit more flexibility to be customised by services, who may wish to do things like highlighting the current section, having some items left-aligned and some right aligned, colour-coding the sections, adding data attributes for testing purposes, or other things we’ve not yet thought of. 

The options are enabled by supporting `classes` and `attributes` as params within the `primaryLinks` objects in the Nunjucks macro, like this:

```nunjucks
{{ header({
  "showNav": "true",
  "showSearch": "false",
  "primaryLinks": [
    {
      "url"  : "https://www.nhs.uk/conditions",
      "label" : "Health A-Z",
      "classes": "app-header__navigation-item--current",
      "attributes": {
        "aria-current": "true"
      }
    },
    {
      "url" : "https://www.nhs.uk/live-well/",
      "label" : "Live Well"
    }
  ]
}) }}
```

## Checklist

- [x] CHANGELOG entry


paulrobertlloyd (dismissed):
This would be super helpful, using all kinds of hacks at the moment to try and style individual navigation items.

Might also be a good idea to bring in the `attributes` macro used by GOV.UK Frontend (and added in #969) via a separate PR so that boolean attributes are supported properly.

edwardhorsford:
> This would be super helpful, using all kinds of hacks at the moment to try and style individual navigation items.
> 
> Might also be a good idea to bring in the `attributes` macro used by GOV.UK Frontend (and added in #969) via a separate PR so that boolean attributes are supported properly.

@paulrobertlloyd I actually brought it in as part of #969 - though I now think it's probably better to bring it in as a separate pr sooner so that it can apply to other components and not be blocked by task list.

frankieroberto:
@edwardhorsford nice - if you can pull that into a separate PR and get that merged, I can use the macro in the one.

anandamaryon1 (approved):

PR #979, opened 2024-07-05
Remove .nhsuk-u-top-and-bottom utility class

frankieroberto:
I spotted that the `.nhsuk-u-top-and-bottom` utility class isn't actually included within the build as it was removed from `core/all.scss` in 2018 (https://github.com/nhsuk/nhsuk-frontend/commit/df303c1f342f0365707234c9e1b6f4c7feab6b51).

From the looks of it, the `top-and-bottom()` mixin was likely only intended for use within other components, and I can’t see any evidence of it having been used as a standalone utility class (see [GitHub search](https://github.com/search?q=nhsuk-u-top-and-bottom&type=code).

So I think this is safe to remove?

## Checklist

- [x] CHANGELOG entry


sarawilcox:
Hi @roshaanbajwa, were you going to check this one?

roshaanbajwa (approved):

PR #980, opened 2024-07-09
Fix changelog: Move bugfix entry to the right spot

frankieroberto:
Fix: the changelog entry in #979 was added to the previous release rather than the unreleased section.


frankieroberto:
@roshaanbajwa sorry, messed up the changelog entry in the previous PR. 🤦

edwardhorsford (approved):

PR #986, opened 2024-07-17
Remove hardcoded Home link from navigation

frankieroberto:
Removes the link to `"/"` labelled `"Home"`, which is currently hardcoded, and only shows up within the navigation menu on mobile screen widths.

This link may not be appropriate for all services, which might not have a homepage, or might use a different path for it.

It is also unclear whether having a homepage link is needed in the navigation if the NHS logo also goes to the homepage, and if it is needed, whether it should always be present rather than only within the mobile menu.

Services who want to continue to include the Home link can do so by adding an additional class using the Header component like this:

```njk
{{ header({
  "showNav": "true",
  "showSearch": "false",
  "primaryLinks": [
    {
      "url"  : "/conditions",
      "label" : "Health A-Z"
    },
    { ... },
    {
      "url" : "/",
      "label" : "Home",
      "classes": "nhsuk-header__navigation-item--home"
    }
  ]
}) }}
```

This should probably be considered a breaking change, as services would need to update their HTML if they want to keep the Home link?

## Checklist

- [ ] CHANGELOG entry


anandamaryon1:
merged into https://github.com/nhsuk/nhsuk-frontend/pull/1058

Issue #921, opened 2024-02-05
Input prefix/suffix overflows containers with large font sizes


pufferfish101007:
# Bug Report

## What is the issue?

On an input with a prefix and/or suffix, eg http://nhsuk.github.io/nhsuk-frontend/components/input/prefix-and-suffix.html, the pre/suffix should remain within its bounding box. However, when the font size is increased, the text escapes out the bottom of the box, like so:
![image](https://github.com/nhsuk/nhsuk-frontend/assets/50246616/69a8eac2-76d2-4ad1-9a62-c45c3046a343)

## What steps are required to reproduce the issue?

1. Go to http://nhsuk.github.io/nhsuk-frontend/components/input/prefix-and-suffix.html
2. Set your browser font size to something quite large (>=28 worked for me)
3. Observe the overflow

## What was the environment where this issue occurred?

* Device: Asus vivobook
* Operating System: Windows 11
* Browser: Chrome
* Browser version: 120
* NHS.UK frontend package version: latest off of master
* Node version: 21.6.1
* npm version: 10.2.4

## Is there anything else you think would be useful in recreating the issue?

If the `height: 40px` style is removed from both `.nhsuk-input` and `.nhsuk-input__prefix, .nhsuk-input__suffix`, then the issue is resolved, and also prevents there being a silly amount of space around the input text with smaller font sizes. Not sure if that's the way to go; I don't know if there was a particular reason for specifying `40px` as the height or not.
https://github.com/pufferfish101007/nhsuk-frontend/tree/921-overflowing-prefix demonstrates the fix.

pufferfish101007:
There is a similar issue in the select component:
![image](https://github.com/nhsuk/nhsuk-frontend/assets/50246616/7404acab-3baf-42d7-95cc-ffb4db3ee689)

It looks like this 40px height was introduced in the govuk-frontend repo, back in 2017 as far as I can tell (https://github.com/alphagov/govuk-frontend/commit/71016217382c97bc9ae562b8105d8311fac83788#diff-e6215fe95ffd15e29e98d761b90a6b21f83c395104791874b7ab17b286e2f3ee), with no reasoning... so it's probably ok to remove?

pufferfish101007:
closing since fixed by c3fd643f363160d94adf9d0173940b4258d17310

PR #987, opened 2024-07-18
Update height and set minimum width of Selects

frankieroberto:
This fixes a minor bug with the select component in Safari, which is that the `min-height` property is ignored when the default native appearance is used, resulting in a select which is only 30px high. (Other browsers such as Chrome and Firefox seem to apply the minimum height.)

This restores the select to its previous fixed height of 40px. As the select will never wrap onto 2 or more lines (instead the text gets truncated), having a fixed height should be fine.

Additionally, this also introduces a minimum width to the select component (this property does work in all browsers), in case the options in the select are very short. In these cases, having a wider select means for a bigger target area, and is closer to what users expect. I’ve followed the GOV.UK Design System in setting the minimum width to match the `width-10` variant of the input component, which works out as 223px (see [their definition of this](https://github.com/alphagov/govuk-frontend/blob/main/packages/govuk-frontend/src/govuk/components/select/_index.scss#L11C5-L15C23)).

## Screenshots (Safari)

| Before | After |
|--------|-------|
| Screenshot 2024-07-18 at 08 55 11 | Screenshot 2024-07-18 at 08 55 28 |
 | Screenshot 2024-07-18 at 09 00 03 | Screenshot 2024-07-18 at 08 59 47 |

## Checklist

- [ ] CHANGELOG entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 - Change "Contact us" in the footer link examples to "Give us feedback" ([PR 972](https://github.com/nhsuk/nhsuk-frontend/pull/972))
 - Remove the pattern from the date input component
+- Fix height of select component in Safari ([PR 987](https://github.com/nhsuk/nhsuk-frontend/pull/987))
+- Set minimum width of select component ([PR 987](https://github.com/nhsuk/nhsuk-frontend/pull/987))
 
 :new: **New features**
 
--- a/packages/components/select/_select.scss
+++ b/packages/components/select/_select.scss
@@ -8,6 +8,11 @@
   border: $nhsuk-border-width-form-element solid $nhsuk-form-border-color;
   box-sizing: border-box;
   height: 40px;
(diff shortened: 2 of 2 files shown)


anandamaryon1:
@frankieroberto Looked into this one to try and fix your safari issue but also fix the initial issue that introduced this bug: https://github.com/nhsuk/nhsuk-frontend/issues/921

Swapped to set the height in rems, seems to work. Sadly can't take credit for the idea – it's how GOV have sized theirs. 

(also refactored your comments to follow the style guide)

frankieroberto:
@anandamaryon1 oh nice! I didn't use rems as we didn't seem to use rems elsewhere, but it will also make it scale with the text size. 🙌 

I _think_ the Safari bug is more down to setting `min-height` (which seems to be ignored) rather than the unit type, but not 100% sure.

anandamaryon1 (approved):

PR #996, opened 2024-08-19
Simplify template logic for header component

paulrobertlloyd:
## Description

Prior to any work on adding account links to the header, or refining its design, this PR aims to reduce the complexity of the component template, and remove repetition of navigation and search markup.

This PR:

- Shows or hides navigation and search markup by querying the presence of values for `params.primaryLinks` and `params.search` respectively
- Removes repetition of navigation and search markup thanks to simplified logic and fewer variables
- Uses the [`default` filter](https://mozilla.github.io/nunjucks/templating.html#default-value-default-boolean) for values as opposed to longer `if`/`else` tags
- Adds an option to set the (visually hidden) label for the search input
- Allows a transactional service name to be used alongside navigation and/or search
- Changes the default search query param from `search-field` to `q`, to match that used by the NHS.UK website (all the other fallback values do this, so this should as well)

As this PR stands, this is a breaking change, although only in terms of the Nunjucks macro options. Here are the changes:

| Option | Change | Note |
|-|-|-|
| `showNav` | **Removed** | Navigation shown if `primaryLinks` option provided |
| `showSearch` | **Removed** | Search shown if `search` option provided |
| `transactional` | **Removed** | Transactional header used if `transactionService` option provided |
| `search` | **New** | Used to collate all search related options |
| `search.placeholder` | **New option** | Change the search placeholder text. Defaults to ‘Search’ |
| `search.visuallyHiddenButton` | **New** | Change the (visually hidden) search button text. Defaults to ‘Search’ |
| `search.visuallyHiddenLabel` | **New** | Change the (visually hidden) search label. Defaults to ‘Search the NHS website’ |
| `searchAction` | **Renamed** | Becomes `search.action`. Defaults to `https://www.nhs.uk/search/` |
| `searchInputName` | **Renamed** | Becomes `search.name`. Defaults to `q` |

It should be noted that the service manual website, as just one example, currently has the label ‘Search the NHS website’, when in fact you can only search the service manual website. The new `search{}.visuallyHiddenLabel` option would make it possible to use a correct label.

If this PR is accepted, a separate PR can be made against the service manual to update these options.

There’s a whole load of other refinements that could be made in terms of markup, styles and the different options provided, but I’ve tried to keep this change tightly scoped to focus only on reducing the complexity of this component’s templating logic, with the side effect being a reduction in the number of interconnected macro options provided to authors.

## Why remove `showNav`, `showSearch` and `transactional` options?

While it could be suggested that these options make it easy to show/hide parts of the header, this can be achieved with Nunjucks in other ways. For example, currently you might do the following:

```njk
{# If `user.loggedIn` is `true`, navigation is shown #}
{{ header({
  service: {
    name: "Example"
  },
  showNav: user.loggedIn,
  primaryLinks: [{
    url : "#",
    label : "Item one"
  }, {
    url : "#",
    label : "Item two"
  }]
}) }}
```

This could be written thus:

```njk
{# When `user.loggedIn` is `true`, navigation is shown #}
{{ header({
  service: {
    name: "Example"
  },
  primaryLinks: [{
    url : "#",
    label : "Item one"
  }, {
    url : "#",
    label : "Item two"
  }] if user.loggedIn
}) }}
```

It can also get tricky and complex if you have to manage the state of the header across multiple options. The changes in this PR hopefully make it easier to manage this.

## Checklist

- [x] CHANGELOG entry

### DIFF
--- a/app/components/all.njk
+++ b/app/components/all.njk
@@ -36,8 +36,7 @@
   }) }}
 
   {{ header({
-      "showNav": "true",
-      "showSearch": "true",
+      "search": true,
       "primaryLinks": [
         {
           "url"  : "https://www.nhs.uk/conditions",
--- a/packages/components/header/template.njk
+++ b/packages/components/header/template.njk
@@ -1,8 +1,10 @@
 {% from "../../macros/attributes.njk" import nhsukAttributes %}
 
-{# Define some defaults #}
-{% set showNav = params.showNav if params.showNav else "false" %}
-{% set showSearch = params.showSearch if params.showSearch else "false" %}
+{%- set searchAction = params.search.action | default('https://www.nhs.uk/search/') %}
+{%- set searchName = params.search.name | default('q') %}
+{%- set searchPlaceholder = params.search.placeholder | default('Search') %}
+{%- set searchVisuallyHiddenButton = params.search.visuallyHiddenButton | default('Search') %}
+{%- set searchVisuallyHiddenLabel = params.search.visuallyHiddenLabel | default('Search the NHS website') %}
 
 {% set nhsLogo %}
 <svg class="nhsuk-logo" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 200 80" height="40" width="100">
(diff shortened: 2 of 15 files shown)


packages/components/header/template.njk
```diff
@@ -1,167 +1,87 @@
...
-              <span class="nhsuk-u-visually-hidden">Browse</span>
-              More
-              <svg class="nhsuk-icon nhsuk-icon__chevron-down" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" aria-hidden="true" focusable="false">
-                <path d="M15.5 12a1 1 0 0 1-.29.71l-5 5a1 1 0 0 1-1.42-1.42l4.3-4.29-4.3-4.29a1 1 0 0 1 1.42-1.42l5 5a1 1 0 0 1 .29.71z"></path>
-              </svg>
-            </button>
-          </li>
-        </ul>
-      </nav>
+{%- if params.transactionalService and not params.primaryLinks and not params.search %}
```
frankieroberto, line 37:
```suggestion
{%- if params.transactionalService %}
```

I think this should just check for the presence of `transactionalService`? This would fix an existing bug which is that currently if you try and have a transactional service name AND navigation items, only the navigation items appear but not the service name.

packages/components/header/template.njk
```diff
@@ -1,167 +1,87 @@
...
-              <svg class="nhsuk-icon nhsuk-icon__chevron-down" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" aria-hidden="true" focusable="false">
-                <path d="M15.5 12a1 1 0 0 1-.29.71l-5 5a1 1 0 0 1-1.42-1.42l4.3-4.29-4.3-4.29a1 1 0 0 1 1.42-1.42l5 5a1 1 0 0 1 .29.71z"></path>
+{%- endif %}
+
+{%- if params.search %}
+    <div class="nhsuk-header__content" id="content-header">
+      <div class="nhsuk-header__search {%- if not params.primaryLinks %} nhsuk-header__search-no-nav{% endif %}">
+        <div class="nhsuk-header__search-wrap" id="wrap-search">
+          <form class="nhsuk-header__search-form" id="search" action="{{ params.search.action | default('https://www.nhs.uk/search/') }}" method="get" role="search">
+            <label class="nhsuk-u-visually-hidden" for="search-field">{{ params.search.label | default('Search the NHS website') }}</label>
```
frankieroberto, line 48: This isn’t done consistently elsewhere, but given this is a new param option, should we follow the convention of having `params.search.label.text` and `params.search.label.html`?

packages/components/header/template.njk
```diff
@@ -1,167 +1,87 @@
...
+            <input class="nhsuk-search__input" id="search-field" name="{{ params.search.name | default('search-field') }}" type="search" placeholder="Search" autocomplete="off">
```
frankieroberto, line 49: Use camelCase for the default name so that the dot syntax can be used to access it (`{{ data.searchField }}`)?

```suggestion
            <input class="nhsuk-search__input" id="search-field" name="{{ params.search.name | default('searchField') }}" type="search" placeholder="Search" autocomplete="off">
```

packages/components/header/template.njk
paulrobertlloyd: Even though the text is hidden? What HTML might you place in this area?

packages/components/header/template.njk
paulrobertlloyd: I spotted this – there’s a lot of logic being used that might be best dealt with in styles or a broader review of how the different parts of the header interact with each other – I didn’t want to change any of that behaviour.

But if we think this is the right place to make this change, happy to do so.

frankieroberto: given we're changing the logic anyway (to test on presence of navigation items and search rather than the booleans) I think we can fix it here? Feels like an (existing) bug.

packages/components/header/template.njk
```diff
@@ -1,167 +1,87 @@
...
     </div>
-
-    {% endif -%}
-
-</header>
\ No newline at end of file
+{%- endif %}
+  </div>
+
+{%- if params.primaryLinks %}
```
frankieroberto, line 63: Is it worth also treating an empty list of primary links as false?

```suggestion
{%- if params.primaryLinks and (params.primaryLinks | length) > 0 %}
```

(would need doing in a few places).

Possibly not needed, but might be handy if you're doing something like this to build the list of links and there’s a possibility there might not be any?

```nunjucks
{{ header({
    "primaryLinks": [
      { url: "/", label: "Test" } if data.foo = "blah",
      { url: "/blah", label: "Blah" }  if data.foo = "blah"
    ]
}) }}
```

packages/components/header/template.njk
frankieroberto: Dunno, but I just keep forgetting which params need an object with a `text` and which ones just need strings, so maybe it's good to try and improve consistency? I can’t think of any good use cases for html though.

Could also support `label.classes` maybe and only default to visually-hidden? But maybe that's extending scope too much?

packages/components/header/template.njk
paulrobertlloyd: Given all the fallback values use the values used by the nhs.uk website, and that the website uses `q`, have updated this to use the same (meaning you can access this value using `data.q`, by default.

packages/components/header/template.njk
paulrobertlloyd: Okay, I have includes this change in a separate commit, currently 94b9ee14962cdd6029ca41fd095a28f5d4d1ad36

packages/components/header/template.njk
paulrobertlloyd: +1 for consistency, -1 for usefulness. Maybe an alternative option name is better, something like `visuallyHiddenLabel`. Are there other examples in the NHS Design System that do something similar?

paulrobertlloyd:
All checks now pass; I hadn’t updated the `header/index.njk` example ☺️

packages/components/header/template.njk
frankieroberto: visuallyHiddenLabel gets my vote. 👍

paulrobertlloyd: Cool. Updated to use that name.

frankieroberto (approved):
Looks great. Think it just needs an entry in the CHANGELOG, and an update to `packages/components/header/README.md`.

There's also a couple of uses of `showNav` and `showSearch` within `app/` to update.

Not sure when the next breaking change release is planned for? @anandamaryon1

paulrobertlloyd:
Updated the component `README.md`, added a line to the `CHANGELOG.md`, and updated the reaming examples.

Not sure this should be part of this PR, but I’d love to simplify the component API further by renaming `primaryLinks` to `navigation`, and merging `service` and `transactionService` options. Interested to hear what people think to making a bunch of breaking changes to the header component in the next major release?

I can create the corresponding PR on the service manual when we are happy with the shape of this PR, and if there any other changes we want to make to the header component at the same time.

frankieroberto (approved):
Looks good to me. I'm not sure what our merge strategy should be here, as this would have to go into a breaking change release. @anandamaryon1?

Given this is already a breaking change, I'd be up for renaming primaryLinks to navigation and merging service and transactionService options as part of a follow-up PR targeting the same release?

anandamaryon1:
@paulrobertlloyd I want to bring this together with the other header related breaking changes, but there are merge conflicts that are a bit complex, would you be able to take a look at merging in main first?

paulrobertlloyd:
Yes, happy to. A few things in this PR have happened on other PRs now, so this should be a simpler change once rebased.

paulrobertlloyd:
Rebased with `main`

frankieroberto:
@paulrobertlloyd @anandamaryon1 if this is still a breaking change release, is it better to focus on any remaining non-breaking changes to the header first, ie #1033?

Possibly some aspects of this PR could be pulled out into a non-breaking change release too, eg adding `search.visuallyHiddenLabel` - but that might be more work than it’s worth?

packages/components/header/template.njk
```diff
@@ -1,96 +1,82 @@
...
-                <svg class="nhsuk-icon nhsuk-icon__search" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" aria-hidden="true" focusable="false">
-                  <path d="M19.71 18.29l-4.11-4.1a7 7 0 1 0-1.41 1.41l4.1 4.11a1 1 0 0 0 1.42 0 1 1 0 0 0 0-1.42zM5 10a5 5 0 1 1 5 5 5 5 0 0 1-5-5z"></path>
-                </svg>
-                <span class="nhsuk-u-visually-hidden">Search</span>
-              </button>
-            </form>
-          </div>
+        <div class="nhsuk-header__search-wrap" id="wrap-search">
+          <form class="nhsuk-header__search-form" id="search" action="{{ params.search.action | default('https://www.nhs.uk/search/') }}" method="get" role="search">
+            <label class="nhsuk-u-visually-hidden" for="search-field">{{ params.search.visuallyHiddenLabel | default('Search the NHS website') }}</label>
```
frankieroberto, line 50: Wonder if it’s worth moving these defaults to the top of the template file and using a bunch of `set variableName = ...` statements, just to make it a bit clearer what all the text strings are that can be set? Just an idea though.

packages/components/header/template.njk
```diff
@@ -1,96 +1,82 @@
...
-          </a>
+          <a class="nhsuk-header__navigation-link" href="{{ item.url }}">{{ item.label }}</a>
         </li>
-        {%- endfor %}
+      {%- endfor %}
         <li class="nhsuk-header__navigation-item nhsuk-header__navigation-item--home">
-          <a class="nhsuk-header__navigation-link"  href="/">
-            Home
-          </a>
+          <a class="nhsuk-header__navigation-link" href="/">Home</a>
```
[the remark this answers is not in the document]
paulrobertlloyd: This has been merged into the breaking changes branch, so we can ignore this change for this PR.

app/components/all.njk
```diff
@@ -36,8 +36,7 @@
   }) }}
 
   {{ header({
-      "showNav": "true",
-      "showSearch": "true",
+      "search": "true",
```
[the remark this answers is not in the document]
paulrobertlloyd: Updated all occurrences to use `true`.

packages/components/header/template.njk
paulrobertlloyd: Done. Also added 2 extra options: `search.placeholder` (defaults to ‘Search’) and `search.visuallyHiddenButton` (defaults to ‘Search’). That pretty much covers everything you might want to customise for the search input.

PR #1005, opened 2024-09-06
Update table styles so nhsuk-table has bottom margin

edwardhorsford:
This fixes some inconsistency with the table styles.

The current table styles rely on a bottom margin being applied to the `table` element rather than `nhsuk-table`. The responsive table variant resets this so it has no bottom margin.

The styles include a `nhsuk-table-container` class that's presumably meant to wrap tables, but this is not used or documented on the design system site. I might guess it was added for a specific use case. For now I've left it, but teams shouldn't need to use it to get the correct bottom margin.

This applies some basic styles to `nhsuk-table` similar to how `govuk-table` styles it. If `nhsuk-table-container` can still be used.

### DIFF
--- a/packages/components/tables/_tables.scss
+++ b/packages/components/tables/_tables.scss
@@ -11,7 +11,7 @@
 .nhsuk-table-container {
   @include nhsuk-responsive-margin(7, "bottom");
 
-  display: block;
+  // display: block;
   -webkit-overflow-scrolling: touch;
   -ms-overflow-style: -ms-autohiding-scrollbar;
   overflow-x: auto;
@@ -20,6 +20,16 @@
   .nhsuk-table {
     margin: 0; /* [1] */
   }
+}
+
+.nhsuk-table {
+  @include nhsuk-font($size: 19);
+  
+  width: 100%;
+  @include nhsuk-responsive-margin(7, "bottom");
+
+  border-spacing: 0;
+  border-collapse: collapse;
 }
 
 /* Table row hover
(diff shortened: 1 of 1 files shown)


frankieroberto:
Might be worth opening a second PR to remove `.nhsuk-table-container` - looks like it’s no longer used for responsive tables at all (those use `.nhsuk-table--responsive)? 

Do you know about this at all @anandamaryon1?

anandamaryon1:
Looks good to me, I've not seen the `.nhsuk-table-container` being used anywhere but looking at the styles it could be used to create a scrolling table? 

Perhaps leave it in for now. 

Based on this, the responsive table example will need both classes, so the block class adding too (`.nhsuk-table`), for the margin bottom. 

Note to self: will need to check service manual aligns.

packages/components/tables/_tables.scss
```diff
@@ -22,6 +22,16 @@
   }
 }
 
+.nhsuk-table {
+  @include nhsuk-font($size: 19);
```
anandamaryon1, line 26: What is the reason for adding this? Could this perhaps mess up any specific type styling that people may have added in their tables?

edwardhorsford: It mirrors [what GOV.UK do](https://github.com/alphagov/govuk-frontend/blob/main/packages/govuk-frontend/src/govuk/components/table/_index.scss#L3) - providing some default sizing. I'd hope it wouldn't override any specific sizing as it should be less specific than any added by users. Not 100% confident about that though.

frankieroberto (dismissed):

PR #1018, opened 2024-09-16
Update changelog with breaking change notes

frankieroberto:
Update changelog with more detailed notes on the breaking changes.


CHANGELOG.md
```diff
@@ -2,12 +2,57 @@
...
 
+:boom: **Breaking changes**
+
+- Updated back link and breadcrumbs ([PR 1002](https://github.com/nhsuk/nhsuk-frontend/pull/1002))
+
+The breadcrumbs component no longer contains its own `<div class="nhsuk-width-container">` container.
+
+Instead, you should move it inside the existing `<div class="nhsuk-width-container">` container for your overall page, but before the `<main>` tag.
+
+This means that instead this:
```
edwardhorsford, line 13:
```suggestion
This means that instead of this:
```

CHANGELOG.md
```diff
@@ -2,12 +2,57 @@
...
 ## Unreleased
 
+:boom: **Breaking changes**
+
+- Updated back link and breadcrumbs ([PR 1002](https://github.com/nhsuk/nhsuk-frontend/pull/1002))
+
+The breadcrumbs component no longer contains its own `<div class="nhsuk-width-container">` container.
+
+Instead, you should move it inside the existing `<div class="nhsuk-width-container">` container for your overall page, but before the `<main>` tag.
+
```
edwardhorsford, line 12: I wonder if this should have a `## What you need to change` heading or similar.

CHANGELOG.md
```diff
@@ -2,12 +2,57 @@
...
+      <li class="nhsuk-breadcrumb__item"><a class="nhsuk-breadcrumb__link" href="#">Home</a></li>
+    </ol>
+  </nav>
+  <main class="nhsuk-main-wrapper" id="maincontent" role="main">
+    ...
+  </main>
+</div>
+```
+
+The back link should also be placed within the `<div class="nhsuk-width-container">` container but before the `<main>` tag, as before.
```
edwardhorsford, line 45: Two things:
* Is this a breaking change? it's unclear if it is or not. Again, might be helpful for more explicit 'you need to do x, y z.
* Include before and after examples?

CHANGELOG.md
```diff
@@ -2,12 +2,57 @@
...
+
+The back link now contains some default margin above it, so you can remove any override classes you added previously, such as `nhsuk-u-margin-top-4`, these can now be removed. However you can still include override classes if you want more or less spacing than the default.
```
anandamaryon1, line 47: Can remove the end of the first sentence, as it's duplicated: 

> The back link now contains some default margin above it, so you can remove any override classes you added previously, such as `nhsuk-u-margin-top-4`**, these can now be removed**.

frankieroberto: Opps, thanks!

CHANGELOG.md
frankieroberto: I tried to follow the conventions from the [existing CHANGELOG](https://github.com/nhsuk/nhsuk-frontend/blob/main/CHANGELOG.md), although that isn't super consistent.

Maybe 'Old code:' and 'New code:` sub-headings?

CHANGELOG.md
frankieroberto: It’s kinda unclear whether this is a breaking change or not, as the existing guidance on the [page template](https://service-manual.nhs.uk/design-system/styles/page-template) suggests having it in this position (within ohs-width-container but before `main`) however the [Back link page](https://service-manual.nhs.uk/design-system/components/back-link) itself is kinda ambiguous, suggesting it could go at the top or the bottom of the page!

Unlike the breadcrumb, the HTML output of the back link component hasn’t changed (as it’s not currently including its own `<div class="nhsuk-width-container">` element), so this feels like less of a breaking change. The only difference is the addition of default margin above.

anandamaryon1: Maybe just make the sentence clearer:

> The back link has not changed position, it should still be placed within the `<div class="nhsuk-width-container">` container but before the `<main>` tag.

CHANGELOG.md
anandamaryon1: Seems to follow previous convention and it does have the title of breaking change, so I'm happy with this as is.

CHANGELOG.md
edwardhorsford: I think it would be good to include the full markup, or else we could link to the layouts page where people can see them being used?

frankieroberto: I'm not having second thoughts about "The back link has not changed position" because whilst it's not changing position according to the current page template guidance, it is changing position according to the back link guidance itself (which suggests having it at the bottom).

Also in practice I think a lot of services mistakenly have it within the `<main>` tag so that it's not being skipped by the skip-to-main-content link.

How about something like:

> The back link should also be placed within the `<div class="nhsuk-width-container">` container but before the `<main>` tag.  Previous guidance suggested placing at the bottom of the page, but this has been updated to recommend placing it at the top.

(Then users can read the back link page for the full guidance)

edwardhorsford: The guidance doesn't show in code how to position the back link so I think I would link to layout too.

frankieroberto: The new guidance in https://github.com/nhsuk/nhsuk-service-manual/pull/2034 says:

> Within the HTML it should appear before the <code>&lt;main&gt;</code> tag. This is so that <a href="/design-system/components/skip-link">skip link</a> skips past the back link to the main content.

Could link to the page template guidance from the What’s new notes too though?

anandamaryon1: Updated the service manual whats new to mention the new position of the breadcrumbs

frankieroberto: Updated.

anandamaryon1 (approved):

PR #1026, opened 2024-09-20
Add pagination options

frankieroberto:
This extends the current [Pagination component](https://service-manual.nhs.uk/design-system/components/pagination) to support a new variant which is suitable for paging between numbered lists of pages, for example in search results or a long list of patients or appointments.

The variant is based heavily on the [GOV.UK Pagination variant](https://design-system.service.gov.uk/components/pagination/#for-navigating-between-pages-of-items), but with a few changes:

* colour changed from GOV.UK blue (`#1d70b8`) to NHS.UK blue (`#005eb8`)
* the arrow icon change to use the one from the existing NHS pagination component, which is wider and rounded
* visited pages use the NHS visited link purple colour (`#330072`) instead of the GOV.UK purple (`#4c2c92`)
* the hover style follows the NHS link style, which is to remove the underline and change to the maroon colour (`#7C2855`) instead of the GOV.UK style which is to have a thicker underline and go slightly darker blue (`#003078`) or purple
* because the existing arrow is blue rather than grey in the GOV.UK pagination, the arrow changes colour when hovered or visited, to match the link text

### Comparison between GOV.UK numbered pagination and this NHS one

| GOV.UK version | NHS version |
| -------------- | ----------- |
| Screenshot 2024-09-20 at 16 11 21 | Screenshot 2024-09-20 at 16 13 33 |
| Screenshot 2024-09-20 at 16 16 19 | Screenshot 2024-09-20 at 16 16 34 |

### Implementation

The existing Previous and Next variant for the NHS pagination component uses different HTML from the GOV.UK one, for example the NHS one uses a `<ul>` list containing 1 or 2 list items, whereas the GOV.UK one uses `<div>`s.

To avoid making any breaking changes to the existing pagination component, I’ve left that as-is, including all the existing HTML and CSS classes.

The new numbered variant uses a modifier class of the top-level `<nav>` element (`nhsuk-pagination--numbered`) to avoid clashing.

### Mobile / narrow screens

This needs a bit of thought. The GOV.UK version switches to centre-alignment below a certain width:

Screenshot 2024-09-20 at 16 22 51

...should we copy that? Possibly it switches too early (when there would be space for it to remain on one line).

The GOV.UK variant also drops some of the page numbers on narrow screens:

| Desktop | Mobile |
| ------- | ------ |
| Screenshot 2024-09-20 at 16 25 58 | Screenshot 2024-09-20 at 16 26 15 |

## TODO

- [ ] Consider mobile view
- [ ] Write guidance
- [ ] Test in [all supported browsers](https://github.com/nhsuk/nhsuk-frontend/blob/master/docs/contributing/testing.md)
- [ ] Unit tests
- [ ] Add to backstop image tests
- [ ] Code review
- [ ] CHANGELOG entry


paulrobertlloyd:
> The existing Previous and Next variant for the NHS pagination component uses different HTML from the GOV.UK one, for example the NHS one uses a `<ul>` list containing 1 or 2 list items, whereas the GOV.UK one uses `<div>`s.

Is there an opportunity for us to follow the GOV.UK decision here? My spidey sense is that using a list item for the now and next page items, inside a navigation landmark labelled pagination, is misleading (there may be more than 2 pages, its just that only the previous and next ones have been linked to).

Further more, if you have numbered pagination with say 3 items, it’ll be announced in a list having 5 items. Which is to say, I think the distinction between those links that appear in a list, and those used for previous and next links, is an important one.

> To avoid making any breaking changes to the existing pagination component, I’ve left that as-is, including all the existing HTML and CSS classes.

How compatible with the GOV.UK markup would this component be, say if you were using – entirely hypothetically you understand – a GOV.UK-based component library that allowed you to use NHS styles by changing class prefixes from `govuk-` to `nhsuk` ([cough](https://govuk-components.netlify.app/components/pagination/))?

---

Separately, see #1027; from both wanting visual consistency, and being able to identify the pagination arrows at these smaller sizes, the existing pagination arrows should have open tips.

frankieroberto:
@paulrobertlloyd:

> Is there an opportunity for us to follow the GOV.UK decision here? My spidey sense is that using a list item for the now and next page items, inside a navigation landmark labelled pagination, is misleading (there may be more than 2 pages, its just that only the previous and next ones have been linked to).
> 
> Further more, if you have numbered pagination with say 3 items, it’ll be announced in a list having 5 items. Which is to say, I think the distinction between those links that appear in a list, and those used for previous and next links, is an important one.

Ah sorry, I didn’t explain this well enough. I’ve followed the GOVUK markup for the new 'numbered pages' variant. This means that the numbers are within a `<ul>` list, but the previous and next links are in a separate `<div>` before or after the list, and not part of the list.

What I’ve not done is changed the markup used for the existing 'previous and next links with page titles' variant, which uses this markup in the NHS version:

```html
<nav><ul>
  <li><a href="#">Previous: page title</a></li>
  <li><a href="#">Next: page title</a></li>
</ul></nav>
```

but this in the GOV.UK version:

```html
<nav>
  <a href="#">Previous: page title</a>
  <a href="#">Next: page title</a>
</nav>
```

(`<div>`s and `<span>`s not shown)

Possibly we should also update this to match the GOV.UK version, but possibly the `<ul>` list is beneficial here (or not?) - and I also wanted to avoid a breaking change, at least initially. But maybe it’s better to do that at the same time?

> How compatible with the GOV.UK markup would this component be, say if you were using – entirely hypothetically you understand – a GOV.UK-based component library that allowed you to use NHS styles by changing class prefixes from govuk-* to nhsuk-* ([cough](https://govuk-components.netlify.app/components/pagination/))?

Right now the markup for the new numbered pages variant should be mostly compatible _except_ for the SVG for the arrows, which is different. There might also be some other differences in the css names.

paulrobertlloyd:
> But maybe it’s better to do that at the same time?

<details>
  <summary>Do it</summary>
  Ben Stiller saying ‘Do it’ from the movie ‘Starsky & Hutch’,
</details>

Arpana18:
**How can we hide (not show) the 'Previous' link when users are on the first page AND "Next" link when users are on the last page?** 

Our user research revealed that participants were confused about what the 'Previous' link would do when they're already on page 1. Similarly, users expect the 'Next' link to be disabled when they reach the last page.

This feedback suggests we need to implement hide states for pagination controls to provide clearer visual cues about available navigation options.

frankieroberto:
> **How can we hide (not show) the 'Previous' link when users are on the first page AND "Next" link when users are on the last page?**

@Arpana18 Hiya - the previous and next links are only shown when specified in the HTML or in the `previous` and `next` params within the Nunjucks macro. We’ll make it clearer in the accompanying guidance how to do this! Thanks for commenting. 👍

frankieroberto:
@colinrotherham @paulrobertlloyd I've updated this from main and fixed the conflicts and deprecations (I think).

Next step is to sort out the mobile view. Follow GOVUK and go centre aligned, with 'Previous' stacked above the numbers and 'Next' stacked below?

colinrotherham:
>Next step is to sort out the mobile view. Follow GOVUK and go centre aligned, with 'Previous' stacked above the numbers and 'Next' stacked below?

I'd say so, but only in the absence of a better (researched) option.

For comparison we could look at ideas in:

* https://github.com/alphagov/govuk-frontend/issues/3324
* https://github.com/alphagov/govuk-frontend/pull/3644

colinrotherham:
Sorry about the force push @frankieroberto

I've removed all the link style overrides so the defaults are now used, including `:visited` etc

Pagination with default link styles

<br>
<br>

Would be good to chat about the `--block` modifier used by GOV.UK Frontend

A few of the default styles were causing conflicts without it (e.g. stretching link `:hover` over the whole item)

colinrotherham:
Sorry about the force push again, 639 commits added and macro options are now ES modules

colinrotherham:
Merged with latest `main` now review app examples are consolidated

frankieroberto:
@colinrotherham I'm planning to pick this up again later this week, as it a component desperately needed by my service.

But this would be a v11 thing.

colinrotherham:
Updated with `main` where reference images were generated again

colinrotherham:
Fixed all the conflicts to bring in:

1. Next and Previous icons via `nhsukIcon()`
2. Link colours from https://github.com/nhsuk/nhsuk-frontend/pull/1367
3. Link underlines from https://github.com/nhsuk/nhsuk-frontend/pull/1366

frankieroberto:
Okay I’ve updated this, with the main aim of making this now backwards-compatible, so that it is no longer a breaking change.

The downside of this is that it remains un-aligned with the govuk frontend one at both the macro level and the HTML output is also slightly different (especially for the existing variant), but I’ve made it as compatible as I can. The use of the nested `nhsukIcon` macro means it’ll never be 100% the same anyway.

We could still decide to deprecate `previousUrl` for `previous.href` and `previousText` for `previous.labelText` etc, if we wanted to.

Still got a bit of fiddling to do on the design and the mobile-responsiveness, but getting closer...

colinrotherham:
Have a look from line 6?

https://github.com/nhsuk/nhsuk-frontend/blob/component-form-inputs/packages/nhsuk-frontend/src/nhsuk/components/input/template.njk

I had to do some wrangling to make the GOV.UK Frontend object syntax for prefix/suffix work with our strings

But it worked. Old and new API,
side by side

If that opens any doorways for you?

frankieroberto:
@colinrotherham yeah, I was considering doing similar to support both `previousUrl` (which could be deprecated) and `previous.href`. Depends how important we think it is to update the Nunjucks params.

For now, I’ll focus on the remaining design tasks.

frankieroberto:
@colinrotherham @anandamaryon1 @paulrobertlloyd @edwardhorsford ok, this is ready for review! I'll make a start on the guidance separately.

This should be completely backwards-compatible, so could be included in a 10.1 release.

colinrotherham:
Thanks @frankieroberto!

Do we want to bring in some fixes for the existing pagination too?

E.g. The full width focus styles

Screenshot 2025-09-26 at 16 36 12

<br>
<br>

I had a couple of fixes from when it was stacked on top of the link colour branch in https://github.com/nhsuk/nhsuk-frontend/commit/700052db77fab8819637d829bd9bf047dd1ff666 and https://github.com/nhsuk/nhsuk-frontend/commit/8f4939c1356473259bbd26f06c639948dbd4760f

Or can we uplift anything else from GOV.UK Frontend?

frankieroberto:
@colinrotherham:

> Do we want to bring in some fixes for the existing pagination too?
> 
> E.g. The full width focus styles
> 
> Screenshot 2025-09-26 at 16 36 12

I’ve kept the existing pagination design as-is for now. Have asked in our internal channels whether people think this 50% width clickable area and focus style is a bug or not. If we think it’s worth changing, it could be done as a separate PR?

anandamaryon1:
Sorry it's taken me so long to take a proper look at this @frankieroberto.

Looking now and I notice a few little things:

Hover on number links does not remove the underline as it should with normal links (currently), or change colour:
image

Our general guidance suggests not to use bold for emphasis (https://service-manual.nhs.uk/content/formatting#bold-italics-and-underlining) which is debatable but, in keeping with it, we could unbold the next and previous links, they so already have an icon to emphasise them. 

Focus style on current item is missing text-colour:
image

Hover and focus states clash:
image

I don't think the ellipsis should receive hover:
image

I know you might not be able to do these right away so I can try implement if you're happy with these changes?

frankieroberto:
@anandamaryon1 all good spots, thanks!

> I know you might not be able to do these right away so I can try implement if you're happy with these changes?

I'm about to be away for 5 days, so do feel free to push any changes to the branch!

anandamaryon1:
Made some edits based on my comments, keen for others review and feedback

PR #1027, opened 2024-09-21
Update arrow icons to be visually consistent with other arrow-like icons

paulrobertlloyd:
## Description

In reviewing #1026, the difference in arrows used for pagination versus chevrons used in other components and the right-pointing action link icon becomes clearer, with the pagination arrows looking like they are from a different icon set.

Also, with the numbered pagination option, it looks like the arrows might be rendered smaller; at which point **the arrow tips become harder to identify**.

This PR update the pagination arrow icons to have open arrow tips, using the same path metrics as those used by the other arrow icons.

This could also potentially be merged into and be part of #1026, or merged separately (with corresponding a changelog entry). 

### Pagination before

Pagination component with current arrow icons.

### Pagination after

Pagination component with revised arrow icons.

### Arrow icons before

All current arrow icons

### Arrow icons after

All revised arrow icons

## Checklist

- [ ] CHANGELOG entry

### DIFF
--- a/packages/assets/icons/icon-arrow-left.svg
+++ b/packages/assets/icons/icon-arrow-left.svg
@@ -1,3 +1,3 @@
 <svg class="nhsuk-icon nhsuk-icon__arrow-left" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" aria-hidden="true" width="34" height="34">
-  <path d="M4.1 12.3l2.7 3c.2.2.5.2.7 0 .1-.1.1-.2.1-.3v-2h11c.6 0 1-.4 1-1s-.4-1-1-1h-11V9c0-.2-.1-.4-.3-.5h-.2c-.1 0-.3.1-.4.2l-2.7 3c0 .2 0 .4.1.6z"></path>
+  <path d="M10.7 6.3c.4.4.4 1 0 1.4L7.4 11H19a1 1 0 0 1 0 2H7.4l3.3 3.3c.4.4.4 1 0 1.4a1 1 0 0 1-1.4 0l-5-5A1 1 0 0 1 4 12c0-.3.1-.5.3-.7l5-5a1 1 0 0 1 1.4 0Z"></path>
 </svg>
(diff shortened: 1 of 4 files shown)


frankieroberto:
@paulrobertlloyd ooh I like this, it looks a lot more consistent! It's interesting how one of the main style differences between govuk and nhs, eg on buttons and arrows, is sharp corners vs rounded corners 😄 

Do you think it’s worth exploring whether the pagination component could even use the chevron variant without the horizontal line? The 'Previous' and 'Next' titles could then perhaps be aligned (left and right) with the page titles underneath?

edwardhorsford:
I agree the new arrows seem more visually in keeping with general styles. Nice work.

paulrobertlloyd:
> Do you think it’s worth exploring whether the pagination component could even use the chevron variant without the horizontal line? The 'Previous' and 'Next' titles could then perhaps be aligned (left and right) with the page titles underneath?

Possibly, although I like the greater directionality (spoken like a true designer) of the arrows versus the chevrons. Aligning the text would be a good call, be good to tidy up the alignment a bit.

frankieroberto:
> greater directionality

made-up phrase 😉 

Whilst we’re here, let’s talk colours? Existing behaviour:

* Back link uses a blue arrow which changes colour on hover and focus to match the link (but curiously not when active)
* Breadcrumbs arrows stay grey - which makes sense as they're not part of the links but between them
* Action link arrows only change colour on focus, but otherwise stay green
* Expander icons change colour on focus but not on hover.
* Pagination arrows change colour on focus, hover and visited but not active

A bit inconsistent, but possibly with good reason?

In general, I think it’s good if there can be some indication on hover that the icon is part of the link - but that could be a background colour change (govuk pagination links use a background grey colour on hover) or an icon fill colour change?

paulrobertlloyd:
For the colouranality (😉), maybe we can have some rules around this; such as when an item is hovered, active or focused, all elements within that link share the same standard link colours for these states?

frankieroberto:
This sent me down a rabbithole trying to work out where the standalone `.svg` file are used, and I think the answer is: they're not? If so, #1028 suggests removing them.

edwardhorsford:
For a prototype today I've played with making a 'forward link' - to go alongside a backlink. I had Claude invert the svg for me but more consistent arrows would have probably been helpful.

Screenshot 2025-03-20 at 14 48 19

anandamaryon1:
Agree these look better and are easier to distinguish 👍 

Yes I think best to merge into pagination PR since the icons are directly part of that component's code via the SVGs.

frankieroberto:
@paulrobertlloyd I've copied this into #1026.

PR #1028, opened 2024-09-23
Remove standalone icon SVG files

frankieroberto:
These standalone `.svg` icon files aren't used by any of the components, as each component either embeds the SVG [directly within the Nunjucks macro](https://github.com/nhsuk/nhsuk-frontend/blob/main/packages/components/pagination/template.njk#L13-L15) or includes it [using a data-url in the CSS](https://github.com/nhsuk/nhsuk-frontend/blob/main/packages/components/details/_details.scss#L140).

On the one hand, maybe it’s useful to have these as standalone files just as a reference, so that all the icons can be easily reviewed together?

But on the other hand, duplicating them here runs the risk of them accidentally deviating from the actual icons used by the components?

Including them within the `nhsuk-frontend` package also adds another 100kb to the package (as they’re included twice, once in `packages/assets` and once in `dist/app/asset`).

Note: It’s possible that these _are_ used somewhere and I’ve missed it somehow, but I’ve done a bunch of services across the code, and I can’t find anything that does.

### DIFF
--- a/packages/assets/icons/icon-arrow-left.svg
+++ b/packages/assets/icons/icon-arrow-left.svg
@@ -1,3 +1,0 @@
-<svg class="nhsuk-icon nhsuk-icon__arrow-left" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" aria-hidden="true" width="34" height="34">
-  <path d="M4.1 12.3l2.7 3c.2.2.5.2.7 0 .1-.1.1-.2.1-.3v-2h11c.6 0 1-.4 1-1s-.4-1-1-1h-11V9c0-.2-.1-.4-.3-.5h-.2c-.1 0-.3.1-.4.2l-2.7 3c0 .2 0 .4.1.6z"></path>
-</svg>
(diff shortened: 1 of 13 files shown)


edwardhorsford:
My guess might be that these are considered the 'master assets' and are archived here - that the data-uri's are generated from them, and the macros copy from them. The repo is being used for persistent storage. If that's the case, I would leave them - though perhaps with a text file describing their purpose. Do you have any clues about their origin / usage?

frankieroberto:
@edwardhorsford I can't see any scripts to generate the data-uris from the files, but perhaps that's been done manually?

I have a hunch they they were previously used as the target of `` tags, but I’ve not dug through the git history to find out.

Could keep them as a reference, but will just have to remember to keep them up-to-date if changing the icons in the css or macros!

If keeping them for that purpose, they could be removed from the build process so that they're not included in the distributed package?

edwardhorsford:
If the purpose is for archival purposes, their location could probably be clearer - and removed from the build process as you say. Perhaps someone who's worked on this repo longer might have some ideas.

anandamaryon1:
They're linked to from here: https://service-manual.nhs.uk/design-system/styles/icons

I assume as a way for people to be able to download them separately, rather than having to rip them out with dev tools? But this is a guess. 

But if they're shipped with the frontend but not used, then yes I suppose they should be removed from the frontend package.

paulrobertlloyd:
After reviewing the usage of icons, including those used in CSS, the only icons left at this point are those that will be removed by #1521 which introduces the `nhsukIcon` macro and becomes the canonical source of paths used for inline SVG icons. As such, closing this issue.

PR #1030, opened 2024-09-24
Use paragraphs for error messages

### LINKED ISSUE #1029: Error message should use paragraphs rather than spans
# Bug Report

Our [error message component](https://service-manual.nhs.uk/design-system/components/error-message) uses spans, where I think it should use paragraphs. I checked GOV.UK and it used to use `spans` prior to v4.0 - it was [changed in this pr](https://github.com/alphagov/govuk-frontend/pull/2452).

I suspect one consequence of this is that VoiceOver will read the error message in a less clear way.

With the gov.uk design system, VoiceOver reads the error message as one line: `Error: The date your passport was issued must be in the past`. With ours, it reads `Error:` on its own, and you have to manually advance to the next bit. I suspect this is because VoiceOver treats spans within a block level element as one thing, but when it's inline elements in inline elements it treats them separately.

edwardhorsford:
## Description
Fixes #1029.

I'm unsure if this should be considered a breaking change or not - might be worth a discussion.

Visually there should be no difference.

|Before| After|
|-|-|
|[Video with span](https://github.com/user-attachments/assets/5585e91a-3320-4c41-8523-879355f7006f)|[Video with paragraph](https://github.com/user-attachments/assets/4cf84e6b-cffd-4d11-9a87-ee13df0fc045)|

## Checklist

- [ ] CHANGELOG entry

### DIFF
--- a/packages/components/error-message/README.md
+++ b/packages/components/error-message/README.md
@@ -11,10 +11,10 @@ Find out more about the error message component and when to use it in the [NHS d
 ### HTML markup
 
 ```html
-<span class="nhsuk-error-message">
+<p class="nhsuk-error-message">
   <span class="nhsuk-u-visually-hidden">Error:</span>
   Error message about full name goes here
-</span>
+</p>
 ```
 
 ### Nunjucks macro
--- a/packages/components/error-message/_error-message.scss
+++ b/packages/components/error-message/_error-message.scss
@@ -8,5 +8,6 @@
   clear: both;
   color: $nhsuk-error-color;
   display: block;
+  margin-top: 0; // Reset any default browser margins for paragraphs
   margin-bottom: nhsuk-spacing(3);
 }
--- a/packages/components/error-message/template.njk
+++ b/packages/components/error-message/template.njk
@@ -2,10 +2,10 @@
 
 {% set visuallyHiddenText = params.visuallyHiddenText | default("Error") -%}
 
-<span class="nhsuk-error-message
+<p class="nhsuk-error-message
 {%- if params.classes %} {{ params.classes }}{% endif %}"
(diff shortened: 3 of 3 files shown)


anandamaryon1:
Had a quick test of this and I don't find any difference in how VoiceOver or NVDA read the error message.

VoiceOver reads error separately and NVDA reads the whole line as one, regardless of `span` or `p`. 

**VoiceOver** (`span` then `p`)
https://github.com/user-attachments/assets/3ae288c4-6128-4cf7-9872-e8a81ca37db2


**NVDA** (`span` then `p`)
https://github.com/user-attachments/assets/0ae8f07f-64db-465c-9910-0afd64bfc779

edwardhorsford:
@anandamaryon1 your first example shows a visual `<p>` in the heading - is that a bug or something you added to tell them apart? If the later, how do they compare to GOV.UK's error messages?

frankieroberto:
@edwardhorsford @anandamaryon1 testing by GDS back in 2021 didn't seem to suggest there was any difference in screen reader treatment either? https://github.com/alphagov/govuk-frontend/issues/2083#issuecomment-979874842

anandamaryon1:
Apologies this got left in limbo (for exactly a year!) @edwardhorsford…

Given that this will align us with GOV, shall we go ahead and merge this? Any reason not to @colinrotherham?

colinrotherham:
Ah perfect, just needs bringing back up to date

We've put https://github.com/nhsuk/nhsuk-frontend/issues/1029 against the v11.0.0 milestone currently

Just need to check if moving from `<span>` to `<p>` is 100% compatible (i.e. optional for non-Nunjucks libraries) then it can go in as a non-breaking change under v10.1.0 instead

PR #1036, opened 2024-09-27
Update error summary markup to improve screen reader announcements

### LINKED ISSUE #1035: Error summary markup needs to be updated to follow GOV.UK
# Bug Report
Whilst testing the error summary with VoiceOver on Safari, I observe that 'There is a problem' gets read out twice. I think this is because the error summary itself is `aria-labelledby` the title, and then Voiceover _also_ reads the contents of the error summary - first of which is the title.

Comparing to GOV.UK, their [markup updated in v4.4](https://github.com/alphagov/govuk-frontend/pull/2677) to move the alert to a child element and adjust other bits for screen readers.

edwardhorsford:
## Description
Fixes #1035.

This replicates [this GOV.UK Frontend pr](https://github.com/alphagov/govuk-frontend/pull/2677).

The issue I was having was 'There is a problem' getting read out twice. That's not mentioned in the GDS pr, but does seem to be fixed in making these improvements.

I don't think this is a breaking change (GDS didn't release it as one), but it does update the markup.

## Checklist

- [ ] CHANGELOG entry

### DIFF
--- a/packages/components/error-summary/template.njk
+++ b/packages/components/error-summary/template.njk
@@ -1,27 +1,32 @@
 {% from "../../macros/attributes.njk" import nhsukAttributes %}
 
 <div class="nhsuk-error-summary
-  {%- if params.classes %} {{ params.classes }}{% endif %}" aria-labelledby="error-summary-title" role="alert" tabindex="-1"
+  {%- if params.classes %} {{ params.classes }}{% endif %}" tabindex="-1"
   {{- nhsukAttributes(params.attributes) }}>
-  <h2 class="nhsuk-error-summary__title" id="error-summary-title">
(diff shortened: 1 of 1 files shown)


mcheung-nhs:
I've had a look and the update does fix the issue of VoiceOver on Mac Safari not reading out the error summary title twice.

However, on NVDA, the summary title ("This is a problem" in the examples below) is not initially read our when compared to the current version.

Tested on NVDA v2024.3.1 on Windows 10.

Current:
https://github.com/user-attachments/assets/09f29c96-e414-4a22-b674-7eb179c457e5

With update:
https://github.com/user-attachments/assets/7b3212c1-5a9c-45b9-83c1-82f44276ed5c

@edwardhorsford - could you confirm this is the case and also whether this is a problem?

PR #1051, opened 2024-10-22
Correctly style disabled warning button

### LINKED ISSUE #1034: Warning button disabled styling
# Bug Report

## What is the issue?

When adding both the `nhsuk-button--disabled` class and the `disabled` attribute to a warning button, the button's colour changes from red to green. This issue also occurs when we just use the `disabled` attribute. We expect this to behave like how it does when only the `nhsuk-button--disabled` is used.

1. Create a warning `button`
2. Add the `nhsuk-button--disabled` class and `disabled` attribute

* Device: N/A
* Operating System: N/A
* Browser: N/A
* Browser version: N/A 
* NHS.UK frontend package version: 8.3 and 9
* Node version: N/A
* npm version: N/A

nhsuk-button-styles

paulrobertlloyd:
## Description

- Simplifies how disabled buttons are styled
- Adds examples of all button variants in disabled state on example page
- Updates backstop images

Fixes #1034.

I’m not really sure why so much CSS is being used to style disabled buttons! We appear to give disabled buttons background colours that have already been defined, and so much else, when all we need to do is give buttons with `nhsuk-button--disabled` class or `disabled` `0.5` opacity, change to the default cursor and remove pointer-events.

Maybe I’m missing something? (I note that GDS do the same for `govuk-frontend` so might be something to ask them about this on X-GOV Slack)?

## Checklist

- [x] CHANGELOG entry


anandamaryon1:
Your simpler approach makes sense to me, but given that GOV do the same, it'd be worth understanding why. Did you ask on xgov slack?

frankieroberto (dismissed):
Looks a lot simpler to me! 🙌 

I wonder if the `.nhsuk-button--disabled` class ought to be removed? I can’t see any valid use case for styling a button as disabled if it isn’t _actually_ disabled, and if it is disabled, then the `.nhsuk-button:disabled` selector will style it anyway?

frankieroberto:
@paulrobertlloyd you’ll have to update/rebase from main to get the newly-split-up tests to run.

paulrobertlloyd:
Rebased with `main`.

Happy to remove `.nhsuk-button--disabled`. Right now this is still needed as the component allows for a disabled link, indicated using `aria-disabled="true"`:

https://github.com/nhsuk/nhsuk-frontend/blob/1ba3a91e63f279bbb8a646e5766e493aa5675777/packages/components/button/template.njk#L30

This is not addressable using the `:disabled` pseudo class, (though we could add `[aria-disabled]` if we wanted to remove the `.nhsuk-button--disabled` modifier).

Not sure if ‘disabled links’ is a markup pattern that should really be supported; GOV.UK Frontend only allows you to disable `button` buttons.

frankieroberto:
Hmm… links styled as disabled buttons… 😵‍💫

paulrobertlloyd:
More investigation… it’s complicated. `govuk-fronted` changes the pointer to `not-allowed` on hover, and thus cannot disable `pointer-events` which means hover styles need to be accounted for.

We disable `pointer-events`, which means we don’t need to account for hover (and active) styles. If we wanted to have the `not-allowed` pointer, we’d need a whole lot of CSS to revert the active plus disabled state… and that gets real ugly due to the box-shadow and faux-button press. Which is to say, I think this simplification works, in our case at least.

anandamaryon1:
I'm happy with this, thanks @paulrobertlloyd 

Links styled as disabled buttons doesn't seem good, I can't think of a legitimate use-case for them. Plus making actually disabled links is a whole messy thing: https://css-tricks.com/how-to-disable-links/

Is it gung-ho to just remove it?

paulrobertlloyd:
> Is it gung-ho to just remove it?

I can well imagine some services using this anti-pattern; if we were to remove it in a minor release, links that should appear/behave disabled would potentially start ‘working’.

Let’s merge this fix in for the next minor release. I can create a PR to remove support for disabled links for the next major release.

PR #1052, opened 2024-10-23
Pin Prettier version

paulrobertlloyd:
## Description

`package.json` lists Prettier as `^3.3.2`. This means that different versions can be installed locally as that ran during CI, which always installs the latest version (currently v3.3.3).

This is [recommended by Prettier](https://prettier.io/docs/en/install#summary):

> Install an exact version of Prettier locally in your project. This makes sure that everyone in the project gets the exact same version of Prettier. Even a patch release of Prettier can result in slightly different formatting, so you wouldn’t want different team members using different versions and formatting each other’s changes back and forth.

This has caused issues recently, with a patch release changing how `box-shadow` rules get formatted.

## Checklist

- [x] CHANGELOG entry

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -37,7 +37,7 @@
         "jest-environment-puppeteer": "^10.0.1",
         "minimist": "^1.2.8",
         "nunjucks": "^3.2.4",
-        "prettier": "^3.3.2",
+        "prettier": "3.3.3",
         "puppeteer": "^22.12.1",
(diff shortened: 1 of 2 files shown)


frankieroberto (approved):
This seems sensible. 👍 

I've also opened #1053 which should help make sure that the version specified in the `package.lock` (for all npm packages) is the one used by the tests.

PR #1056, opened 2024-10-23
Split up test jobs so that they can be run concurrently

frankieroberto:
Currently all the tests run one after each other, taking just under 5 minutes in total.

Aim here is to run them concurrently, which should roughly half the time it takes to run them all (slowest one is visual regression tests which take about 2.5 mins).

In addition they’ll all get reported on the PR separately, which should make it easier to quickly see which one failed, rather than having to dig into the actions log to work it out.

## Before

Screenshot 2024-10-23 at 19 13 53

## After

Screenshot 2024-10-23 at 19 20 42


paulrobertlloyd (approved):
Nice!

paulrobertlloyd:
Not sure why this can’t be merged; something about needing all required statuses?

frankieroberto:
@paulrobertlloyd yeah, will need an admin to update the branch protection rules, removing the requirement for the current general “Pull request” job to pass and instead adding the 4 new ones.

@roshaanbajwa is this something you could do?

Also be good to check if the names for all 4 jobs make sense? I wasn’t sure if “JavaScript unit test” and “Visually regression tests” were the right terms or not?

frankieroberto:
Possibly the overall workflow should be renamed from “Pull request” too?

The names only affect the interface I think, but given these tests will also all run if any internal or outside new contributors open a pull request, the less confusing they are the better? (Although we can also help anyone who gets stuck)

anandamaryon1:
Seems very sensible to me 👍 

@roshaanbajwa is on leave now but I'm sure we can get this checked and merged on his return.

frankieroberto:
@anandamaryon1 @roshaanbajwa the settings should be on this page: https://github.com/nhsuk/nhsuk-frontend/settings/branches under 'branch protection rules'.

roshaanbajwa (approved):

PR #1057, opened 2024-10-23
Add dependabot config

frankieroberto:
This sets up Dependabot to monitor for both updates to GitHub Actions and NPM packages.

See [Dependabot configuration](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file) for the different options.

Open to suggestions on the best update frequencies!

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+  - package-ecosystem: "npm"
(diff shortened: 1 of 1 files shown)


paulrobertlloyd:
I think monthly for both. Daily is too frequent and noisy, and sometimes you want to wait to make sure there are not further patches on previous patches.

frankieroberto:
@paulrobertlloyd yeah that might make sense, especially for a frontend library which is only released every few weeks anyway. I think Dependabot might still flag security updates more quickly?

paulrobertlloyd (approved):

PR #601, opened 2020-04-06
Vendor in Sass-MQ

pezholio:
## Description

We're trying to use NHSUK Frontend on a project which has a non-standard location for `node_modules`, and as a result, we were getting the following error:

```
sass.CompileError: Error: File to import not found or unreadable: ./node_modules/sass-mq/mq.
        on line 17:1 of assets/_dev/node_modules/nhsuk-frontend/packages/core/tools/_sass-mq.scss
        from line 9:1 of assets/_dev/node_modules/nhsuk-frontend/packages/core/tools/_all.scss
        from line 6:1 of assets/_dev/scss/screen.scss
>> @import './node_modules/sass-mq/mq'; // [1] //
```

After a bit of Googling, I found that govuk-frontend had a similar problem and solved this by [vendoring in SassMQ](https://github.com/alphagov/govuk-frontend/pull/657). I've attempted to replicate this approach in NHS Frontend too.

Thoughts welcome!

## Checklist

- [x] Tested against our [testing policy](https://github.com/nhsuk/nhsuk-frontend/blob/master/docs/contributing/testing.md) (Resolution, Browser & Accessibility)
- [x] Follows our [coding standards and style guide](https://github.com/nhsuk/nhsuk-frontend/blob/master/docs/contributing/coding-standards.md)
- [x] CHANGELOG entry

PR #1064, opened 2024-11-03
Upgrade Sass to 1.80.6 (and fix as many deprecations as possible)

frankieroberto:
This upgrades Sass from 1.77.6 to 1.80.6.

The new version introduces a lot more deprecation warnings ahead of Sass 2.0.0 – so I’ve tried to fix as many as possible.

A lot of the warnings are from the [vendored Sass-mq](https://github.com/nhsuk/nhsuk-frontend/pull/601) however. I've not touched that at all as it feels a bit icky to edit a vendor file (although it has had some code style tweaks over the years).

The GOV.UK Design System team are [considering removing the sass-mq dependency](https://github.com/alphagov/govuk-frontend/issues/5143) altogether. However there was a [6.0.0 release](https://github.com/sass-mq/sass-mq/releases/tag/v6.0.0) of it last week which should resolve the deprecations.

The other remaining deprecation warnings are about [use of @import](https://sass-lang.com/documentation/breaking-changes/import/) - however moving away from is bigger job (changing every instance of `@import` to `@use` and namespacing all the imported functions).

## Checklist

- [ ] CHANGELOG entry

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -39,7 +39,7 @@
         "nunjucks": "^3.2.4",
         "prettier": "3.3.3",
         "puppeteer": "^22.12.1",
-        "sass": "^1.77.6",
+        "sass": "^1.80.6",
         "start-server-and-test": "^2.0.4",
         "stylelint": "^13.13.1",
         "stylelint-config-prettier": "^9.0.5",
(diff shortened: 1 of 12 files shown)


paulrobertlloyd:
> The other remaining deprecation warnings are about [use of @import](https://sass-lang.com/documentation/breaking-changes/import/) - however moving away from is bigger job (changing every instance of @import to @use and namespacing all the imported functions).

From [the Sass docs](https://sass-css.org/documentation/at-rules/use):

> You can even load a module without a namespace by writing `@use "<url>" as *`.

In my limited experience trying to move over to using `@use`, this means you can do the following:

```diff
- @import 'core.scss'
+ @use 'core.scss' as *
```

This then means you don’t need to use namespaced variables. However, the Sass documentation goes on to say:

> We recommend you only do this for stylesheets written by you, though; otherwise, they may introduce new members that cause name conflicts!

Which is to say this may be a short term fix, but longer term, a breaking change could introduce an `nhsuk` namespace for all mixins, functions and variables owned by `nhsuk-frontend`.

frankieroberto:
@paulrobertlloyd ooh nice, I hadn’t spotted that. It certainly could be a short-term option, and then we could gradually switch to namespaces, starting with the most-reused functions and  mixins and so on?

paulrobertlloyd:
Trying to resolve these deprecation warnings in my prototype, one thing with the above approach (`@use "<url>" as *`) is that it seemingly causes conflicts if 2 SCSS files export a variable, mixin or function with the same name. As this project shares some names with GOV.UK Frontend, I suspect there’d be clashes if you were to import anything from GOV.UK Frontend.

This module system seems nice un all, but get the feeling it’s an all or not-at-all kinda thing. 😕 Definitely warrants more investigation.

frankieroberto:
@colinrotherham I think I only got about halfway through this before giving up (there were still a bunch of deprecations) but maybe fixing some of them is better than nothing?

I'm not sure about silencing the deprecations though - does that introduce the risk that we might continue putting it off? 😅

frankieroberto:
> the

frankieroberto:
> It's so noisy without though!

The noise is the point! (Sort-of). Our other option might be to downgrade or pin the sass version until we’re ready to upgrade having fixed deprecations?

> Shall we open a new GitHub issue for the migration from `@import` to `@use`?

Yeah good idea.

I think nowadays I’d be half tempted to skip sass altogether and just use plain css with some of the new features like variables, and basic concatenation. But we are where we are.

packages/components/task-list/_task-list.scss
```diff
@@ -5,9 +5,11 @@
 /// @link https://github.com/alphagov/govuk-frontend Original code taken from GDS (Government Digital Service)
 ////
 
+@use "sass:color";
+
 // NHS pages have a grey background, so we need a slightly daker colour for the hover
 // This produces 1.1:1 contrast, the same as GOV.UK’s
-$nhsuk-task-list-hover-color: darken($color_nhsuk-grey-5, 6%);
+$nhsuk-task-list-hover-color: color.adjust($color_nhsuk-grey-5, $lightness: -6%);
```
frankieroberto, line 12: @colinrotherham I think this should result in the same colour, but I didn't actually check. [`darken` is deprecated](https://sass-lang.com/documentation/modules/color/#darken). Not sure if we need to also define colour space too though?

PR #1075, opened 2024-11-20
Remove support for `nhsuk-button--disabled` class

paulrobertlloyd:
## Description

Removes support for disabled links, which are not really a thing (or at least, shouldn’t be).

- Removes support in `button` Nunjucks macro for adding `aria-disabled="true"` if `element` is `a` and `disabled` is true
- Removes `nhsuk-button--disabled` modifier; buttons are styled as disabled if they have the `disabled` attribute applied
- Uses the `disabled` boolean attribute (as opposed to `disabled="disabled"`

## Checklist

- [x] CHANGELOG entry


frankieroberto (approved):
Makes sense to me. 👍

A breaking change so not sure when it should get merged. @anandamaryon1?

anandamaryon1:
Note to self: update this (now incorrect) guidance on the [button page](https://service-manual.nhs.uk/design-system/components/buttons): 

> Disabled buttons
Disabled buttons have poor contrast and can confuse some users. Only use them if user research shows it makes things easier for users to understand.
> We have developed disabled versions of the 3 buttons but we haven't tested them yet. You can [get the latest disabled button code in the NHS.UK frontend library in GitHub](https://github.com/nhsuk/nhsuk-frontend/tree/main/packages/components/button).

frankieroberto:
> Note to self: update this (now incorrect) guidance on the [button page](https://service-manual.nhs.uk/design-system/components/buttons):
> 
> > Disabled buttons
> > Disabled buttons have poor contrast and can confuse some users. Only use them if user research shows it makes things easier for users to understand.
> > We have developed disabled versions of the 3 buttons but we haven't tested them yet. You can [get the latest disabled button code in the NHS.UK frontend library in GitHub](https://github.com/nhsuk/nhsuk-frontend/tree/main/packages/components/button).

@anandamaryon1 I think this guidance would still stand (unless you have any more recent research?) but with the caveat that only disabled actual `<button>`s are supported, and links styled as disabled buttons aren’t supported? But possibly that’s such a weird use-case that it doesn’t need to be spelled out?

frankieroberto (approved):

colinrotherham:
One of the merge commits restored `--disabled` which is why we had to remove it again

I've rebased and restored the change (credit to @paulrobertlloyd)

Thanks

PR #1079, opened 2024-11-21
Scope pull-request worflows to paths

paulrobertlloyd:
Currently, visual regression tests are run when any file changes in a pull request. Should you make a tiny change to any file in the rep, such as `package.json` or a configuration file, it’ll take ~3.45mins to run visual regression tests. Changes outside of components get needlessly tested, but more importantly, untold resources are consumed while this task works away on a remote server somewhere.

This PR breaks out the previous `pull-request.yml` workflow into 3 separate workflows:

- `analyse.yml` - Perform static analyse if any file has changed
- `lint.yml` - Check code style if any file has changed (note that linters configured to only examine certain files within `packages/components` and `/dist/app/components` folders, arguably JS outside of `packages/components` should be linted, too)
- `test.yml` - Test JavaScript and perform visual regression tests only if a file has been changed in the `packages` folder. This is the change that should hopefully reduce the number of times these tests are run, notably when dependency updates are made to `package.json` and `package-lock.json`.


.github/workflows/test.yml
```diff
@@ -0,0 +1,37 @@
+name: Pull request
+
+on:
+  pull_request:
+    paths:
+      - '/packages/**'
```
frankieroberto, line 6: I think to be safe the tests would have to run if either anything in `/package` has changed OR the tests themselves (in `/tests`) OR in `/app` - as the visual regression tests rely on examples within `/app` to check the rendering?

...but we could skip the tests if a PR doesn’t touch any of those (eg a change to docs)?

We’d also have to change the branch protection rules to turn off the requirements for the tests to pass (which might defeat the point?) - unless there’s a way for the job to succeed without actually running the tests, if none of files in the paths have been changed? 

```suggestion
      - '/packages/**'
      - '/tests/**'
      - '/app/**'
```

frankieroberto (changes requested):

.github/workflows/test.yml
paulrobertlloyd: Good point, updated. I can’t see the branch protection rules, but am assuming only test that are run need to pass; if the visual regression test isn’t run then it won’t fail, no?

frankieroberto: @paulrobertlloyd if it’s not run then it'll show as pending and won't let you merge the pull request (unless you let admins override it). I had a quick look at the `paths` documentation in GitHub but couldn’t see an obvious way around it.

frankieroberto: Doesn’t seem possible:

Screenshot 2024-11-28 at 11 34 25

paulrobertlloyd: Ah, bugger. Looks like the path scoping needs to happen in the branch protection rules, not the workflow.

I wonder if it might be easier to update the visual regression test script to pass if no files have changed instead?

paulrobertlloyd:
Using [tj-actions/changed-files](https://github.com/tj-actions/changed-files) action, we can test for changed files, and only continue to run workflows if a subset of files has changed.

I tested changing the HTML in a component, and the visual regression tests ran (for 3 minutes):

Screenshot 2024-11-28 at 12 13 47

I reverted that change, and that test completed in only 24 seconds:

Screenshot 2024-11-28 at 12 22 35

Using this action, we could revert to having all pull request workflows in one file. @frankieroberto what do you think? One file, or many?

frankieroberto:
@paulrobertlloyd stick with one file for now (easier to review the diff)?

Given that most of the repo is in scope for possible changes which affect the tests, I slightly wonder useful it’ll be, but it would be nice to have speedier ✅ on PRs which  just affect the README or whatever.

.github/workflows/pull-request.yml
```diff
@@ -60,9 +70,17 @@ jobs:
...
+      with:
+        files: |
+          packages/**
+          tests/**
+          app/**
     - name: Install dependencies
+      if: steps.changed-ui_tests.outputs.any_changed == 'true'
       run: npm ci
-    - name: Run tests
-      run: npm test
```
paulrobertlloyd, line 66: @frankieroberto Remove (mostly) duplicate test run here, though potentially given concurrent tests, you’d want this workflow to fail early if there’s a failing test before running expensive visual regression tests?

paulrobertlloyd:
This was mostly solved in #1164 by diffing the build output.

Updated this PR to ignore performing checks if file changes only relate to Markdown or YAML files.

Is this what you were thinking @colinrotherham?

frankieroberto:
@colinrotherham having a separate workflow which reverses the match and marks all the required steps as passed sounds pragmatic (if annoying). It ought to be able to run pretty fast if it’s not even checking out the code?

colinrotherham:
I'm going to close this one, don't think we need to solve it anymore

Things have been (mostly) good since the diff checks went in and solving https://github.com/nhsuk/nhsuk-frontend/pull/1079#issuecomment-2723965028 might be too hard

Thanks @paulrobertlloyd

paulrobertlloyd:
I've been hovering over the close button on this one for a while, thanks for putting this PR out of its misery!

PR #1080, opened 2024-11-25
Fix reverse button text colour

davidhunter08:
## Description

Fixes https://github.com/nhsuk/nhsuk-service-manual-community-backlog/issues/7#issuecomment-2495496548

## Checklist

- [ ] CHANGELOG entry

### DIFF
--- a/app/components/button/reverse.njk
+++ b/app/components/button/reverse.njk
@@ -15,11 +15,6 @@
             "text": "Save and continue",
             "classes": "nhsuk-button--reverse"
           }) }}
-          {{ button({
-            "text": "Save and continue",
-            "classes": "nhsuk-button--reverse",
-            "href": "#"
-          }) }}
         </div>
       </div>
     </main>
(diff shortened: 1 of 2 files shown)


vickytnz (approved):
Works for me after trying out CSS amend locally. 

page with white button and black text

I'm not familiar enough with NHS styling to know if default reverse text colour is supposed to be black rather than NHS blue but this is not my call, I'm just happy that I can now read it!

frankieroberto (dismissed):
Nice, this fixes the bug (caused by `.nhsuk-link:visited` overriding `.nhuk-link--reverse` as it's more specific).

frankieroberto:
@vickytnz it’s black in the [Design System reversed button](https://service-manual.nhs.uk/design-system/components/buttons#reverse-button) and on the [NHS App design system page](https://design-system.nhsapp.service.nhs.uk). I think it works better than if it was blue on white like the GOV.UK one, which somehow seems less visible when on a blue background?

PR #1083, opened 2024-11-27
Change primaryLinks to use href and text instead of url and label

frankieroberto:
Currently we have a mix of components which use `href` and `text` for links (Action link, Back link, Breadcrumbs, Card, Contents list, Summary list, Skip link, Task list) and those which use `url` and `label` (Header and Footer)

This is pretty confusing, and caught me out again the other day.

Given we’re changing the Header component anyway, maybe now is the time to switch it over to `href` and `text` for improved consistency?

We could continue to support the old param names (as this PR current does) for backwards-compatibility, or drop them and advise people to switch over straight away?


paulrobertlloyd:
Yes, been meaning to suggest this. Worth making the corresponding change to the footer component too?

While we're at it, shall we change `primaryLinks` to `navigation`?

edwardhorsford (approved):
Confirm this is not a breaking change?

frankieroberto:
@edwardhorsford depends if we want to keep it backwards-compatible or not (currently in the PR it is). But I was thinking that this would go out with all the other Header changes in the next breaking change release – this PR is to merge into the other PR for that, #1058.

frankieroberto:
I'll merge this into #1058 for now, and we can decide whether to keep the backwards-compatibility later.

PR #1105, opened 2025-01-23
Make it easier to set checkbox radio checked options

frankieroberto:
This adds a new param to both the Checkboxes and Radios components to make it easier to set which items should be checked by default.

This behaviour matches the Checkbox and Radios components within GOV.UK Frontend.

## Description

Currently to set which radio or checkbox items should be pre-ticked, you have to set a `checked` boolean on the relevant items.

When using the NHS prototype kit, this means code like this:

```njk
{{ radios({
  idPrefix: "injectionSite",
  name: "injectionSite",
  fieldset: {
    legend: {
      text: "Where did you give the injection?",
      classes: "nhsuk-fieldset__legend--l",
      isPageHeading: true
    }
  },
  items: [
    {
      value: "Left arm",
      text: "Left arm",
      checked: (data.injectionSite == 'Left arm')
    },
    {
      value: "Right arm",
      text: "Right arm",
      checked: (data.injectionSite == 'Right arm')
    }
]}) }}
```

This pull request makes this simpler and less error-prone by allowing you to set the current selected value, and having the nunjucks macro set the matching item as checked, like this:

```njk
{{ radios({
  idPrefix: "injectionSite",
  name: "injectionSite",
  fieldset: {
    legend: {
      text: "Where did you give the injection?",
      classes: "nhsuk-fieldset__legend--l",
      isPageHeading: true
    }
  },
  value: data.injectionSite,
  items: [
    {
      value: "Left arm",
      text: "Left arm"
    },
    {
      value: "Right arm",
      text: "Right arm"
    }
]}) }}
```

A similar feature has been added for checkboxes, except that the param is called `values`, and works with an array of values which should be checked:

```njk
{{ checkboxes({
  idPrefix: "eligibility",
  name: "eligibility",
  fieldset: {
    legend: {
      text: "Why are you giving them the vaccine?",
      classes: "nhsuk-fieldset__legend--l",
      isPageHeading: true
    }
  },
  values: data.eligibility,
  items: [
    {
      value: "age-based"
      text: "Based on age"
    },
    {
      value: "pregnant"
      text: "Pregnant"
    },
    {
      value: "immunosuppressed",
      text: "Immunosuppressed"
    }
  ]
}) }}
```

## Checklist

- [x] Tested against our [testing policy](https://github.com/nhsuk/nhsuk-frontend/blob/master/docs/contributing/testing.md) (Resolution, Browser & Accessibility)
[...]

PR #1112, opened 2025-02-03
Remove the -1 suffix from radio and checkbox IDs

frankieroberto:
This follows the convention from GOV.UK Frontend and makes it slightly easier to link from error summaries to the first item, as you’d no longer have to remember the `-1` suffix.

This should probably be considered a breaking change though?

## Checklist

- [ ] CHANGELOG entry

### DIFF
--- a/packages/components/checkboxes/template.njk
+++ b/packages/components/checkboxes/template.njk
@@ -45,7 +45,7 @@
   <div class="nhsuk-checkboxes {%- if params.classes %} {{ params.classes }}{% endif %} {%- if isConditional %} nhsuk-checkboxes--conditional{% endif -%}"
     {{- nhsukAttributes(params.attributes) }}>
     {% for item in params.items %}
-    {% set id = item.id if item.id else idPrefix + "-" + loop.index %}
+    {% set id = item.id if item.id else idPrefix + ("-" + loop.index if loop.index > 1 else "")  %}
     {% set name = item.name if item.name else params.name %}
     {% set conditionalId = "conditional-" + id %}
     {% set hasHint = true if item.hint.text or item.hint.html %}
--- a/packages/components/radios/template.njk
+++ b/packages/components/radios/template.njk
@@ -46,7 +46,7 @@
   {%- if params.classes %} {{ params.classes }}{% endif %} {%- if isConditional %} nhsuk-radios--conditional{% endif -%}"
   {{- nhsukAttributes(params.attributes) }}>
     {%- for item in params.items %}
-    {%- set id = item.id if item.id else idPrefix + "-" + loop.index %}
+    {%- set id = item.id if item.id else idPrefix + ("-" + loop.index if loop.index > 1 else "") %}
     {% set conditionalId = "conditional-" + id %}
     {%- if item.divider %}
     <div class="nhsuk-radios__divider">{{ item.divider }}</div>


frankieroberto:
Suggested by @paulrobertlloyd in #1105.

frankieroberto:
Example of the equivalent code in govuk-frontend: https://github.com/alphagov/govuk-frontend/blob/main/packages/govuk-frontend/src/govuk/components/checkboxes/template.njk#L24

paulrobertlloyd:
Thanks @frankieroberto! :shipit:

CHANGELOG.md
```diff
@@ -148,6 +148,42 @@ Use either point 22 or point 26 instead.
 
 This change was introduced in [#1139: Remove 24px from typography scale](https://github.com/nhsuk/nhsuk-frontend/pull/1139)
 
+#### Update links in the error summary to the first checkbox or radio item
+
+The default HTML `id` for the first checkbox or radio item no longer has a `-1` suffix.
+
+This means you should change your error summary from:
```
colinrotherham: For some of our changelog entries, it's nice to have a quick exit when the breaking change doesn't apply

E.g. The GOV.UK Frontend entry starts with "If you've linked from an error summary component…"

Similarly, can we widen the instructions so they don't look Nunjucks specific?

```suggestion
#### Update links in the error summary to the first checkbox or radio item

If you've linked from an error summary component to the first input in a [radios](https://design-system.service.gov.uk/components/radios/) or [checkboxes](https://design-system.service.gov.uk/components/checkboxes/) component, the link will no longer work.

This is because the `id` of the first input no longer has the suffix `-1`.

If there are links back to radios or checkboxes components in your error summary component, remove `-1` from the end of the `href` attribute.

For example, if you're using the `errorSummary` Nunjucks macro:
```

CHANGELOG.md
```diff
@@ -148,6 +148,42 @@ Use either point 22 or point 26 instead.
...
+
```
colinrotherham, line 160: Shall we use "Before:" and "After:" like the others?

```suggestion

Before:

```

CHANGELOG.md
```diff
@@ -148,6 +148,42 @@ Use either point 22 or point 26 instead.
...
+  errorList: [
+    {
+      text: "Select how you like to be contacted",
+      href: "#contact-preference-1"
+    }
+  ]
+}) }}
+```
+
+to
```
colinrotherham:
```suggestion
After:
```

CHANGELOG.md
colinrotherham: Thanks @frankieroberto!

frankieroberto: It's tricky, as it also doesn't apply at all if you select explicit `id`s for each radio or checkbox item, which some of our examples do, eg https://service-manual.nhs.uk/design-system/components/error-summary

It also doesn’t necessarily apply if you’re using a non-Nunjucks implementation, which might apply its own convention for `id`s...

CHANGELOG.md
frankieroberto: Done.

CHANGELOG.md
frankieroberto: Have had a go at tweaking - let me know what you think (and feel free to commit any further updates direct to the branch).

colinrotherham: I felt like the links to the affected components were important, so I've put those in.

I've clarified in https://github.com/nhsuk/nhsuk-frontend/pull/1112/commits/8c70983c0037ee298119c5f1862368cab5ea4104 that it's "…when rendered using the Nunjucks macros"

PR #1113, opened 2025-02-05
Add support for text param for Inset text

### LINKED ISSUE #1108: Inset text does not support a text param
The [Inset text component](https://service-manual.nhs.uk/design-system/components/inset-text) only supports an `html` param at the moment, which confused me when I was trying to use it locally with `text`.

Supporting a `text` param (which gets wrapped in a `<p>`) would be more consistent with other components, which always support both, and with the [govuk Inset text component](https://design-system.service.gov.uk/components/inset-text/).

frankieroberto:
This adds support for a `text` param for the Inset text component, for use where only a single paragraph is included. Resolves #1108

This means that instead of:

```njk
{{ insetText({
  html: "<p>This is a note.</p>"
}) }}
```

you can do this:

```njk
{{ insetText({
  text: "This is a note."
}) }}
```

When the `text` param is used, the text is enclosed by paragraph tags.

## Checklist

- [x] CHANGELOG entry


packages/components/inset-text/template.njk
```diff
@@ -5,5 +5,9 @@
 {{- nhsukAttributes(params.attributes) }}>
   <span class="nhsuk-u-visually-hidden">Information: </span>
   {# params.HTML supported for backwards compatibility - see issue #950 #}
-  {{ (params.html or params.HTML) | safe }}
+  {% if params.html or params.HTML %}
+    {{ (params.html or params.HTML) | safe }}
+  {% else %}
+    <p>{{ params.text }}</p>
```
edwardhorsford, line 11: This will end up outputting an empty paragraph if no args are passed - probably fine though?

edwardhorsford (approved):

packages/components/inset-text/template.njk
```diff
@@ -5,5 +5,9 @@
 {{- nhsukAttributes(params.attributes) }}>
   <span class="nhsuk-u-visually-hidden">Information: </span>
   {# params.HTML supported for backwards compatibility - see issue #950 #}
-  {{ (params.html or params.HTML) | safe }}
+  {% if params.html or params.HTML %}
+    {{ (params.html or params.HTML) | safe }}
+  {% else %}
```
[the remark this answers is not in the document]
frankieroberto: @colinrotherham not sure - I don’t think we’re currently super consistent in whether we try to output nice indenting or not? Can't do any harm though.

packages/components/inset-text/template.njk
frankieroberto: @edwardhorsford I think so? We don't really have any way of handling missing arguments gracefully across all the components.

anandamaryon1 (approved):

Issue #1115, opened 2025-02-10
Warnings during initial `npm install`


colinrotherham:
# Bug Report

## What is the issue?

Noticed a few npm `ERESOLVE` warnings on the first `npm install` following initial setup:

```console
npm warn ERESOLVE overriding peer dependency
npm warn While resolving: eslint-config-standard@17.1.0
npm warn Found: eslint-plugin-n@17.7.0
npm warn node_modules/eslint-plugin-n
npm warn   eslint-plugin-n@"^17.5.1" from @logux/eslint-config@53.1.0
npm warn   node_modules/@logux/eslint-config
npm warn     peer @logux/eslint-config@">=36.0.0" from eslint-config-postcss@4.0.0
npm warn     node_modules/eslint-config-postcss
npm warn
npm warn Could not resolve dependency:
npm warn peer eslint-plugin-n@"^15.0.0 || ^16.0.0 " from eslint-config-standard@17.1.0
npm warn node_modules/eslint-config-standard
npm warn   eslint-config-standard@"^17.1.0" from @logux/eslint-config@53.1.0
npm warn   node_modules/@logux/eslint-config
npm warn
npm warn Conflicting peer dependency: eslint-plugin-n@16.6.2
npm warn node_modules/eslint-plugin-n
npm warn   peer eslint-plugin-n@"^15.0.0 || ^16.0.0 " from eslint-config-standard@17.1.0
npm warn   node_modules/eslint-config-standard
npm warn     eslint-config-standard@"^17.1.0" from @logux/eslint-config@53.1.0
npm warn     node_modules/@logux/eslint-config
```

These are caused by a conflict in the (unused) `eslint-config-postcss` package:

```console
└─┬ eslint-config-postcss@4.0.0
  └─┬ @logux/eslint-config@53.1.0
    ├─┬ eslint-config-standard@17.1.0
    │ └── eslint-plugin-n@17.7.0 deduped invalid: "^15.0.0 || ^16.0.0 " from node_modules/eslint-config-standard
    └── eslint-plugin-n@17.7.0 invalid: "^15.0.0 || ^16.0.0 " from node_modules/eslint-config-standard
```

## What steps are required to reproduce the issue?

1. Clone `nhsuk-frontend`
2. Run `npm install`

## What was the environment where this issue occurred?

* Device: Mac mini
* Operating System: macOS 15.3
* Browser: N/A
* Browser version: N/A
[...]

PR #1116, opened 2025-02-10
Fix warnings during initial `npm install`

colinrotherham:
## Description

This PR removes all unused npm packages + runs `npm audit fix` to consolidate a few outdated packages

Fixes https://github.com/nhsuk/nhsuk-frontend/issues/1115

## Checklist

- [ ] CHANGELOG entry

☝️ Let me know if a CHANGELOG entry is needed

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3953,10 +3953,11 @@
       }
     },
     "node_modules/axios": {
-      "version": "1.7.2",
-      "resolved": "https://registry.npmjs.org/axios/-/axios-1.7.2.tgz",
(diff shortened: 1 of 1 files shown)


frankieroberto:
Looks like [Husky](https://typicode.github.io/husky/) is [used in the NHS Service Manual repo](https://github.com/nhsuk/nhsuk-service-manual/tree/main/.husky) but not here... 🤔

frankieroberto:
@colinrotherham I’ve found in the past that pre-commit hooks can cause issues/annoyances (especially with less technical contributors), and don’t always play nice with apps like GitHub Desktop or whatever. However it might be a nice thing to document as an optional local config?

The other thing I’ve done in the past is set up auto-fixing of style issues using GitHub Actions (see [example](https://github.com/x-govuk/govuk-services-list/blob/main/.github/workflows/json-tidy.yml)) which has some advantages, although doesn’t work with pull requests from forks (as the action doesn’t have permission to push commits to those).

PR #1138, opened 2025-02-14
Left align header navigation items by default

### LINKED ISSUE #1093: Header responsive layout breaks with JS disabled
# Bug Report

## What is the issue?

The responsive layout without JS is a bit broken, with the nav links overflowing rather than wrapping.
image

Ideally we should make the header work without JS enabled and progressively enhancing the responsive layout via JS.

Disable JS and view on a small screen.

paulrobertlloyd:
## Description

- Left align items in the header navigation
- Provide a new modifier class, `.nhsuk-header__navigation-list--justified` which restores the previous behaviour.
- Remove the previous modifier class, `.nhsuk-header__navigation-list--left-aligned`, as this presentation is now the default.
- Increase the space between navigation items from 24px to 32px above the `desktop` breakpoint, which is a little closer the previous design (items look a bit too closer together above this breakpoint without making this change)
- Fixes #1093, navigation items not wrapping if JavaScript is not enabled
- Fixes issue with menu button being shown too early or not early enough on smaller screens
- Fixes issue with older browsers (namely Safari 14 device-locked to older iPads) where `column-gap` is not supported on flex containers. Items are now separated by inline padding on navigation list items 

## Checklist

- [x] CHANGELOG entry


frankieroberto (approved):
Love it.

anandamaryon1:
Thanks @paulrobertlloyd this is great, and bonus that it fixes #1093 !

Testing it, I find that items are being hidden behind the 'More' link earlier than they should? (Pretty sure there is enough room to not hide the last item)

Maybe a corresponding change to the JS needed?

This branch:
image

Header breaking branch (without this chnage):
image

paulrobertlloyd:
@anandamaryon1 Good catch! There was a simple fix to this, but the more I played around resizing the browser, I uncovered other issues.

So, this PR now also addresses a number of other issues relating to navigation items, such as showing the menu button correctly, not too early or too late, occasionally showing more navigation items on smaller screen, and uses padding instead of `column-gap` to separate items so this part of the header should appear less broken in Safari 14/older iPads.

anandamaryon1:
@paulrobertlloyd 
Thanks Paul, nice work. 

Showing the Menu button depending on viewport width seems to be working nicely now. 

Spacing older on Safari's also looking improved 👍 
image

On the `.nhsuk-header__navigation-list--justified` class, is there a way to add that as a Nunjucks option? It seems currently that if you want a justified nav you'd have to use only HTML? (unless I'm missing something)

frankieroberto:
@paulrobertlloyd @anandamaryon1 I think it’s fine for it to be based on adding the extra modifier class, but maybe we should also add a way to add extra classes on the navigation list using the Nunjucks macro?

I think it’d just be a case of supporting `params.primaryLinks.classes`, which would be backwards-compatible (and might be useful anyway)?

paulrobertlloyd:
> I think it’d just be a case of supporting params.primaryLinks.classes, which would be backwards-compatible (and might be useful anyway)?

I like this suggestion. Do we want to include support for `params.primaryLinks.classes` as part of this PR?

anandamaryon1:
Leave Nunjucks option for justified out, but leave justified class available.

packages/components/header/_header-white.scss
```diff
@@ -67,7 +67,7 @@
     }
   }
 
-  .nhsuk-header__navigation-list {
+  .nhsuk-navigation {
```
frankieroberto, line 70: Can you remember why this class name changed? I’d have expected all the css to be scoped within `.nhsuk-header*` but I might have forgotten something?

paulrobertlloyd: I haven’t changed any class names, just changed which element these styles are applied to.

The classes used in this component are a bit all over the place (as elsewhere). For example, the header contains:

- `nhsuk-navigation-container`
- `nhsuk-navigation`
- `nhsuk-mobile-menu-container`
- `nhsuk-organisation-*`
- `nhsuk-logo`
- `nhsuk-search__*`

If we want to fix this, perhaps something for a different PR?

PR #1139, opened 2025-02-14
Remove 24px from typography scale

paulrobertlloyd:
This removes the 24 point from the typography scale, after having previously deprecated it in #1294.

## Reasons for removing it

- 24px is not listed as font size in the [typography scale](https://service-manual.nhs.uk/design-system/styles/typography)
- The difference between 24px and 26px is slight, and having 3 sizes in the 22px - 26px range, reduces the contrast between heading sizes (which is why the medium heading size was increased from 24px to 26px last year).
- The 24px font size was only used for a few components, and these have now been updated:
  - Panel headings (feature card, panel table, warning callout, do and don’t lists)
  - Pagination
  - Panel
  - Lede text (`.nhsuk-lede-text`)
  - Large text (`.nhsuk-body-l`)
- Using 24px for panel headings means these don’t align with other headings on a page

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -144,7 +144,7 @@ This change ensures consistency with other components, where `text` or `html` pa
 
 This change was made in [pull request #1259: Review legacy Nunjucks params](https://github.com/nhsuk/nhsuk-frontend/pull/1259).
 
-#### 24 point typography size removed
+#### Stop using `nhsuk-u-font-size-24` and '24' on the typography scale
 
 The point 24 (24px large screens, 20px small screens) on the typography scale has been removed, after previously being deprecated.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -146,7 +146,7 @@ This change was made in [pull request #1259: Review legacy Nunjucks params](http
 
 #### Stop using deprecated 24 point on the typography scale
 
-The point 24 (24px large screens, 20px small screens) on the typography scale has been removed, after previously being deprecated.
+The point 24 (24px large screens, 20px small screens) on the typography scale has been removed, after previously being deprecated in [version 9.5.0](https://github.com/nhsuk/nhsuk-frontend/releases/tag/v9.5.0).
 
 Use either point 22 or point 26 instead.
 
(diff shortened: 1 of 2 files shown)


anandamaryon1:
I'm supportive. I left it in there whilst doing the typography edits last year because it was being used by the other components you listed. But you call out that in usage, it now means that various cards and styles can inconsistently use 24 or 26, I'm happy to consolidate.

paulrobertlloyd:
Be good to get this into the next release. Outstanding questions are around our deprecation strategy:

- Do we have a method for deprecating the `24` size in the typography scale?
- Deprecating the `nhsuk-font-size-24` class. Do we want to keep this at 24px for now? Do we maintain support for it going forward, but supply a 26px font size. Or at some point does it stop working?

We could defer such questions to later, and make the simplest change which would be to use 26px font size consistently within components and leave 24 in the typography scale and available via a utility class, albeit left undocumented.

@colinrotherham Any thoughts?

paulrobertlloyd:
> Sorry about the force push, it was 500+ commits behind

No problem. Don’t tell @frankieroberto though, force pushes make him twitchy 😂

frankieroberto:
> No problem. Don’t tell @frankieroberto though, force pushes make him twitchy 😂

Haha. Don’t tell anyone but I do sometimes do force pushes too. But not routinely.

My main complaint is getting notifications that lead to this annoying message:

IMG_2847

frankieroberto (approved):

CHANGELOG.md
```diff
@@ -144,7 +144,7 @@ This change ensures consistency with other components, where `text` or `html` pa
 
 This change was made in [pull request #1259: Review legacy Nunjucks params](https://github.com/nhsuk/nhsuk-frontend/pull/1259).
 
-#### Stop using `nhsuk-u-font-size-24` and '24' on the typography scale
+#### Stop using deprecated 24 point on the typography scale
 
 The point 24 (24px large screens, 20px small screens) on the typography scale has been removed, after previously being deprecated.
```
colinrotherham: Should we say when it was deprecated?

>… previously being deprecated in [NHS.UK frontend v9.5.0](https://github.com/nhsuk/nhsuk-frontend/releases/tag/v9.5.0).

frankieroberto: Good shout, done.

PR #1150, opened 2025-02-19
Run visual regression tests without Docker

colinrotherham:
## Description

This PR runs the BackstopJS visual regression tests without Docker

The following commands:

```console
npm run test:visual
npm run test:visual -- --filter="Action link"
```

Can now be run against another `BASE_URL` like the deployed application:

```console
BASE_URL="https://nhsuk.github.io/nhsuk-frontend" \
  npm run test:visual --ignore-scripts
```

Or combined with `--filter="Action link"` to run specific checks only:

```console
BASE_URL="https://nhsuk.github.io/nhsuk-frontend" \
  npm run test:visual --ignore-scripts -- \
  --filter="Action link"
```

## Checklist

- [ ] CHANGELOG entry

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -77,7 +77,7 @@ jobs:
         run: npm run test:visual --ignore-scripts
 
       - name: Save backstop report
-        if: ${{ always() && steps.changes.outputs.changes == 'true' }}
+        if: ${{ !cancelled() && steps.changes.outputs.changes == 'true' }}
         uses: actions/upload-artifact@v4
         with:
           name: Backstop report
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,57 @@
 
 :new: **New features**
 
-- Output source maps and use minified code in examples ([PR 1152](https://github.com/nhsuk/nhsuk-frontend/pull/1152))
-- Add table cell `classes` and `attributes` param support [PR 1172](https://github.com/nhsuk/nhsuk-frontend/pull/1172)
+#### Stop long words breaking out of components with `nhsuk-u-text-break-word`
+
+We've added a new override class to help display long or unpredictable words on narrow screens, such as an email address entered by a user.
+
+Wrapping content with the `nhsuk-u-text-break-word` class or applying it to table cells will force words that are too long to break onto a new line.
+
+```html
+A confirmation email will be sent to <span class="nhsuk-u-text-break-word">aleksandrina.featherstonehaughwhitehead23@folkestonepharmacy.test.com</span>.
+```
+
+Sass users can also use the `nhsuk-text-break-word` mixin.
+
+This change was introduced in [pull request #1061: Add break-word typography helper](https://github.com/nhsuk/nhsuk-frontend/pull/1061).
+
+#### Add classes and attributes to table cells
+
+You can now add `classes` and `attributes` to table cells, for example using the break-word typography helper:
+
+```njk
+{{ table({
+  caption: "Users",
+  head: [
+    {
+      text: "Name"
+    },
+    {
+      text: "Email address"
+    }
+  ],
+  rows: [
+    [
+      {
+        text: "Aleksandrina Featherstonehaugh-Whitehead",
+        classes: "nhsuk-u-text-break-word"
+      },
+      {
+        text: "aleksandrina.featherstonehaughwhitehead23@folkestonepharmacy.test.com",
+        classes: "nhsuk-u-text-break-word"
+      },
+    ]
+  ]
+}) }}
+```
+
+This was added in [pull request #1172: Add table cell `classes` and `attributes` param support](https://github.com/nhsuk/nhsuk-frontend/pull/1172).
+
+#### Source maps for precompiled files
+
+You can now use [source maps](https://firefox-source-docs.mozilla.org/devtools-user/debugger/how_to/use_a_source_map/index.html) to help identify errors and console messages from NHS.UK Frontend precompiled files.
+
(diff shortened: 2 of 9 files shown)


paulrobertlloyd:
Is there a better time and place to run visual regression tests? This question might be a question more of people and process than of technology and configuration!

For example, while updating the header component, a lot of time (and resources) has been spent checking in and updating backstop images, which more often than not are almost deliberately wrong until the final set of changes has been agreed.

I wonder if there’s a way to do it both locally and in CI, but without the images being stored and updated? Or maybe we need guidance about at which point any updates to backstop images should be checked-in?

.github/workflows/pull-request.yml
```diff
@@ -66,7 +66,7 @@ jobs:
 
   ui_tests:
     name: Visual regression tests
-    runs-on: ubuntu-latest
+    runs-on: macos-latest
```
regularfry, line 69: Is this necessary?  If we believe that the regenerated images should be within threshold because of the changed settings below then we should probably try to stay on the cheaper platform if we can.

tests/backstop/bitmaps_reference/nhsuk-frontend_Card_-_Card_with_image_0_document_0_iPhone_5_SE.png
regularfry, line 1: This one and the corresponding iphone 6 test fail for me:
image

The error output is:
```
[1]       compare | ERROR { requireSameDimensions: true, size: isDifferent, content: 0.24%, threshold: 0.2% }: Card - Card with image nhsuk-frontend_Card_-_Card_with_image_0_document_1_iPhone_6-8.png
[1]       compare | ERROR { requireSameDimensions: true, size: isDifferent, content: 0.23%, threshold: 0.2% }: Card - Card with image nhsuk-frontend_Card_-_Card_with_image_0_document_0_iPhone_5_SE.png
```

regularfry (changes requested):

.github/workflows/pull-request.yml
regularfry: Right you are!

tests/backstop/bitmaps_reference/nhsuk-frontend_Card_-_Card_with_image_0_document_0_iPhone_5_SE.png
regularfry: Well this was a rabbit-hole.

a. I don't have scaling options on the display settings.
b. It turns out that the tests pass if I run `npm run test:visual --  --filter="Card with image"` but *not* if I run the full suite with `npm run test:visual`.

Trying to run with `HEADLESS=false` fails every single test.  From a quick inspection it's a scaling issue, but it's not consistent.  For instance:
image
image

I *think* the inconsistency there is that the chromium window has a minimum width greater than 320px.  So the assumption would be that unless the chromium window hits its lower limit, the image is smaller than it should be.

As a further piece of information, if I switch the `Reduce motion` accessibility setting on to limit window animations, only *one* of the tests fails.  Not both.

So bizarre though it looks, I think the problem is applying window scaling to headless renders.  It's not chromium animations, I've tried turning them off. The window scale animation is effectively a zoom from zero to the eventual window dimensions over a short time period.  I think what we're seeing in my headless failures are anti-aliasing artefacts from the fact that before the end of the animation, the page is rendered into a buffer smaller than its eventual dimensions, and then the buffer gets scaled up either by the desktop compositor or the browser itself.  The fact that the tests pass when run on their own means that if the CPU/GPU aren't under pressure, they're able to get to the full-scale frame before backstop decides to take the snapshot. But if they get run in the middle of the suite, the CPU and GPU are already busy, and the animation slows down enough to cause a problem.

This is consistent with what's going on with the `HEADLESS=false` tests, because they're *all* failing as though the snapshot was taken before the new window was at full size.  The `delay` parameter doesn't seem to do anything interesting: I think it's intended to paper over any server start warmup than browser load times.

I do have a fix, but I'm not sure I like it: all the tests pass with `--disable-gpu` in the engine options.  That costs an additional 50% run-time, going from roughly a minute to roughly 1m30.

backstop.config.js
```diff
@@ -1,23 +1,33 @@
...
+      '--disable-skia-runtime-opts',
+      '--force-device-scale-factor=1'
+    ],
     browser: 'chromium',
-    gotoParameters: { waitUntil: 'load' }
+    chromePath: executablePath(),
+    gotoParameters: { waitUntil: 'load' },
+
+    // Allow headless mode switching using `HEADLESS=false`
+    asyncCaptureLimit: HEADLESS ? 1 : 4,
```
regularfry, line 26: I've run into two problems here.  First, the `asyncCaptureLimit` setting needs to be at the top level or it has no effect.  Second, you're taking the truth value of a string, not a boolean: if you run `HEADLESS=false npm run...` then `HEADLESS ? 1 : 4` evaluates to 1, not 4.  It needs to be:

```
  asyncCaptureLimit: HEADLESS === "false" ? 4 : 1,
```

regularfry (changes requested):

tests/backstop/bitmaps_reference/nhsuk-frontend_Card_-_Card_with_image_0_document_0_iPhone_5_SE.png
regularfry: I should clarify: `--disable-gpu` fixes the headless tests.  The `HEADLESS=false` tests still all fail for me.

regularfry: No joy, `HEADLESS=false` are all still failing at #8eafcfdd.  I think it's still a scaling problem.  Here's an example of what the diffs look like:

image

Issue #1151, opened 2025-02-20
Puppeteer and visual tests do not check bundled code


colinrotherham:
# Bug Report

## What is the issue?

Puppeteer and visual tests do not currently check bundled code

Issues affecting build tool output won't be picked up, for example:

### JavaScript build

* Babel transforms
* Terser minification
* Webpack bundling

### Stylesheet build

* PostCSS transforms (Autoprefixer)
* CSSNANO minification

## What steps are required to reproduce the issue?

1. Add breaking change to `nhsuk-9.3.0.min.css` or `nhsuk-9.3.0.min.js`
2. All tests continue to pass

## Is there anything else you think would be useful in recreating the issue?

Switching the documentation site to use minified bundles would catch the issues above

See how GOV.UK Frontend switched their review app to use minified code in:

* https://github.com/alphagov/govuk-frontend/pull/3050

This helped discover an Internet Explorer issue caused by incompatible minification:

* https://github.com/alphagov/govuk-frontend/issues/3136

PR #1152, opened 2025-02-20
Output source maps and use minified code in examples

colinrotherham:
## Description

This PR updates the [documentation site](https://nhsuk.github.io/nhsuk-frontend/) to use minified bundles which fixes https://github.com/nhsuk/nhsuk-frontend/issues/1151

Source maps have been enabled to preserve original sources when inspecting code

As a bonus, we no longer need a separate `build` (non-minified) versus `bundle` (minified) build tasks

Stylesheet source maps

JavaScript source maps

## Checklist

- [x] CHANGELOG entry

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -10,6 +10,7 @@ const postcss = require('gulp-postcss')
 const rename = require('gulp-rename')
 const gulpSass = require('gulp-sass')
 const terser = require('gulp-terser')
+const PluginError = require('plugin-error')
 const dartSass = require('sass')
 const webpack = require('webpack-stream')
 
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+:new: **New features**
+
+- Output source maps and use minified code in examples ([PR 1152](https://github.com/nhsuk/nhsuk-frontend/pull/1152))
+
```
frankieroberto, line 8: Not sure this needs a changelog entry as it doesn't affect downstream services at all? (And it could be confusing?) 🤷

frankieroberto: Ah I guess it the npm package includes them then it's ok to mention? 

I had assumed most services import the source from the package and recompile it together with their own stuff - but I guess you could also use the precompiled versions as in Option 2: https://service-manual.nhs.uk/design-system/production

frankieroberto (approved):

frankieroberto (approved):

PR #1155, opened 2025-02-22
Support form submission from header account items

paulrobertlloyd:
## Description

Some items in the header account area, but most likely any ‘Log out’ button, will require form submission rather than linking to a page.

This PR adds support for having button items (in addition to having a link or only text), and adds the following 2 options:

`[]item.action` – The form action, for example `"/log-out"`
`[]item.method` – The form method, optional and defaults to `"post"`

If you provide `item.action`, a button will be used instead. Using `item.href` will override any `item.action` parameter.

These options currently presume the presence of a surrounding `<form>`; perhaps we should add a `<form>` element if you use one or more `item.action` parameters?

The button is styled to look and behave like a link:

<p>Screenshot of header navigation with ‘Log out’ button styled as a link.</p>

If this PR is accepted, we’d need to update the service manual guidance… and possibly add some information around when you might want to use a button instead of a link.

## Checklist

- [ ] CHANGELOG entry

### DIFF
--- a/packages/components/header/template.njk
+++ b/packages/components/header/template.njk
@@ -13,11 +13,11 @@
 {% endset %}
 
 {% macro _accountItem(item) %}
-{% if item.icon %}
+{%- if item.icon -%}
 <svg class="nhsuk-icon nhsuk-icon__user" xmlns="http://www.w3.org/2000/svg" width="20" height="20" viewBox="0 0 16 16" aria-hidden="true" focusable="false">
   <path d="M8 0c4.4 0 8 3.6 8 8s-3.6 8-8 8-8-3.6-8-8 3.6-8 8-8Zm0 1a7 7 0 1 0 0 14A7 7 0 0 0 8 1Zm-1.5 9h3a2.5 2.5 0 0 1 2.5 2.5V14a1 1 0 0 1-1 1H5a1 1 0 0 1-1-1v-1.5A2.5 2.5 0 0 1 6.5 10ZM8 9C6.368 9 5 7.684 5 6s1.316-3 3-3c1.632 0 3 1.316 3 3S9.632 9 8 9"/>
 </svg>
-{% endif -%}
+{%- endif -%}
 {{- item.html | safe if item.html else item.text -}}
 {% endmacro %}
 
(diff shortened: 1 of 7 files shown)


packages/components/header/template.njk
```diff
@@ -81,6 +81,10 @@
           <a class="nhsuk-header__account-link" href="{{ item.href }}">
             {{ _accountItem(item) }}
           </a>
+        {% elif item.action %}
+          <button class="nhsuk-header__account-button" formaction="{{ item.action }}" formmethod="{{ item.method or "post" }}">
```
frankieroberto, line 85: Must admit I’ve never come across the `formaction` or `formmethod` attributes for buttons, but they’re in the spec and look like they’re well supported?

Does still need a separate `form` element though, as you say.

I did wonder if it’d be simpler for the nunjucks to generate a separate form tag for each of the buttons (in most cases it’d be be log out?) - but then I guess in production you might also have extra hidden inputs for cross-site request forgery (CSRF) tokens or whatever?

frankieroberto, line 85: Should this have ‘role=link‘ so that it behaves more like how it looks (eg will respond to “click link log out” for voice control)? 🤔

frankieroberto (commented):
Nice!

packages/components/header/template.njk
paulrobertlloyd: Precisely. This is the corresponding (Rails) implementation we currently have on Mavis:

```html
<span class="app-header__account-item">
  <form class="button_to" method="post" action="/logout">
    <input type="hidden" name="_method" value="delete" autocomplete="off">
    <button class="app-header__account-button" type="submit">Log out</button>
    <input type="hidden" name="authenticity_token" value="XXX" autocomplete="off">
  </form>
</span>
```

Unsure if adding a `<form>` in the Nunjucks macro if any button includes an action is helpful or not. As Nunjucks is mostly used by prototypes, maybe it is?

paulrobertlloyd: I quickly peeked at the spec, and this role seems aimed at those using non-semantic elements for hyperlinks; a button is a button, no?

frankieroberto: I'm not sure. The NHS button adds `role="button"` to links that are styled as buttons, and the GOVUK button additionally [adds some javascript](https://github.com/alphagov/govuk-frontend/blob/main/packages/govuk-frontend/src/govuk/components/button/button.mjs#L29-L57) so that links-styled-as-buttons get clicked if the spacebar is hit when they’re focused.

Neither the GOVUK nor NHS frontend current supports buttons-styled-like-links at the moment, but there’s a backlog discussion about it here: https://github.com/alphagov/govuk-design-system-backlog/issues/159

In our case, maybe it’s a bit less relevant as the "Log out" item in the header kinda looks half-way between a link and button (especially if the underlines get removed in #1136?

anandamaryon1:
Add a form to the Nunjucks. After merged into base PR (https://github.com/nhsuk/nhsuk-frontend/pull/1058) then double check button styled as link accessibilty.

PR #1164, opened 2025-03-03
Skip visual regression tests without a diff

colinrotherham:
## Description

This PR skips visual regression tests when there are no `git diff` changes

It does this by comparing `npm run build` between the PR head and base commits

### Future uses

We could extend this in future to:

* Add PR diff comments like in [govuk-frontend/pull/#4511](https://github.com/alphagov/govuk-frontend/pull/4511)
* Run visual regression tests against a previous build without committed screenshots

## Checklist

- [ ] CHANGELOG entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,57 @@
 
 :new: **New features**
 
-- Output source maps and use minified code in examples ([PR 1152](https://github.com/nhsuk/nhsuk-frontend/pull/1152))
-- Add table cell `classes` and `attributes` param support [PR 1172](https://github.com/nhsuk/nhsuk-frontend/pull/1172)
+#### Stop long words breaking out of components with `nhsuk-u-text-break-word`
+
+We've added a new override class to help display long or unpredictable words on narrow screens, such as an email address entered by a user.
+
+Wrapping content with the `nhsuk-u-text-break-word` class or applying it to table cells will force words that are too long to break onto a new line.
+
+```html
+A confirmation email will be sent to <span class="nhsuk-u-text-break-word">aleksandrina.featherstonehaughwhitehead23@folkestonepharmacy.test.com</span>.
+```
+
+Sass users can also use the `nhsuk-text-break-word` mixin.
+
+This change was introduced in [pull request #1061: Add break-word typography helper](https://github.com/nhsuk/nhsuk-frontend/pull/1061).
+
+#### Add classes and attributes to table cells
+
+You can now add `classes` and `attributes` to table cells, for example using the break-word typography helper:
+
+```njk
+{{ table({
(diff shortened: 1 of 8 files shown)


.github/workflows/actions/diff/action.yml
```diff
@@ -0,0 +1,95 @@
...
+      uses: actions/cache/restore@v3
+      id: cache-head
+      with:
+        path: dist/
+        key: build-${{ github.event.pull_request.head.sha }}
+
+    - name: Install dependencies
+      if: ${{ steps.cache-head.outputs.cache-hit == 'true' }}
+      shell: bash
+      run: npm install
```
regularfry, line 66: Why do we need this `npm install`?  We've already got the build (from the cache) and we don't run anything after this point that uses the dependencies, unless I've missed something.

.github/workflows/actions/diff/action.yml
```diff
@@ -0,0 +1,95 @@
...
+runs:
+  using: composite
+
+  steps:
+    - name: Set up diff drivers
+      shell: bash
+      run: |
+        npm install -g js-beautify
+        git config diff.minjs.textconv "js-beautify --editorconfig --type js"
+        git config diff.mincss.textconv "js-beautify --editorconfig --type css"
```
regularfry, line 17: Not clear to me what the function of the diff config (and thus `js-beautify`) here is.  All we see at the end of the process is an exit code, the diff format shouldn't matter.

regularfry (changes requested):

.github/workflows/actions/diff/action.yml
regularfry: Yeah, it's worth avoiding having to wait for it in my mind.

regularfry (dismissed):

.github/workflows/pull-request.yml
```diff
@@ -52,28 +52,32 @@ jobs:
...
+        id: changes
 
       - name: Run backstop
-        run: npm run test:visual
+        if: ${{ steps.changes.outputs.changes == 'true' }}
+        run: npm run test:visual --ignore-scripts
 
       - name: Save backstop report
-        if: ${{ always() }}
+        if: ${{ always() && steps.changes.outputs.changes == 'true' }}
```
regularfry, line 80: Do we need the `always()` here?  It can run when the previous step hasn't, and there's no report to upload.

regularfry (changes requested):

.github/workflows/pull-request.yml
anandamaryon1: Seems like this is done 👍

anandamaryon1 (approved):

Issue #985, opened 2024-07-16
Table component 'firstCellIsHeader' doesn't work with 'responsive: true'


anandamaryon1:
# Bug Report

## What is the issue?
The table option of `firstCellIsHeader: true` doesn't work as expected (making the first row cell a row header) when used in conjunction with `responsive: true`. It seems to do nothing when `responsive: true` is enabled, but works correctly when `responsive:` is set to `false`.

## What steps are required to reproduce the issue?
Create a responsive table and set `firstCellIsHeader: true`.

## Is there anything else you think would be useful in recreating the issue?

| Responsive: true      | Responsive: false |
| ----------- | ----------- |
| ![image](https://github.com/user-attachments/assets/f6388091-9a3b-4465-a980-81aa5096cc5c)      | ![image](https://github.com/user-attachments/assets/2c60fc51-33cd-40f4-b437-0501f4a043ad)       |

PR #1172, opened 2025-03-05
Add table cell `classes` and `attributes` param support

colinrotherham:
## Description

Closes https://github.com/nhsuk/nhsuk-frontend/issues/1171

## Checklist

- [x] Tested against our [testing policy](https://github.com/nhsuk/nhsuk-frontend/blob/master/docs/contributing/testing.md) (Resolution, Browser & Accessibility)
- [x] Follows our [coding standards and style guide](https://github.com/nhsuk/nhsuk-frontend/blob/master/docs/contributing/coding-standards.md)
- [x] CHANGELOG entry

Issue #1171, opened 2025-03-05
Support table column widths like GOV.UK Frontend


colinrotherham:
The Table component currently lacks `classes` and `attributes` params on its cells

Adding these would support [**Custom column widths**](https://design-system.service.gov.uk/components/table/#custom-column-widths) from GOV.UK Frontend

```patch
  caption: "Ibuprofen syrup dosages for children",
  firstCellIsHeader: false,
  head: [
    {
-     text: "Age"
+     text: "Age",
+     classes: "nhsuk-u-width-one-half"
    },
```

PR #1174, opened 2025-03-06
Fix responsive table when `firstCellIsHeader` is set

colinrotherham:
## Description

This PR fixes the [**Responsive table**](https://service-manual.nhs.uk/design-system/components/table#responsive-table) to restore `scope="row"` when `firstCellIsHeader` is enabled

It was split out from https://github.com/nhsuk/nhsuk-frontend/pull/1172 as part of https://github.com/nhsuk/nhsuk-frontend/issues/1171 and fixes https://github.com/nhsuk/nhsuk-frontend/issues/985

```patch
- <td class="nhsuk-table__cell">
+ <th scope="row" class="nhsuk-table__cell">
```

## Checklist

- [x] CHANGELOG entry


anandamaryon1:
Nice, thanks for looking at this one. I don't think when in stacked layout that the bolded `th` is helpful, BUT we're refactoring the responsive table layout in https://github.com/nhsuk/nhsuk-frontend/pull/1156 anyway, and because it swaps to a `dl` for the stacked layout, then this shouldn't be an issue.

anandamaryon1 (approved):
Looks good, thanks!

PR #1180, opened 2025-03-12
Run visual tests on `$mq-breakpoints` targeting `main` only

colinrotherham:
## Description

This PR includes changes to:

1. Run visual regression tests on PRs that target `main` only
2. Update visual regression tests to use NHS breakpoints

Only mobile and desktop included by default (unless configured)

* **mobile** 320px
* **tablet** 641px
* **desktop** 769px
* **large-desktop** 990px

This reduces total images from 420 to only 276

### Before
Viewports include excess height to match device resolutions

Test image before, with excess height

### After
Viewport height is set to 100px minimum

Test image after

## Checklist

- [ ] CHANGELOG entry

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -82,6 +82,8 @@ jobs:
     runs-on: macos-latest
     needs: [build]
 
+    if: ${{ github.base_ref == 'main' }}
+
     steps:
       - name: Checkout
         uses: actions/checkout@v4


.github/workflows/pull-request.yml
```diff
@@ -106,11 +106,13 @@ jobs:
           path: node_modules
 
       - name: Setup Node
+        if: ${{ github.base_ref == 'main' }}
```
frankieroberto, line 109: Could move this up to the job level so the whole job is skipped if the pull request isn't targeting `main`?

frankieroberto: Ah... we could change the branch protection rules so that they only apply for pull requests targeting main I think? 

Looks like it should be possible with the new (more complicated) "Rulesets"?

Screenshot 2025-03-13 at 09 03 05

frankieroberto: @colinrotherham I don’t have admin access to this repo, but @anandamaryon1 does - I asked him to tweak the settings after we split out the tests into parallel jobs last time.

paulrobertlloyd:
Does this replace the need for #1079?

frankieroberto (approved):

PR #1186, opened 2025-03-13
Add template to package

### LINKED ISSUE #1072: Include a page template within the nhsuk-frontend npm package
Currently the Service Manual documents a [Page template](https://service-manual.nhs.uk/design-system/styles/page-template), but a page template is not actually included within the `nhs-frontend` npm package itself.

There is a [layout.njk](https://github.com/nhsuk/nhsuk-frontend/blob/main/app/_templates/layout.njk) file within the repository, but that is used only for the site used for previewing and testing the components, and isn’t part of the published package.

The Prototype kit also has a [template.html](https://github.com/nhsuk/nhsuk-prototype-kit/blob/main/docs/views/template.html) (confusingly in the `docs/` folder) which is similar (but not identical).

I think ideally we’d include a page template within NHS Frontend, and then use that from within the Prototype Kit. Any other downstream users of NHS Frontend which use Nunjucks could also then use that.

frankieroberto:
This adds a `template.njk` Nunjucks page template to the `nhsuk-frontend` npm package, so that services can use (and extend) it rather than having to define their own.

This would be helpful for the NHS Prototype kit, and any other website and live services which use Nunjucks.

The PR also updates the app used to preview the components to use the new template from the package.

Fixes #1072

Largely the template follows the existing guidance for [Page template](https://service-manual.nhs.uk/design-system/styles/page-template) in defining blocks and variables.

I've also compared it to the [GOVUK template](https://github.com/alphagov/govuk-frontend/blob/main/packages/govuk-frontend/src/govuk/template.njk) which it largely mirrors.

One new option is added:

| Option |  Option type | Description |
|--------|--------------|-------------|
| `htmlLang` | Variable | Sets the language of the overall page. Defaults to `en` (English). |
| `htmlAttributes` | Object | Custom attributes on the root `<html>` element |

## Checklist

- [x] CHANGELOG entry


frankieroberto:
@anandamaryon1 @colinrotherham I think this is ready to review. It looks like a bit PR, but it's mostly adding a single `template.njk` file, and the rest is updating all the files in `app/` to use it.

One other beneficiary of this will be the Service Manual website itself, which will be able to update [`layout.njk`](https://github.com/nhsuk/nhsuk-service-manual/blob/main/app/views/includes/layout.njk) to extend the template in the npm package. That’ll simplify it a bit and make it more consistent. (eg that layout file is using an `outerBody` block, which isn’t one of the blocks defined in the page template)

I’ve made minimal changes, but we might want to review some of the default meta tags and icon sizes at some point.

anandamaryon1 (approved):
Nice, a lot of work here, thank you. Happy for this to be merged. I'll make myself a ticket to implement on service manual.

PR #1196, opened 2025-03-21
Add white and blue panel variants

frankieroberto:
This adds 2 new variants of the Panel component, one with a solid blue background, and one with a white background but a blue border.

An additional `headingClasses` param is also added to enable a smaller title size class (`nhsuk-panel__title--s`).

## Blue panel

Screenshot 2025-03-21 at 15 55 38

This is intended for use when you need to stop users in a journey to tell them important information. This should be used vary sparingly, as there is some evidence that users can end up ignoring it, however it could work in some contexts, such as when shown after many regular pages with the off-white background.

The blue panel would be used with the [reversed Button variant](https://service-manual.nhs.uk/design-system/components/buttons#reverse-button) - which already mentions ‘interruption cards’ in the guidance.

See relevant threads for examples and discussion:

* [Stop users in a journey to tell them important information](https://github.com/nhsuk/nhsuk-service-manual-community-backlog/issues/241) on NHS design backlog
* [Interruption card](https://github.com/alphagov/govuk-design-system-backlog/issues/27) on GOVUK design backlog
* [Interruption card](https://design-patterns.service.justice.gov.uk/components/interruption-card/) in MoJ Design System

## White panel

Screenshot 2025-03-21 at 16 03 50

This is intended for secondary content on the page, and would appear above the main content, such as an invitation to complete a survey, or to submit feedback about something you’ve just done.

It could be contain the upcoming new [Secondary button style](https://github.com/nhsuk/nhsuk-frontend/pull/1077).

See example: [Add feedback panel](https://github.com/NHSDigital/record-a-vaccination-prototype/pull/249):

Screenshot 2025-03-21 at 16 10 50

## Checklist

- [ ] CHANGELOG entry

### DIFF
--- a/app/pages/examples.njk
+++ b/app/pages/examples.njk
@@ -106,6 +106,8 @@
     <li><a href="../components/label/page-heading.html">Label as page heading</a></li>
     <li><a href="../components/pagination/index.html">Pagination</a></li>
     <li><a href="../components/panel/index.html">Panel</a></li>
+    <li><a href="../components/panel/blue.html">Panel - blue</a></li>
+    <li><a href="../components/panel/white.html">Panel - white</a></li>
     <li><a href="../components/radios/index.html">Radios</a></li>
     <li><a href="../components/radios/inline.html">Radios inline</a></li>
     <li><a href="../components/radios/disabled.html">Radios disabled</a></li>
--- a/packages/components/panel/_panel.scss
+++ b/packages/components/panel/_panel.scss
@@ -42,11 +42,26 @@ $nhsuk-border-width-panel: nhsuk-spacing(1);
   }
 }
 
+.nhsuk-panel--blue {
+  background-color: $color_nhsuk-blue;
+}
+
+.nhsuk-panel--white {
+  background-color: $color_nhsuk-white;
+  color: $color_nhsuk-black;
+  border-color: $color_nhsuk-blue;
+}
+
 .nhsuk-panel__title {
   @include nhsuk-typography-responsive(48);
   @include nhsuk-responsive-margin(5, "bottom");
 
   margin-top: 0;
+}
+
+.nhsuk-panel__title--s {
+  @include nhsuk-typography-responsive(22);
+  @include nhsuk-responsive-margin(4, "bottom");
 }
 
 .nhsuk-panel__title:last-child {
--- a/app/components/panel/blue.njk
+++ b/app/components/panel/blue.njk
@@ -1,0 +1,26 @@
+{% set title = 'Panel - blue' %}
+{% from 'components/panel/macro.njk' import panel %}
+{% from 'components/button/macro.njk' import button %}
+
+{% extends 'example.njk' %}
+
+{% set panelHtml %}
+  <p class="nhsuk-body nhsuk-u-margin-bottom-6">Some text to explain what you are doing, and to check you’ve understood the consequences.</p>
+  {{ button({
+    text: "I understand, continue anyway",
+    href: "#",
+    classes: "nhsuk-button--reverse"
+  }) }}
+{% endset %}
+
+{% block content %}
+  <div class="nhsuk-grid-row">
+    <div class="nhsuk-grid-column-two-thirds">
+      {{ panel({
+        titleText: "You are doing something unusual",
+        html: panelHtml,
+        classes: "nhsuk-panel--blue"
+      }) }}
+    </div>
+  </div>
+{% endblock %}
--- a/app/components/panel/white.njk
+++ b/app/components/panel/white.njk
@@ -1,0 +1,23 @@
+{% set title = 'Panel - white' %}
+{% from 'components/panel/macro.njk' import panel %}
+{% from 'components/button/macro.njk' import button %}
+
+{% extends 'example.njk' %}
+
+{% set panelHtml %}
+  <p class="nhsuk-body nhsuk-u-margin-bottom-0">This is another short line of content.</p>
+{% endset %}
+
+{% block content %}
+  <div class="nhsuk-grid-row">
+    <div class="nhsuk-grid-column-two-thirds">
+      {{ panel({
+        titleText: "This is a title",
+        headingLevel: 2,
+        headingClasses: "nhsuk-panel__title--s",
+        html: panelHtml,
+        classes: "nhsuk-panel--white"
+      }) }}
+    </div>
+  </div>
+{% endblock %}
--- a/packages/components/panel/template.njk
+++ b/packages/components/panel/template.njk
@@ -5,7 +5,7 @@
 <div class="nhsuk-panel
   {%- if params.classes %} {{ params.classes }}{% endif %}"
   {{- nhsukAttributes(params.attributes) }}>
-  <h{{ headingLevel }} class="nhsuk-panel__title">
+  <h{{ headingLevel }} class="nhsuk-panel__title{% if params.headingClasses %} {{ params.headingClasses }}{% endif %}">
     {{ params.titleHtml | safe if params.titleHtml else params.titleText }}
   </h{{ headingLevel }}>
   {% if caller or params.html or params.text %}


app/pages/examples.njk
```diff
@@ -106,6 +106,8 @@
     <li><a href="../components/label/page-heading.html">Label as page heading</a></li>
     <li><a href="../components/pagination/index.html">Pagination</a></li>
     <li><a href="../components/panel/index.html">Panel</a></li>
+    <li><a href="../components/panel/blue.html">Panel - blue</a></li>
+    <li><a href="../components/panel/white.html">Panel - white</a></li>
```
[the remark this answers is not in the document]
frankieroberto: Done, ta.

packages/components/panel/_panel.scss
```diff
@@ -42,13 +42,28 @@ $nhsuk-border-width-panel: nhsuk-spacing(1);
...
+}
+
 .nhsuk-panel__title {
   @include nhsuk-typography-responsive(48);
   @include nhsuk-responsive-margin(5, "bottom");
 
   margin-top: 0;
 }
 
+.nhsuk-panel__title--s {
```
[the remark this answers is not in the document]
frankieroberto: Yeah, maybe?!

I did think that maybe the existing `nhsuk-heading-s` class would work, but that gets overridden by `nhsuk-panel__title` (probably source order) so the panel titles need their own modifiers - which is probably better anyway?

paulrobertlloyd:
Initial gut reaction; yes to the blue variant, however the white variant is something that appears to be _very_ different, and a component that possibly overlaps with a few others (card, notifications, warnings…).

frankieroberto:
@paulrobertlloyd the white one is definitely 'experimental'... 😂 

The [NHS warning callout](https://service-manual.nhs.uk/design-system/components/warning-callout) is quite different, and we don’t have a notification component (yet - maybe we should). The [Card component](https://service-manual.nhs.uk/design-system/components/card) is quite similar though (basically a different border) - but maybe that’s more designed for where the whole thing is clickable and you have several of them tiled up? (That said, there is a non-clickable one which I forgot about...)

frankieroberto:
Here’s our feedback panel using the existing card rather than the white panel variant:

Screenshot 2025-03-21 at 17 03 45

Kinda works, but maybe stands out a bit less? 🤔

packages/components/panel/_panel.scss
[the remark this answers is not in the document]
frankieroberto: not sure! `nhsuk-panel__title--s` _feels_ more better intuitive somehow, especially with title vs heading - but I could be persuaded either way.

michaelgallagher:
re: the white card, worth considering [this](https://github.com/nhsuk/nhsuk-service-manual-community-backlog/issues/504)?

re: the blue card, minor thing, but should there be less padding between the button and bottom of the card?

frankieroberto:
> re: the white card, worth considering [this](https://github.com/nhsuk/nhsuk-service-manual-community-backlog/issues/504)?

Ooh I hadn’t seen that. Although in our case we’d need to use the secondary button as we already have a green primary button further down the page. And to me the blue border pairs nicely with the new blue outlined secondary button.

> re: the blue card, minor thing, but should there be less padding between the button and bottom of the card?

Yeah I think so. It’s caused by the extra default spacing below the button - but that can be reduced.

anandamaryon1:
Nice, blue panel for interruption cards is helpful. With the minor tweak to button margin-bottom, `nhsuk-u-margin-bottom-3` seems about right to my eyes (accounting for the button shadow and the heading text increasing the padding at the top visually). 

Inclined to agree with others that the white variant doesn't seem to fit with the panel component.

packages/components/panel/_panel.scss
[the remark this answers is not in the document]
anandamaryon1: Agree, if we're going to include heading sizes, it makes sense to include `s`, `m` and `l`.

sarawilcox:
I think the blue one is fine. We've had teams using something like this for interruption cards for some time. Like other people, I'm not sure about the white one.

packages/components/panel/_panel.scss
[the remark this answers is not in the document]
frankieroberto: I’ve dropped this for now, and removed the headingClasses param, as I'm not sure it’s needed for the Interruption card (default h1 size should be ok).

frankieroberto:
@anandamaryon1 @paulrobertlloyd @colinrotherham @sarawilcox @michaelgallagher ok I’ve updated this pull request to purely add the blue panel variant (aka Interruption card).

Also reduced the margin below the button, although this is purely done using an `nhsuk-u-margin-bottom-3` override class rather than being automatic.

As the interruption card would normally be the only thing on the page, I think there’s no immediate need to support `headingClasses` in order to reduce the font size of the h1?

anandamaryon1:
Looking at this and reading some backlog discussions about interruption cards, I've seen some suggestions to use start pages instead, as a way to break up a journey and provide some new context (e.g https://github.com/nhsuk/nhsuk-service-manual-community-backlog/issues/241#issuecomment-632587931 & https://github.com/nhsuk/nhsapp-frontend/issues/19#issuecomment-2501391420). These seem reasonable suggestions. 

They would not be suitable for what I'd think of as 'system' messages, where the system has to tell you something, e.g something has changed and now you need to do something, but in those cases a notification banner seems more appropriate?

The difference between this interruption card and a 'normal' page is just the visual treatment, which makes you stop and think twice. I'm wondering whether we have any evidenced use-cases for this?

frankieroberto:
@anandamaryon1 I think there’s definitely a valid debate about whether you need the different visual treatment of a blue background, or whether just a regular content page with a green button with a distinct label is sufficient.

I don’t have any strong evidence for the need for the blue background personally, but others might? I think the MOJ Design system team did some research on it (see [their page](https://design-patterns.service.justice.gov.uk/components/interruption-card/)) and @edwardhorsford might remember the research from back on the passports service?

There would likely have to be some clear guidance to encourage people not to over-use it though, as I suspect part of the value of the blue background is it being distinct from the usual pattern and infrequently seen?

Do we know why the ["reverse" button](https://service-manual.nhs.uk/design-system/components/buttons#reverse-button) was first added to the NHS design system? The guidance for that already mentions interruption cards, but no other context offered!

> White buttons on solid background colour are good on components like interruption cards where link text, primary and secondary buttons would be lost.

paulmsmith:
111 online has at least two patterns that I would say fits the pattern use-case of "need to stop users in a journey to tell them important information."

**Module Zero**

image

**Some notes about this pattern:** 

- Referred to internally as Module zero or "Mod Zero", that name is taken from the Pathways software tool which is organised into 'modules'. 
- I think when describing it in our terms it is probably better called something like the 'vital to confirm' interrupt or even just as the content says the "Check it's not an emergency" interrupt.
- Research showed the need for a strong visual difference from the rest of the flow.
- It is actually a bit of a hack/work-around because of how long it would take to walk people though those conditions/causes-for-concern page-by-page.there was an assumption that people in a life threatening emergency were unlikely to go online
- Mod zero content is there as a "just in case"

I'm going to invite @LaurenMcAllister to contribute about her work on what we lovingly call the 'weird questions' interrupt.

frankieroberto:
@colinrotherham @anandamaryon1 I've fixed the conflicts with main and updated this to fit our new code structure.

Do we still think this could be released, or shall we wait for more research?

@paulmsmith any update on that 111 emergency panel pattern?

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # NHS.UK frontend Changelog
 
+## Unreleased
+
+### New features
```
colinrotherham: I kept the emoji (and the bold is redundant, but a different colour in code)

```suggestion
### :new: **New features**
```

CHANGELOG.md
colinrotherham, line 1: Do you mind undoing the "Trim Trailing Whitespace" changes?

Probably keep it turned off and let Prettier handle it?

We've used it in a couple of places to [preserve line breaks](https://www.markdownguide.org/basic-syntax/#line-break-best-practices)

packages/nhsuk-frontend/src/nhsuk/components/panel/_index.scss
```diff
@@ -50,6 +50,10 @@ $nhsuk-border-width-panel: $nhsuk-panel-border-width;
     }
   }
 
+  .nhsuk-panel--blue {
```
paulrobertlloyd: I don’t like that this references a colour (which could conceivably change, though unlikely).

It’s shame we can’t mimic the classes we use for notification banners (where the green variant has a `--success` modifier), but I guess it makes sense that you only provide modifiers for lesser used component variants.

Given that, maybe we can use either `--information` or `--interruption` for the blue variant of this panel?

paulrobertlloyd:
I’m for us adding this in, but I think the design is the easiest part of adding this, whereas getting the guidance right will be crucial. It’s odd that this is such a well used pattern across government, but only added to the MoJ Design System. Maybe their guidance can be a starting point for ours?

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
...
+
+#### Blue panel variant
+
+There is a new variant of the Panel component with a solid blue background and white text. This can be used as an interruption card. [Pull request #1196](https://github.com/nhsuk/nhsuk-frontend/pull/1196)
```
colinrotherham: I think we lowercase the component names when inline?

Looks good with the PR link split out for consistency

```suggestion
We've added a new variant of the panel component with a solid blue background and white text. This can be used as an interruption card.
  
This was added in [pull request #1196: Add blue panel (Interruption card) variant](https://github.com/nhsuk/nhsuk-frontend/pull/1196).
```

packages/nhsuk-frontend/src/nhsuk/components/panel/macro-options.mjs
```diff
@@ -76,6 +76,16 @@ export const examples = {
     screenshot: {
       viewports: ['mobile', 'tablet', 'desktop']
     }
+  },
+  blue: {
+    context: {
+      titleText: 'You are doing something unusual',
+      html: '<p>Some text to explain.</p><a href="#" class="nhsuk-button nhsuk-button--reverse" data-module="nhsuk-button">Continue anyway</a>',
```
colinrotherham: Would you mind using our render helper so the button HTML comes from Nunjucks?

```suggestion
      html: outdent`
        <p>Some text to explain.</p>
        ${components.render('button', {
          context: {
            text: 'Continue anyway',
            classes: 'nhsuk-button--reverse',
            href: '#'
          }
        })}
      `,
```

<br>

You'll need to import outdent:

```patch
  import { components } from '@nhsuk/frontend-lib'
+ import { outdent } from 'outdent'
```

CHANGELOG.md
frankieroberto: Ahh, I've got my editor set up to do it automatically so didn't even notice... Will undo!

packages/nhsuk-frontend/src/nhsuk/components/panel/macro-options.mjs
frankieroberto: @colinrotherham aha yes! Didn't realise this was possible! 😄

frankieroberto: Done in 6d348781129e5464796c05308d0382bdb5405192

CHANGELOG.md
frankieroberto: Done. Not one trailing space but two!

packages/nhsuk-frontend/src/nhsuk/components/panel/_index.scss
frankieroberto: I wasn't sure what to call it either. `--interruption` would get my vote I think.

It will probably get used for non-interruption purposes, but maybe in using the `--interruption` modifier and having strong guidance we can try and limit its use? (Given that part of the point is for it to be unusual)

frankieroberto:
I've just pushed b5ee59a8d03c50f1b4709b7cc54d90bf41808934 which adds support for a `titleClasses` macro param and a `nhsuk-panel__title--l` modifier which drops the font size down from 48 (XL) to 36 (L).

The XL size seems ok for the green confirmation panel, which typically uses a short h1 like "Application complete", but for the blue interruption panel the text could be quite a bit longer - eg see [MOJ examples](https://design-patterns.service.justice.gov.uk/components/interruption-card/).

packages/nhsuk-frontend/src/nhsuk/components/panel/_index.scss
```diff
@@ -50,13 +50,22 @@ $nhsuk-border-width-panel: $nhsuk-panel-border-width;
...
+    background-color: nhsuk-colour("blue");
```
colinrotherham: Assuming the blue should always match the hero component?

We should use the brand colour variable if so:

```suggestion
    background-color: $nhsuk-brand-colour;
```

packages/nhsuk-frontend/src/nhsuk/components/panel/_index.scss
```diff
@@ -50,13 +50,22 @@ $nhsuk-border-width-panel: $nhsuk-panel-border-width;
...
+  }
+
   .nhsuk-panel__title {
     margin-top: 0;
 
     @include nhsuk-font-size(48);
     @include nhsuk-responsive-margin(5, "bottom");
   }
 
+  .nhsuk-panel__title--l {
```
colinrotherham, line 73: @anandamaryon1 Do you [still want the other sizes adding](https://github.com/nhsuk/nhsuk-frontend/pull/1196#discussion_r2010010655)?

anandamaryon1: I was thinking for completeness it makes sense to add them all, because seeing `--l` makes me assume there would be `--m` etc too. 

But, I can't think of a good use-case to have a smaller than `l` heading. 

Could be more intuitive to have both `--xl` and `--l`, rather than having `xl` as the default. And then leave it at those two, which align to our general page heading guidance of `xl` or `l`.

anandamaryon1:
We discussed this component at the Design System Review Panel. We agreed it seems to have many NHS-specific use-cases and we are continuing to gather examples (adding them to [this issue](https://github.com/nhsuk/nhsuk-service-manual-community-backlog/issues/241)) with a view to releasing this once we have finalised this PR and drafted guidance for the service manual.

Some points raised for consideration:
- Adding a ‘no’ option beside the main ‘ok, continue’ button. Eg. to give a clear no action if panel is used as a confirmation, instead of relying on a normal back button. 
- Styling: will one visual treatment work for all use cases or do we need different variants? i.e. is the presentation flexible enough?

colinrotherham:
Would be great to add as many examples to this PR as we can

anandamaryon1:
> Would be great to add as many examples to this PR as we can

Deffo!

I've asked some colleagues whether they're OK with me sharing their examples in GitHub. I was going to add them to the issue, since that'll stay open once this PR is closed.

packages/nhsuk-frontend/src/nhsuk/components/panel/_index.scss
frankieroberto: @paulrobertlloyd @anandamaryon1 renamed in 02416fb02481e7b4fb3f00664e97c780223d5136 to `--interruption`

packages/nhsuk-frontend/src/nhsuk/components/panel/_index.scss
frankieroberto: Hard to say! My [GP’s website](https://www.islingtoncentralmedicalcentre.org) uses purple as the brand colour, but I’m not sure whether it’d make sense to have the interruption panel inherit that or not? 🤷  I'd be inclined to think not - and then if someone wants to apply their brand colour they can override it?

packages/nhsuk-frontend/src/nhsuk/components/panel/_index.scss
frankieroberto: I’ve added `--xl` and `--m` in e7ac7a7824046e6c24464831399e8203681babfe (as well as `--l`).

I think we’d need to keep `--xl` as the default, as otherwise it’d be a breaking change? XL also feels like it makes sense for the majority case, which is the confirmation page?

packages/nhsuk-frontend/src/nhsuk/components/panel/_index.scss
colinrotherham: Secondary button, notification banner, hero and header all use `$nhsuk-brand-colour`

I agree with your reasoning but until we change (probably) 3 of those to `nhsuk-colour("blue")` I think it should follow the current precedent. What do you reckon?

i.e. You might be right but the decision needs making consistently

colinrotherham: Your GP's `$app-brand-colour` would be different to the NHS one

frankieroberto: ok, done in 3ecc3da1cb0f82cb5960d7ecec5faa48dff53bd1

frankieroberto:
@anandamaryon1 @colinrotherham @paulrobertlloyd think I’ve addressed all the comments!?

We can keep collecting examples in https://github.com/nhsuk/nhsuk-service-manual-community-backlog/issues/241

I've added a 'Cancel' link beside the button in the example, but we can update the wording if we have a better example.

In terms of flexibility, I’ve added support for the `titleClasses` macro option, and `--xl`, `--l` and `--m` options for the title. Is that enough for now do we think?

colinrotherham (approved):
Just merged in `main` so we can see the new examples page:
https://nhsuk-frontend-br-add-panel-va.herokuapp.com/nhsuk-frontend/components/panel/

Looking good! Approved ✅

Before merging, did we get any real service examples?

E.g. Like the one from MoJ

Image

frankieroberto:
@colinrotherham thanks for the ✅ .  @anandamaryon1 are you happy with this too?

Happy to update the example if we’ve got a better one. (Or that could be a follow-up PR).

Shall I make a start on the guidance too?

sarawilcox:
Could we put out a call for tested NHS examples? Or is there anything suitable here? https://github.com/nhsuk/nhsuk-service-manual-community-backlog/issues/241

frankieroberto:
@sarawilcox most of those examples are from the app and are perhaps a slightly different use case?

We could do something vaccinations related?

> (patient name) had a COVID-19 vaccine less than 3 months ago
> 
> They had a COVID-19 vaccine on (date).
> 
> For most people, the minimum recommended gap between COVID-19 vaccine doses is 3 months.
>
> [Continue anyway]

frankieroberto:
Or could use this example from @paulrobertlloyd: [#152](https://github.com/nhsuk/manage-vaccinations-in-schools-prototype/pull/152)

Screenshot of interruption card.

sarawilcox:
> @sarawilcox most of those examples are from the app and are perhaps a slightly different use case?
> 
> We could do something vaccinations related?
> 
> > (patient name) had a COVID-19 vaccine less than 3 months ago
> > They had a COVID-19 vaccine on (date).
> > For most people, the minimum recommended gap between COVID-19 vaccine doses is 3 months.
> > [Continue anyway]

Yes, good to have a vaccine/staff facing one. I think this is a good example. The one above (Paul's) is a bit complex for a coded example.

frankieroberto:
@sarawilcox done. The content in that example comes from our live service - written by @Anna-Sutton.

sarawilcox:
If you fancy having a go at drafting the guidance, @frankieroberto, that would be really helpful. I can then take a look at it.

colinrotherham:
Thank you for adding the example. Does the spacing below the link need another look?

image

frankieroberto:
@colinrotherham:

> Thank you for adding the example. Does the spacing below the link need another look?

That’s caused by the default margin below links in the button group. I've had a look at tweaking that separately in #1609

frankieroberto:
@anandamaryon1 @colinrotherham is this mergeable?

sarawilcox:
Hi Frankie, might be best if @colinrotherham @anandamaryon1 and I discuss this at our Monday PR catchup.

colinrotherham:
I've updated this PR to target [~`main`~](https://github.com/nhsuk/nhsuk-frontend/tree/main) [`support/10.x`](https://github.com/nhsuk/nhsuk-frontend/tree/support/10.x) ready for [v10.3.0](https://github.com/nhsuk/nhsuk-frontend/milestone/9)

packages/nhsuk-frontend/src/nhsuk/components/panel/template.njk
```diff
@@ -5,7 +5,7 @@
 <div {%- if params.id %} id="{{ params.id }}"{% endif %} class="nhsuk-panel
   {%- if params.classes %} {{ params.classes }}{% endif %}"
   {{- nhsukAttributes(params.attributes) }}>
-  <h{{ headingLevel }} class="nhsuk-panel__title">
+  <h{{ headingLevel }} class="nhsuk-panel__title {%- if params.titleClasses %} {{ params.titleClasses }}{% endif %}">
```
frankieroberto: @colinrotherham @anandamaryon1 @edwardhorsford @paulrobertlloyd following #1708 should we also add a `size` option here too?

Then we could do this:

```njk
{{ panel({
  titleText: "Confirm you want to cancel your hospital appointment",
  titleSize: "l",
  classes: "nhsuk-panel--interruption",
  html: panelHtml
}) }}
```

rather than this:

```njk
{{ panel({
  titleText: "Confirm you want to cancel your hospital appointment",
  titleClasses: "nhsuk-panel__title--l",
  classes: "nhsuk-panel--interruption",
  html: panelHtml
}) }}
```

Another option would be to move to a nested object, but also support `titleText` for backwards-compatibility:

```njk
{{ panel({
  title: {
    text: "Confirm you want to cancel your hospital appointment",
    size: "l"
  },
  classes: "nhsuk-panel--interruption",
  html: panelHtml
}) }}
```

anandamaryon1: Nested looks nicer to me. Could it / would it still support classes on the title?

frankieroberto: Yep I think either way we could support both size and classes, in case you wanted to add some other override classes?

colinrotherham: I had this same dilemma for `captionSize` versus `caption.size`

I chose `captionSize` (for now) because it aligns with `captionClasses` used by both NHS and GOV

I'd do the same here and vote for `titleSize` (for now) but add a v11 issue to fix them all

frankieroberto: Ok I've added `titleSize` for now. `titleClasses` will also work.

packages/nhsuk-frontend/src/nhsuk/components/panel/template.njk
colinrotherham, line 1: Would you mind pulling out all the variables so they're set at the top?

We've not done it to all components yet, but you'll see this elsewhere:

```njk

{#- Set classes for this component #}
{%- set classNames = "nhsuk-panel" -%}
{%- set classNamesTitle = "nhsuk-panel__title" -%}

{%- if params.titleSize and
  params.titleSize in ["s", "m", "l", "xl"] and (
    not params.titleClasses or (
      not "nhsuk-panel__title--s" in params.titleClasses and
      not "nhsuk-panel__title--m" in params.titleClasses and
      not "nhsuk-panel__title--l" in params.titleClasses and
      not "nhsuk-panel__title--xl" in params.titleClasses
    )
  )
%}
  {% set classNamesTitle = classNamesTitle ~ " nhsuk-panel__title--" ~ params.titleSize %}
{% endif %}

{%- if params.classes %}
  {% set classNames = classNames ~ " " ~ params.classes %}
{% endif %}

{%- if params.titleClasses %}
  {% set classNamesTitle = classNamesTitle ~ " " ~ params.titleClasses %}
{% endif %}

{%- set headingLevel = params.headingLevel if params.headingLevel else 1 -%}
```

frankieroberto: @colinrotherham done in 68d385ea39393c064b2314f17ff46743b9fae67a - would you mind checking?

packages/nhsuk-frontend/src/nhsuk/components/panel/macro-options.mjs
```diff
@@ -23,6 +23,17 @@ export const params = {
     description:
       'If `titleText` is set, this is not required. HTML to use within the panel. If `titleHtml` is provided, the `titleText` option will be ignored.'
   },
+  titleSize: {
+    type: 'string',
+    required: false,
+    description: 'Size of the title – `"s"`, `"m"`, `"l"` or `"xl"`.'
```
colinrotherham: Ah, there isn't an `--s` modifier is there?

packages/nhsuk-frontend/src/nhsuk/components/panel/template.njk
```diff
@@ -1,11 +1,33 @@
...
+      not "nhsuk-panel__title--s" in params.titleClasses and
+      not "nhsuk-panel__title--m" in params.titleClasses and
+      not "nhsuk-panel__title--l" in params.titleClasses and
+      not "nhsuk-panel__title--xl" in params.titleClasses
+    )
+  )
+%}
+  {% set classNamesTitle = classNamesTitle ~ " nhsuk-panel__title--" ~ params.titleSize %}
+{% endif %}
+{%- if params.titleClasses %}
```
colinrotherham, line 23: Missing a new line here 🙏

```suggestion

{%- if params.titleClasses %}
```

packages/nhsuk-frontend/src/nhsuk/components/panel/template.njk
```diff
@@ -1,11 +1,33 @@
 {% from "nhsuk/macros/attributes.njk" import nhsukAttributes %}
 
-{% set headingLevel = params.headingLevel if params.headingLevel else 1 -%}
+{%- set headingLevel = params.headingLevel if params.headingLevel else 1 -%}
 
-<div {%- if params.id %} id="{{ params.id }}"{% endif %} class="nhsuk-panel
-  {%- if params.classes %} {{ params.classes }}{% endif %}"
-  {{- nhsukAttributes(params.attributes) }}>
-  <h{{ headingLevel }} class="nhsuk-panel__title">
+{# Set classes for this component #}
```
colinrotherham:
```suggestion
{#- Set classes for this component #}
```

packages/nhsuk-frontend/src/nhsuk/components/panel/template.njk
```diff
@@ -1,11 +1,33 @@
...
+{%- set classNames = "nhsuk-panel" %}
```
colinrotherham: This one's not necessary but it'll help with find & replace if we match other components

```suggestion
{%- set classNames = "nhsuk-panel" -%}
```

packages/nhsuk-frontend/src/nhsuk/components/panel/template.njk
```diff
@@ -1,11 +1,33 @@
 {% from "nhsuk/macros/attributes.njk" import nhsukAttributes %}
 
-{% set headingLevel = params.headingLevel if params.headingLevel else 1 -%}
+{%- set headingLevel = params.headingLevel if params.headingLevel else 1 -%}
```
colinrotherham: If you can stick heading level last?

I needed the closing `-%}` to remove the empty space before the opening HTML tag

packages/nhsuk-frontend/src/nhsuk/components/panel/template.njk
colinrotherham: Resolved in https://github.com/nhsuk/nhsuk-frontend/pull/1196/commits/3a5dabb60f4d6cd83df5020f309a5a999ef35c77

packages/nhsuk-frontend/src/nhsuk/components/panel/template.njk
colinrotherham: Resolved in https://github.com/nhsuk/nhsuk-frontend/pull/1196/commits/3a5dabb60f4d6cd83df5020f309a5a999ef35c77

packages/nhsuk-frontend/src/nhsuk/components/panel/template.njk
colinrotherham: Resolved in https://github.com/nhsuk/nhsuk-frontend/pull/1196/commits/3a5dabb60f4d6cd83df5020f309a5a999ef35c77

colinrotherham:
I've brought the examples over from https://github.com/nhsuk/nhsuk-service-manual/pull/2316

* [**Panel component** example listing page](https://nhsuk-frontend-br-add-panel-va.herokuapp.com/nhsuk-frontend/components/panel/)

Can update them as necessary

packages/nhsuk-frontend/src/nhsuk/components/panel/template.njk
colinrotherham: Resolved in https://github.com/nhsuk/nhsuk-frontend/pull/1196/commits/3a5dabb60f4d6cd83df5020f309a5a999ef35c77

packages/nhsuk-frontend/src/nhsuk/components/panel/macro-options.mjs
colinrotherham: Resolved in https://github.com/nhsuk/nhsuk-frontend/pull/1196/commits/3a5dabb60f4d6cd83df5020f309a5a999ef35c77

colinrotherham (approved):

PR #1198, opened 2025-03-25
Change reverse button's shadow colour to match other buttons

anandamaryon1:
## Description
Updates the reverse button's shadow colour so that it visually matches better with other buttons. On primary buttons the shadow is an extension of the button, imagining that the button is 3d. Button's hover state also changed accordingly. Active state changed from black to a dark grey shade, to match primary and warning variants that use a very dark shade rather than black. 

| Before    | After |
| -------- | ------- |
| image  | image   |
| image    | image |
| image    | image |
| image    | image |

## Checklist

- [x] CHANGELOG entry

### DIFF
--- a/packages/components/button/_button.scss
+++ b/packages/components/button/_button.scss
@@ -140,6 +140,7 @@ $button-shadow-size: 4px;
 
   &:hover {
     background-color: $nhsuk-reverse-button-hover-color;
+    box-shadow: 0 $button-shadow-size 0 $nhsuk-reverse-button-shadow-hover-color;
     color: $nhsuk-reverse-button-text-color;
   }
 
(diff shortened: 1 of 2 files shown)


frankieroberto:
@anandamaryon1 nice! Could go further and switch the active state of the reversed button to be black text on grey background?

The way that the colours reverse on click feels a bit jarring to me, and is inconsistent with the other buttons, where the text colour remains the same and only the background colour changes.

anandamaryon1:
Agree @frankieroberto, I've tweaked the `active` state to align better with other buttons and make it less jarring on click. Still has over 7.5:1 text contrast and 3:1 background contrast (on NHS blue at least). Updated screens in opening comment.

frankieroberto (approved):
Looks good to me!

paulmsmith:
My initial personal reaction to this is. The new one looks less like a button to me.

frankieroberto:
The lighter shadow looks ok and a bit more natural to me, but no strong feelings either way. Could also explore a shade of grey between the 2 options?

@michaelgallagher any input on this from the NHS App perspective? You use a lot of reversed buttons...

michaelgallagher:
i'm also not convinced that it doesn't look as clickable, but that's just me. have we tried putting this in front of humans who don't work here?

frankieroberto (approved):

frankieroberto:
> Taking the shadow to ~20%~ 30% shade gives us this

Looks good to me. Feels like a nice balance between the current black shadow (“beard”) and the originally proposed light grey. 👍

PR #1199, opened 2025-03-25
[Draft / Proof of concept] Exploration of how we can improve our Nunjucks templates compatibility with Jinja2

MatMoore:
## Motivation

[Nunjucks templates are almost (but not quite) compatible with Jinja.](https://mozilla.github.io/nunjucks/faq.html#can-i-use-the-same-templates-between-nunjucks-and-jinja2-what-are-the-differences). As it stands, nhsuk-frontend nunjucks templates are not compatible enough for Jinja to render them as-is, but they are close enough to Jinja compatible that it is feasible that we could produce a version that renders in both template engines.

We are not proposing that NHS Frontend officially supports Jinja at this stage, but we are exploring whether we could get to the point where it is technically possible to render the templates in a Python app, without placing a javascript app in front of it.

There is prior art for porting design system components from Nunjucks to Jinja:
- [govuk-frontend-jinja](https://github.com/LandRegistry/govuk-frontend-jinja) which is a port of [govuk-frontend](https://github.com/alphagov/govuk-frontend)
- [ONS design system](https://service-manual.ons.gov.uk/design-system/guidance/templating-in-nunjucks#using-nunjucks-in-jinja2-templates) supports Jinja rendering with a small workaround ([example app](https://github.com/ONSdigital/srm-rh-ui))

## General approach

This branch shows how a relatively complex component (the checkboxes component) might look if we made it renderable by Jinja. Some changes are relatively minor, some are a little convoluted. The biggest problems we've run into are:
1. [Jinja's scoping behaviour](https://jinja.palletsprojects.com/en/stable/templates/#builtin-filters:~:text=call_something()%20%25%7D-,Scoping%20Behavior,-Please%20keep%20in) which prevents modifying a variable in a loop, and then using it outside of the loop.
2. Jinja is far stricter and prevents you accessing attributes on `undefined`.
3. The design system uses some parameter names that collide with built in objects in Python: `params.items` and `params.values`

Whether or not we move towards a separate package (like govuk-frontend-jinja), or cross-compatible templates (like ONS design system), the less incompatibilities we have in the codebase, the easier it will be to support Jinja *somewhere*.

However, increased compatibility with Jinja should not come at the cost of diverging too much from the upstream govuk-frontend templates, as this would increase the maintenance burden.

We are thinking that a first step might be to put together a style guide with patterns of what to avoid and what to do instead when making changes to nhsuk-frontend. This would be ideally supported by a linter tool.

## What's been ruled out
- Nunjucks has an [option to enable further jinja compatibility](https://mozilla.github.io/nunjucks/api.html#installjinjacompat), however this breaks a lot of templates that use `params.items`, since it completely overrides the variable lookup, regardless of which syntax you use, rather than following [Jinja's rules for variable lookup](https://jinja.palletsprojects.com/en/stable/templates/#variables). Switching this on would force us to change components' parameter names, so I've ruled it out for now.

## Options still to explore
- Creating custom filters or functions that we could package up with nhsuk-frontend, a Jinja port, or both, which would provide syntactic sugar that papers over differences between the rendering engines (e.g. the different scoping rules)
- Contributing to Jinja/Nunjucks itself

## Checklist

- [ ] CHANGELOG entry


packages/components/checkboxes/template.njk
```diff
@@ -12,39 +12,39 @@
...
-  {% if item.conditional %}
-    {% set isConditional = true %}
-  {% endif %}
-{% endfor %}
+{#- Jinja compatability - avoid referencing params.items/params.values directly
+  as it clashes with dict.items() and dict.values() -#}
+{% set items = params["items"] if "items" in params else [] %}
+{% set values = params["values"] if "values" in params else [] %}
+
+{% set isConditional = items | selectattr("conditional") | list | length > 0 %}
```
MatMoore, line 20: ❗️This expression doesn't generalise to the radios component, which is called with an object like this:

```json
"items": [
  {
      "value": "email",
      "text": "Email",
      "conditional": {
      "html": emailHtml
    }
  }
]
```

jinja_app/main.py
```diff
@@ -0,0 +1,48 @@
...
+
+app = Flask(__name__, static_url_path="/")
+
+
+TEMPLATE_PATH = (Path(__file__) / "../templates").resolve()
+
+# TODO: this script should copy assets over using glob paths, similar to docs.js
+
+
+class MultiRootEnvironment(Environment):
```
MatMoore, line 15: Not needed - use ChoiceLoader instead

packages/components/checkboxes/template.njk
```diff
@@ -65,19 +65,19 @@
...
       {{- nhsukAttributes(item.attributes) }}>
       {{ label({
-        html: item.html,
-        text: item.text,
-        classes: 'nhsuk-checkboxes__label' + (' ' + item.label.classes if item.label.classes),
-        attributes: item.label.attributes,
-        for: id
+        "html": item.html,
+        "text": item.text,
+        "classes": 'nhsuk-checkboxes__label' + (' ' + item.label.classes if item.label.classes else ''),
```
MatMoore, line 70: the else block might not be needed if the operator is changed from + to ~

packages/components/checkboxes/template.njk
```diff
@@ -12,39 +12,39 @@
    aria-describedby – for example hints or error messages -#}
 {% set describedBy = "" %}
 
-{% set isConditional = false %}
-{% for item in params.items %}
-  {% if item.conditional %}
-    {% set isConditional = true %}
-  {% endif %}
-{% endfor %}
+{#- Jinja compatability - avoid referencing params.items/params.values directly
```
MatMoore, line 15: I had another idea for dealing with the name collision here - we could transform it within the macro.njk, so that the templates themselves use a different name (e.g. `checkoxItems`) but the external API remains the same.

packages/nunjucks_extensions.mjs
```diff
@@ -0,0 +1,15 @@
+export const setupNunjucksExtensions = (env) => {
```
MatMoore, line 1: This is the change I'm least comfortable with, as it requires all users of the nunjucks templates enable custom filters.

It is only needed for the conditional check in the radios component.

Alternatively, we could choose to fork this one component only (so there is a radios macro just for Jinja)

PR #1216, opened 2025-04-04
Upgrade to Node.js 22

colinrotherham:
This PR updates Node.js to meet our guidance regarding LTS versions

E.g. See requirements in both [**Running locally**](https://github.com/nhsuk/nhsuk-frontend/blob/main/docs/contributing/running-locally.md#requirements) and [**Installing using npm**](https://github.com/nhsuk/nhsuk-frontend/blob/main/docs/installation/installing-with-npm.md#requirements) where we state:

>We recommend to use the [long-term support (LTS) version of Nodejs](https://nodejs.org/en/download/), rather than the latest beta version.

See https://github.com/nhsuk/nhsuk-frontend/pull/1216#discussion_r2028520259 below where Node.js 20 compatibility is maintained

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "9.3.0",
   "description": "NHS.UK frontend contains the code you need to start building user interfaces for NHS websites and services.",
   "engines": {
-    "node": "^22.11.0"
+    "node": "^20.9.0 || ^22.11.0"
   },
   "scripts": {
     "postinstall": "playwright install chromium --with-deps --only-shell",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -3,7 +3,7 @@
   "version": "9.3.0",
   "description": "NHS.UK frontend contains the code you need to start building user interfaces for NHS websites and services.",
   "engines": {
-    "node": ">=20.0.0"
+    "node": "^22.11.0"
```
frankieroberto, line 6: Should we continue to maintain compatibility with Node 20, or are there any features of Node 22 that would mean we’d want everyone to upgrade?

(This will affect the prototype kit too, as well as production services)

anandamaryon1 (approved):

PR #1217, opened 2025-04-04
Adopt GOV.UK Frontend packages structure

colinrotherham:
Closes https://github.com/nhsuk/nhsuk-frontend/issues/1351

### DIFF
--- a/gulpfile.mjs
+++ b/gulpfile.mjs
@@ -1,8 +1,6 @@
 import { join } from 'path'
 
-import browserSync from 'browser-sync'
-import gulp from 'gulp'
-import * as config from 'nhsuk-frontend-config'
+import * as config from '@nhsuk/frontend-config'
 import {
   buildHTML,
   validateHTML,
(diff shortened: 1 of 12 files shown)


colinrotherham:
>Hopefully this will make it easier to port stuff across.

Up for discussion of course, but yeah makes it much easier to bring the ports into this repo if desired

colinrotherham:
I've also followed https://github.com/alphagov/govuk-frontend/pull/4091 and used [scope `@nhsuk`](https://api.npms.io/v2/search?q=scope:nhsuk) for internal packages

**Before**

```console
nhsuk-frontend-config
nhsuk-frontend-helpers
nhsuk-frontend-tasks
```

**After**

```console
@nhsuk/frontend-config
@nhsuk/frontend-helpers
@nhsuk/frontend-tasks
```

This avoids potential [**"typosquatting" / dependency confusion**](https://docs.npmjs.com/threats-and-mitigations/#by-typosquatting--dependency-confusion) ([a bug existed once](https://github.com/npm/cli/issues/3637))

>A variant of this attack is when a public package is registered with the same name of a private package that an organization is using. We strongly encourage using [scoped packages](https://github.blog/2021-02-12-avoiding-npm-substitution-attacks/) to ensure that a private package isn’t being substituted with one from the public registry.

PR #1191, opened 2025-03-14
Switch to simpler review app styling etc

colinrotherham:
## Description

This PR includes:

1. Switch to simpler review app styling
2. Removes "Install" and "About" guidance pages
3. Deploys to GitHub pages automatically when merging into `main`

Point 3) means we can preview the latest [`nhsuk-frontend` app](https://nhsuk.github.io/nhsuk-frontend/) between releases

<br>

Simple review app

## Checklist

- [x] Tested against our [testing policy](https://github.com/nhsuk/nhsuk-frontend/blob/master/docs/contributing/testing.md) (Resolution, Browser & Accessibility)
- [x] Follows our [coding standards and style guide](https://github.com/nhsuk/nhsuk-frontend/blob/master/docs/contributing/coding-standards.md)
- [ ] CHANGELOG entry

PR #1218, opened 2025-04-07
Use GitHub Pages actions for deploying the development app

colinrotherham:
## Description

This PR was split out from https://github.com/nhsuk/nhsuk-frontend/pull/1191#discussion_r2030927771 and switches to using the official GitHub Pages actions


frankieroberto:
@colinrotherham there’s some example code here for doing NPM releases from GitHub workflows: https://docs.github.com/en/actions/use-cases-and-examples/publishing-packages/publishing-nodejs-packages

Looks like rather than using an action it just defines a step which runs this:

```
run: npm publish --provenance --access public
        env:
          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
```

Would that work?

Issue #1226, opened 2025-04-10
Header navigation items hidden after resize


colinrotherham:
# Bug Report

## What is the issue?

The header can calculate navigation item placement incorrectly because:

* Width on first load overflows past 100%
* Width on first load excludes "More" menu button
* Width includes the optionally hidden "Home" link
* Gap between items is calculated for the last item (in JS, not CSS)

## What steps are required to reproduce the issue?

1. Reduce the viewport size until an item is moved to the "More" menu
2. Restore the viewport size and original placement is not restored

This issue causes flaky tests in the following example:
https://nhsuk.github.io/nhsuk-frontend/components/header/header-service-name-with-nav.html

### Before

Width calculations (first load) allows space for the "Accessibility" link

<img src="https://github.com/user-attachments/assets/a61c0f6c-4778-4054-8d17-199303119b4c" alt="" width="80%">

### After

Width calculations (resizing smaller, then larger) incorrectly prevents space for the "Accessibility" link

<img src="https://github.com/user-attachments/assets/eb4ffb48-3427-40ba-9066-7378f0c42451" alt="" width="80%">

<br>
<br>

Similarly, starting at full width (resizing smaller) demonstrates how the "More" menu button overflows because it wasn't included in the width calculations on first load because all items had enough space:

![Image](https://github.com/user-attachments/assets/700da037-44a5-4536-861d-a9a4f6e90b91)

PR #1228, opened 2025-04-11
Fix header navigation item allocation on resize

colinrotherham:
## Description

This PR fixes https://github.com/nhsuk/nhsuk-frontend/issues/1226 and adds 2x commits from https://github.com/nhsuk/nhsuk-frontend/pull/1058 as follows:

1. Wrap header navigation items when JavaScript is not enabled
2. Add space between header navigation items using padding

Most helpful was 2) where CSS `gap` is removed

## Checklist

- [x] CHANGELOG entry


packages/components/header/_header.scss
```diff
@@ -460,8 +460,9 @@
   @extend %nhsuk-navigation-list;
   @include govuk-width-container;
   display: flex;
-  gap: nhsuk-spacing(4);
-  width: 100%;
+  flex-wrap: wrap;
+  column-gap: nhsuk-spacing(4);
+  width: calc(100% - $nhsuk-gutter);
```
MatMoore: What's the reason for changing the width here?

packages/components/header/header.js
```diff
@@ -211,8 +211,14 @@ class Header {
...
     }
 
-    if (document.body.offsetWidth !== this.width && this.menuIsOpen) {
-      this.closeMobileMenu()
+    // Check for width changes
+    if (this.width !== document.body.offsetWidth) {
+      this.width = document.body.offsetWidth
+
+      // Automatically close menu
+      if (this.menuIsOpen) {
```
MatMoore, line 244: Not related to your change but I'm wondering whether the auto-closing behaviour actually makes sense here? If you have the menu open and you resize the window, wouldn't you expect it to stay open, but update its contents?

packages/components/header/header.js
```diff
@@ -67,11 +68,24 @@ class Header {
...
    */
   calculateBreakpoints() {
-    let childrenWidth = 0
-    for (let i = 0; i < this.navigationList.children.length; i++) {
-      childrenWidth += this.navigationList.children[i].offsetWidth
-      this.breakpoints[i] = childrenWidth
+    let right = 0
+
+    // Add navigation items on first load only
+    if (!this.navigationItems.length) {
```
MatMoore, line 74: I'm not sure, but perhaps this block can be moved to its own method that gets called directly from the constructor, as it looks like it's just initialising the `navigationItems` array.

packages/components/header/header.js
```diff
@@ -224,7 +224,12 @@ class Header {
 
     // Move items based on available width
     items.forEach((item) => {
-      if (item.right <= this.navigationList.offsetWidth) {
+      const maxRight =
```
MatMoore, line 227: I think this would be a bit clearer if you handled the menu bar width on the other side of the comparison, e.g. outside the scope of the forEach you can have

```js
let availableSpace = this.navigationList.offsetWidth
if (items.length !== itemsList.length) {
  // subtract the space taken up by the mobile menu
  availableSpace -= this.mobileMenuContainer.offsetWidth;
}
```

then the comparison in the loop looks like `item.right <= availableSpace` again.

tests/backstop/engine_scripts/playwright/onReadyResize.js
```diff
@@ -0,0 +1,6 @@
+module.exports = async (page, scenario) => {
+  await require('./onReady')(page, scenario);
+
+  await page.evaluate(() => window.dispatchEvent(new Event('resize')));
+  await page.waitForTimeout(300);
```
MatMoore, line 5: does this actually need to be increased to 300ms or can we reduce this? It looks like the resize debounce is 100ms and there is a 200ms delay in responding to page orientation. Since it's used in a few tests it would be a shame to add to the runtime.

paulrobertlloyd:
This all looks good to me, some nice tidy up within the JS for this component too.

Take it this PR also ticks off updating the mocked navigation HTML used in the component test to actually represent the markup now being used as well?

MatMoore (commented):
I skipped over some of the CSS details, but the javascript changes look good to me. Extracting methods for showing/hiding the menu bar is a nice improvement 👍🏻 

Have a couple of minor suggestions inline, but the approach seems good.

paulrobertlloyd (approved):

packages/components/header/_header.scss
```diff
@@ -458,21 +462,20 @@
...
   display: flex;
-  gap: nhsuk-spacing(4);
-  width: 100%;
-
-  @include mq($from: tablet) {
-    border-top: 1px solid $color_transparent_nhsuk-white-20;
-  }
+  flex-wrap: wrap;
+  width: calc(100% + $nhsuk-gutter);
+  margin: 0 $nhsuk-gutter-half * -1;
```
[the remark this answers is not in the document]
paulrobertlloyd: Oh bugger, I was comparing this branch with `main`, not the change in `header-breaking-changes`! Either way, 🚢!

MatMoore (dismissed):
👍🏻

anandamaryon1:
Not sure whether it's really an issue, but in testing this I observe that the number of items shown varies between first load and any subsequent resizing at 768px wide. 

For example, if I resize from wider down to 768, I see the More button:
image


But if I refresh / load up on 768 wide, I see all the nav items at once, without the More button:
image

Not really a usability issue but maybe a symptom of some breakpoints not quite aligning?

anandamaryon1 (dismissed):

packages/components/header/header.js
```diff
@@ -35,11 +36,10 @@ class Header {
 
     this.setupMobileMenu()
     this.setupNavigation()
-    this.calculateBreakpoints()
     this.updateNavigation()
 
     this.handleResize = this.debounce(() => {
-      this.calculateBreakpoints()
+      this.resetNavigation()
```
MatMoore, line 42: minor thing so feel free to ignore, but the naming here confused me for a bit as resetNavigation and updateNavigation sound similar, and also `resetNavigation` has two side effects now (updating `this.navigationItems`, and moving the elements in the DOM)

I guess an alternate approach would be to leave this as `calculateBreakpoints`, remove the short circuiting behaviour, then have `updateNavigation` handle all the DOM manipulation.

I'm thinking in the "Move items based on available width" loop you would first moving everything back to the navigationList, then move the overflowing items to the mobile menu.

packages/components/header/README.md
```diff
@@ -50,7 +50,7 @@ compiled JavaScript for all components `nhsuk.min.js` or the individual componen
               <svg class="nhsuk-icon nhsuk-icon__chevron-right" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" aria-hidden="true" height="34" width="34"><path d="M15.5 12a1 1 0 0 1-.29.71l-5 5a1 1 0 0 1-1.42-1.42l4.3-4.29-4.3-4.29a1 1 0 0 1 1.42-1.42l5 5a1 1 0 0 1 .29.71z"></path></svg>
             </a>
           </li>
-          <li class="nhsuk-mobile-menu-container js-show">
+          <li class="nhsuk-mobile-menu-container">
```
MatMoore, line 53: It seems like `js-show` is still referenced here https://github.com/nhsuk/nhsuk-frontend/blob/header-calculation/packages/components/header/_header.scss#L132

tests/integration/jsdom/header.test.mjs
```diff
@@ -1,49 +1,65 @@
...
+
 import Header from '../../../packages/components/header/header.js'
 
+const user = userEvent.setup()
+
 describe('Header class', () => {
+  /** @type {HTMLElement} */
+  let navigation
+
+  /** @type {HTMLElement} */
```
MatMoore, line 14: I haven't seen these @type annotations before, what syntax is this?

tests/integration/jsdom/header.test.mjs
```diff
@@ -59,212 +75,209 @@ describe('Header class', () => {
...
+    navigation = getByRole(container, 'navigation')
+    menuButton = getByRole(container, 'button', { name: 'Browse More' })
+
+    listWidth = 800
+    itemWidth = 100
+
+    jest
+      .spyOn(HTMLElement.prototype, 'offsetWidth', 'get')
+      .mockImplementation(function () {
+        return this instanceof HTMLUListElement
```
MatMoore, line 90: [minor] Is there a way we could make this mock more targeted, like only applying it applying to a particular class or test id?

tests/integration/jsdom/header.test.mjs
```diff
@@ -59,212 +75,209 @@ describe('Header class', () => {
...
+        const listItems = navigation.querySelectorAll('nav > ul > li')
+        const menuItems = navigation.querySelectorAll('nav > ul > li li')
+
+        expect(listItems).toHaveLength(expected.listItems)
+        expect(menuItems).toHaveLength(expected.menuItems)
+      }
+    )
 
-    navigationOffsetWidthSpy.mockRestore()
+    it.each(examples)(
```
MatMoore, line 262: Nice work on these tests 👍🏻

MatMoore (dismissed):

MatMoore (approved):

packages/components/header/header.js
```diff
@@ -193,42 +190,32 @@ class Header {
...
+    this.width -= this.mobileMenuContainer.offsetWidth
 
-      // Subtract space for menu button
-      if (items.length !== itemsList.length) {
-        maxRight -= this.mobileMenuContainer.offsetWidth
-      }
-
-      if (item.right > maxRight) {
+    // Move items based on available width
+    this.navigationItems.forEach((item) => {
```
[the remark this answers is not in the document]
MatMoore: *Edit* no it can't, the calculation needs to be redone to take into account the menu button width

PR #1230, opened 2025-04-15
Add syntax highlighting to fenced code blocks in readmes

MatMoore:
## Description

I noticed a lot of code blocks in the documentation don't have syntax highlighting enabled, so have gone through and added in the language where missing. This is not a user-facing change; it just improves the experience of reading documentation on github.

As a side effect, I noticed css examples in the linting guide themselves now get picked up by stylelint, so I've configured it to be a lot more liberal when processing this file.

### scss
Screenshot 2025-04-15 at 15 30 14

### nunjucks
Screenshot 2025-04-15 at 15 30 37

### shell
(this one doesn't do much)
Screenshot 2025-04-15 at 15 31 00

## Checklist

- [ ] CHANGELOG entry (not sure if this is relevant enough for an entry - if so please let me know how to format it)

### DIFF
--- a/.stylelintrc.js
+++ b/.stylelintrc.js
@@ -5,12 +5,39 @@ module.exports = {
     '**/vendor/**',
 
     // Ignore CSS-in-JS (including dotfiles)
-    '**/?(.)*.{cjs,js,mjs}'
+    '**/?(.)*.{cjs,js,mjs}',
+
+    // Prevent CHANGELOG history changes
+    'CHANGELOG.md'
   ],
   overrides: [
     {
       customSyntax: 'postcss-markdown',
       files: ['**/*.md']
+    },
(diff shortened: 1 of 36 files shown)


edwardhorsford:
Would it be possible to add some screenshots to the pr for each of the languages? good as a reference of before vs after, etc.

frankieroberto:
Nice! I think the plan is to remove most of those Nunjucks and HTML examples from the README files for each component, as they’re hard to keep up-to-date (some may even be out-of-date already), and largely duplicate the examples on https://service-manual.nhs.uk/design-system/components - and instead replace with a minimal set of text and a link to the relevant page on the design system website.

However not sure when that's planned for? @anandamaryon1 @colinrotherham

PR #1272, opened 2025-04-30
Remove component `init()` functions

colinrotherham:
## Description

This PR removes component `init()` functions as part of https://github.com/nhsuk/nhsuk-frontend/issues/1235

See related GOV.UK Frontend decision:

* [**Remove init() method from JavaScript components**](https://github.com/alphagov/govuk-design-system-architecture/blob/main/decision-records/010-remove-init-method.md)

JavaScript components are not currently exported so this change is non-breaking

## Checklist

- [ ] CHANGELOG entry

### DIFF
--- a/packages/components/button/button.js
+++ b/packages/components/button/button.js
@@ -1,10 +1,21 @@
+const KEY_SPACE = 32
+const DEBOUNCE_TIMEOUT_IN_SECONDS = 1
+
 class Button {
   constructor($module) {
-    this.KEY_SPACE = 32
-    this.DEBOUNCE_TIMEOUT_IN_SECONDS = 1
+    if (!$module) {
+      return this
+    }
 
     this.$module = $module
     this.debounceFormSubmitTimer = null
+
+    /**
+     * Initialise an event listener for keydown at document level
+     * this will help listening for later inserted elements with a role="button"
+     */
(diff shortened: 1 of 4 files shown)


packages/components/button/button.js
```diff
@@ -1,10 +1,21 @@
+const KEY_SPACE = 32
+const DEBOUNCE_TIMEOUT_IN_SECONDS = 1
+
 class Button {
   constructor($module) {
-    this.KEY_SPACE = 32
-    this.DEBOUNCE_TIMEOUT_IN_SECONDS = 1
+    if (!$module) {
```
MatMoore, line 6: I find it a bit confusing that the element is called $module here... is it worth renaming it? I checked govuk-frontend and those components call this argument `$root` instead.

packages/components/button/button.js
```diff
@@ -1,10 +1,21 @@
...
+      return this
```
MatMoore, line 7: Not a blocker, but this behaviour of returning a partially instantiated object if called with invalid arguments or dom structure seems a bit strange to me. Is it something you're planning to revisit?

I would expect it to throw an error so it can be handled if necessary, similar to the type checking here https://github.com/alphagov/govuk-frontend/blob/main/packages/govuk-frontend/src/govuk/component.mjs#L60

MatMoore (approved):

PR #1273, opened 2025-04-30
Increase component test coverage

colinrotherham:
## Description

Formats and improves JSDOM tests for components and automatic initialisation

### Before

Test coverage before

```console
-------------------------------------|---------|----------|---------|---------
File                                 | % Stmts | % Branch | % Funcs | % Lines
-------------------------------------|---------|----------|---------|---------
All files                            |   34.74 |    30.45 |   38.39 |   34.82
-------------------------------------|---------|----------|---------|---------

Test Suites: 11 passed, 11 total
Tests:       106 passed, 106 total
```

### After

Test coverage after

```console
-------------------------------------|---------|----------|---------|---------
File                                 | % Stmts | % Branch | % Funcs | % Lines
-------------------------------------|---------|----------|---------|---------
All files                            |   82.09 |    82.03 |   84.14 |   82.09
-------------------------------------|---------|----------|---------|---------

Test Suites: 15 passed, 15 total
Tests:       172 passed, 172 total
```

## Checklist

- [ ] CHANGELOG entry

### DIFF
--- a/packages/components/button/button.jsdom.test.mjs
+++ b/packages/components/button/button.jsdom.test.mjs
@@ -4,7 +4,7 @@ import initButton from './button.js'
 
 describe('Button', () => {
   /** @type {HTMLButtonElement} */
-  let button
+  let $button
 
   beforeEach(() => {
     document.body.innerHTML = `
@@ -13,22 +13,22 @@ describe('Button', () => {
       </button>
     `
 
-    button = getByRole(document.body, 'button')
+    $button = getByRole(document.body, 'button')
 
-    jest.spyOn(button, 'addEventListener')
+    jest.spyOn($button, 'addEventListener')
   })
 
   describe('Initialisation', () => {
     it('should add event listeners', () => {
       initButton()
 
-      expect(button.addEventListener).toHaveBeenNthCalledWith(
+      expect($button.addEventListener).toHaveBeenNthCalledWith(
         1,
         'keydown',
         expect.any(Function)
       )
 
-      expect(button.addEventListener).toHaveBeenNthCalledWith(
+      expect($button.addEventListener).toHaveBeenNthCalledWith(
         2,
         'click',
         expect.any(Function)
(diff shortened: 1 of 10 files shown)


packages/components/button/button.jsdom.test.mjs
```diff
@@ -1,11 +1,59 @@
...
+    `
+
+    button = getByRole(document.body, 'button')
+
+    jest.spyOn(button, 'addEventListener')
+  })
+
+  describe('Exports', () => {
+    it('should export init function', () => {
+      expect(initButton).toBeInstanceOf(Function)
```
MatMoore, line 23: this kind of test seems a bit unnecessary - if I was to export something else instead I would expect the tests below to fail anyway.

packages/components/details/details.jsdom.test.mjs
```diff
@@ -0,0 +1,166 @@
...
+        expect(() => initDetails({ scope })).not.toThrow()
+      })
+    })
+
+    describe('Accessibility', () => {
+      beforeEach(() => {
+        initDetails()
+      })
+
+      it('should add to summary the button role', async () => {
```
MatMoore, line 128: _extremely_ minor but do you think a slightly different sentence structure would read better here, e.g.

- 'should set the summary's role to "button"'
- 'should set the summary's aria-expanded to false'

MatMoore (commented):
Looks good to me.

I haven't checked whether the general conventions match up with govuk-frontend as I'm not familiar with that codebase, but I think the closer we can make it the easier it will be to bring in new components. Overall though the changes make sense, and more coverage is always good 👍🏻 

Regarding the polyfills, given there's only a couple of them, it doesn't seem like it would be worth the effort of switching out JSDOM just for this, so it seems fine to keep them, unless you had other reasons for moving to Happy DOM.

If you're concerned about the mocked out behaviour being unexpected, another option might be to inline it into the tests that actually need it rather than doing it globally.

MatMoore (approved):

Issue #1240, opened 2025-04-17
Discuss strategy for adopting GOV.UK grid layout


Smartin684:

Smartin684:
Waiting on the next design system review panel

Smartin684:
Scheduled for front end release 11 (breaking change)

PR #1296, opened 2025-05-13
Apply grid classes from tablet (not desktop)

colinrotherham:
## Description

We've updated our grid column styles to be applied at **tablet** width (641px and up)

Previously they were applied from **desktop** width (769px and up) making it difficult to cater for smaller screen sizes

* [**Grid layout** example page](https://nhsuk-frontend-pr-1296.herokuapp.com/nhsuk-frontend/examples/grid/)

### Grid column classes

In addition to the breakpoint change, new "from mobile" and "from desktop" flavours are included:

```console
nhsuk-grid-column-[grid-size]-from-mobile
nhsuk-grid-column-[grid-size]-from-desktop
```

For example, you can make a column one-half on tablet but expand to two-thirds on desktop sized screens:

```patch
- <div class="nhsuk-grid-column-one-half">
+ <div class="nhsuk-grid-column-two-thirds nhsuk-grid-column-one-half-from-desktop">
```

Part of https://github.com/nhsuk/nhsuk-frontend/issues/1240

---

### Mobile width override classes

Grid column classes ending `-from-mobile` are applied at mobile width (320px) and above.

If you're using the following mobile width utility classes, you must:

- Replace `nhsuk-u-one-half` with `nhsuk-grid-column-one-half-from-mobile`
- Replace `nhsuk-u-one-third` with `nhsuk-grid-column-one-third-from-mobile`
- Replace `nhsuk-u-two-thirds` with `nhsuk-grid-column-two-thirds-from-mobile`
- Replace `nhsuk-u-one-quarter` with `nhsuk-grid-column-one-quarter-from-mobile`
- Replace `nhsuk-u-three-quarters` with `nhsuk-grid-column-three-quarters-from-mobile`

### Tablet width override classes

Grid column classes not ending `-from-mobile` or `-from-desktop` are applied at tablet width (641px) and above.

If you're using the following tablet width utility classes, you must:

- Replace `nhsuk-u-one-half-tablet` with `nhsuk-grid-column-one-half`
- Replace `nhsuk-u-one-third-tablet` with `nhsuk-grid-column-one-third`
- Replace `nhsuk-u-two-thirds-tablet` with `nhsuk-grid-column-two-thirds`
- Replace `nhsuk-u-one-quarter-tablet` with `nhsuk-grid-column-one-quarter`
- Replace `nhsuk-u-three-quarters-tablet` with `nhsuk-grid-column-three-quarters`

## Checklist

- [x] CHANGELOG entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,54 @@
 # NHS.UK frontend Changelog
 
+## Unreleased
+
+### :boom: **Breaking changes**
+
+#### Apply grid column widths from tablet (not desktop) width
+
+We've updated our grid column styles to be applied at tablet width (641px and up).
+
+Previously they were applied from desktop width (769px and up) making it difficult to cater for smaller screen sizes.
+
+Please carefully review your pages. If necessary, different grid behaviour for the mobile and desktop breakpoints can be applied using new classes ending `-from-mobile` and `-from-desktop`.
+
+For example, you can make a column three-quarters on tablet but reduce to two-thirds on desktop sized screens:
+
+```patch
+  <div class="nhsuk-grid-row">
+-   <div class="nhsuk-grid-column-two-thirds">
++   <div class="nhsuk-grid-column-three-quarters nhsuk-grid-column-two-thirds-from-desktop">
+      <!-- Component -->
+    </div>
+  </div>
+```
+
+##### Mobile width override classes
+
+Grid column classes ending `-from-mobile` are applied at mobile width (320px) and above.
+
+If you're using the following mobile width utility classes, you must:
+
+- Replace `nhsuk-u-one-half` with `nhsuk-grid-column-one-half-from-mobile`
+- Replace `nhsuk-u-one-third` with `nhsuk-grid-column-one-third-from-mobile`
+- Replace `nhsuk-u-two-thirds` with `nhsuk-grid-column-two-thirds-from-mobile`
+- Replace `nhsuk-u-one-quarter` with `nhsuk-grid-column-one-quarter-from-mobile`
+- Replace `nhsuk-u-three-quarters` with `nhsuk-grid-column-three-quarters-from-mobile`.
+
+##### Tablet width override classes
+
+Grid column classes not ending `-from-mobile` or `-from-desktop` are applied at tablet width (641px) and above.
+
+If you're using the following tablet width utility classes, you must:
+
+- Replace `nhsuk-u-one-half-tablet` with `nhsuk-grid-column-one-half`
+- Replace `nhsuk-u-one-third-tablet` with `nhsuk-grid-column-one-third`
+- Replace `nhsuk-u-two-thirds-tablet` with `nhsuk-grid-column-two-thirds`
+- Replace `nhsuk-u-one-quarter-tablet` with `nhsuk-grid-column-one-quarter`
+- Replace `nhsuk-u-three-quarters-tablet` with `nhsuk-grid-column-three-quarters`
+
+This change was introduced in [pull request #1296: Apply grid classes from tablet (not desktop)](https://github.com/nhsuk/nhsuk-frontend/pull/1296).
+
 ## 10.2.2 - 4 December 2025
 
 Note: This release was created from the `support/10.x` branch.


anandamaryon1:
Reviewed at the Design System Review Panel and signed-off. 

Just one to-do for me over on service manual: https://github.com/nhsuk/nhsuk-service-manual/issues/2268.

anandamaryon1 (approved):

colinrotherham:
Thanks! For v11 yeah?

We could do with moving all the v10 releases to a support branch to free up `main`

colinrotherham:
@anandamaryon1 Can you check this again before we merge into v11?

Our design system [Styles: **Layout** guidance](https://service-manual.nhs.uk/design-system/styles/layout) also has these sections:

* **Tablet and mobile specific grid classes**
* **Tablet specific grid classes**

Unlike GOV.UK Frontend we (confusingly) have two sets of width overrides.

I'll need your help in deciding what to do:

## Tablet and mobile specific grid classes

As mentioned above, we confusingly use both:

* ✅ `nhsuk-u-width-[grid-size]`
* 🤔 `nhsuk-u-[grid-size]`

To solve this, I've proposed a new "always applied" column style matching the grid class naming convention:

- Rename `nhsuk-u-one-half` class to `nhsuk-grid-column-one-half-always`
- Rename `nhsuk-u-one-third` class to `nhsuk-grid-column-one-third-always`
- Rename `nhsuk-u-two-thirds` class to `nhsuk-grid-column-two-thirds-always`
- Rename `nhsuk-u-one-quarter` class to `nhsuk-grid-column-one-quarter-always`
- Rename `nhsuk-u-three-quarters` class to `nhsuk-grid-column-three-quarters-always`

Their purpose is now clear and avoids overlap with [`nhsuk-u-width-[grid-size]` width override classes](https://service-manual.nhs.uk/design-system/styles/layout#width-override-classes)

**Question:** Should we align with GOV.UK Frontend and _remove_ the "always applied" width classes?

You can see them all in use on the [grid layout example page](https://nhsuk-frontend-pr-1296.herokuapp.com/nhsuk-frontend/examples/grid/)

## Tablet specific grid classes

As mentioned above, this PR moves to "from ~tablet~ desktop" classes (not modifiers)

This means the following tablet override grid classes have been removed and must be renamed:

- Rename `nhsuk-u-one-half-tablet` class to `nhsuk-grid-column-one-half`
- Rename `nhsuk-u-one-third-tablet` class to `nhsuk-grid-column-one-third`
- Rename `nhsuk-u-two-thirds-tablet` class to `nhsuk-grid-column-two-thirds`
- Rename `nhsuk-u-one-quarter-tablet` class to `nhsuk-grid-column-one-quarter`
- Rename `nhsuk-u-three-quarters-tablet` class to `nhsuk-grid-column-three-quarters`

```patch
  <div class="nhsuk-grid-row">
-   <div class="nhsuk-grid-column-two-thirds nhsuk-u-one-half-tablet">
+   <div class="nhsuk-grid-column-one-half nhsuk-grid-column-two-thirds-from-desktop">
      <!-- Component -->
    </div>
  </div>
```

I'm happy with this

You can see them all in use on the [grid layout example page](https://nhsuk-frontend-pr-1296.herokuapp.com/nhsuk-frontend/examples/grid/)

## Mobile only grid classes

I appreciate it's hard to people to flip their thinking from desktop to tablet

To help with this I've also included "until tablet" classes:

- Added `nhsuk-grid-column-one-half-until-tablet`
- Added `nhsuk-grid-column-one-third-until-tablet`
- Added `nhsuk-grid-column-two-thirds-until-tablet`
- Added `nhsuk-grid-column-one-quarter-until-tablet`
- Added `nhsuk-grid-column-three-quarters-until-tablet`

**Question:** Should we align with GOV.UK Frontend and _remove_ the "until tablet" width classes?

You can see them all in use on the [grid layout example page](https://nhsuk-frontend-pr-1296.herokuapp.com/nhsuk-frontend/examples/grid/)

frankieroberto:
@colinrotherham re:

> Rename `nhsuk-u-one-half` class to `nhsuk-grid-column-one-half-always`

Does this imply that `nhsuk-grid-column-one-half-always` should always be included within a `nhsuk-grid-row`? If so, this makes sense to me!

And then `nhsuk-u-width-one-half` is for components like inputs or anything else?

I must admit I’ve always found the `from` and `until` classes a bit confusing, I'm not sure if it’s just me! I think it’s because they only make sense when you assume a direction of small-to-big.

Would something like this be clearer maybe?

* `nhsuk-grid-column-one-half-desktop-only` (instead of `-from-desktop`)
* `nhsuk-grid-column-one-half-mobile-only` (instead of `-until-tablet`)

colinrotherham:
> > Rename `nhsuk-u-one-half` class to `nhsuk-grid-column-one-half-always`
> 
> Does this imply that `nhsuk-grid-column-one-half-always` should always be included within a `nhsuk-grid-row`? If so, this makes sense to me!

It means the grid width is "always applied" without any breakpoints

I'm open to name suggestions instead of ~`nhsuk-u-one-half`~ `nhsuk-grid-column-one-half-always`

Grid column one-half-always

<br>

> And then `nhsuk-u-width-one-half` is for components like inputs or anything else?

Yeah we only need one set of non-grid [width override classes](https://service-manual.nhs.uk/design-system/styles/layout#width-override-classes)

<br>

> I must admit I’ve always found the `from` and `until` classes a bit confusing, I'm not sure if it’s just me! I think it’s because they only make sense when you assume a direction of small-to-big.

Ah sorry, I've assumed as part of https://github.com/nhsuk/nhsuk-frontend/issues/1240 we're happy with [mobile first](https://developer.mozilla.org/en-US/docs/Glossary/Mobile_First) and it matches Sass `$from` and `$until`

<br>

> Would something like this be clearer maybe?
> 
> * `nhsuk-grid-column-one-half-desktop-only` (instead of `-from-desktop`)
> * `nhsuk-grid-column-one-half-mobile-only` (instead of `-until-tablet`)

Yeah it's clearer, although "mobile" only starts from 320px and up—what about below that?

Don't forget the GOV.UK Frontend grid classes are cumulative whereas "only" implies a min- and max-width

---

Some history btw

GOV.UK Frontend accidently left the `nhsuk-grid-column()` Sass variable as `$at` instead of `$from`

From the [PR description](https://github.com/alphagov/govuk-frontend/pull/1094#issue-391045759) they mention "an `at-tablet` class" too:

1. `-one-half` "from tablet" by default
2. `-one-half-at-tablet` "from tablet until desktop"
3. `-one-half-from-desktop` "from desktop"

But currently they only support 1) and 3)

anandamaryon1:
Yeah I've never understood why we have the two tbh…

Why:
`<div class="nhsuk-grid-column-one-half nhsuk-u-one-half">`

Isn't just:
`<div class="nhsuk-u-one-half">`

So the `-always` class instead of that makes sense to me. Maybe in an ideal world we wouldn't need `-always` and it'd be default, but as currently and with gov, it's very opinionated and stacks at mobile by default, so I think we need to retain this to not make it such a big change.  

On the naming, I think naming from mobile-first makes sense (it's been long enough!). Part of me wants to say `-always` should be `-from-mobile` but then we'd have to have all of them being `from` or `until`. 

On your question of following gov.uk and removing the "always applied" width classes, I feel this would be a step too far given it's been available for a long time. 

In summary: 
* Yes to replacing `nhsuk-u-one-half` with `nhsuk-grid-column-one-half-always `
* Yes to removing / renaming the now redundant `-tablet` classes
* I think no to mobile only classes (should be overridden at tablet if you want mobile only). Although, now I write that, it doesn't make sense with `-always` 🤔  … now I'm getting confused! This is why you've added the `until-tablet` classes is it? 

Thanks for raising @colinrotherham and making the test page, it helps a lot

colinrotherham:
Thanks @anandamaryon1

Yeah I only picked `-always` because we needed a replacement for `nhsuk-u-[grid-size]`

I'd be happy to remove `-until-tablet` and add `-from-mobile`?

It would align with our lower-level of stacking where we only make prefixes and suffixes inline `$from: mobile`

### Watch breakpoint (224px)

Text input with prefix and suffix at 224px watch breakpoint

### Mobile breakpoint (320px)

Text input with prefix and suffix at 320px mobile breakpoint

anandamaryon1:
Yeah I'd prefer `-from-mobile` but then we do need to think of a replacement for `-always` because that name suggests you can't override it. 🤔 

And yeah, I don't mind forcing stacked on watch / sub-mobile size.

colinrotherham:
>And yeah, I don't mind forcing stacked on watch / sub-mobile size.

In that case, we can tell people to use `-from-mobile` instead?

We don't need `-always` if watch / sub-mobile should stack

anandamaryon1:
> > And yeah, I don't mind forcing stacked on watch / sub-mobile size.
> 
> In that case, we can tell people to use `-from-mobile` instead?
> 
> We don't need `-always` if watch / sub-mobile should stack

Ah yes of course. 👍 

Does this mean we would not have the plain ones, eg. `nhsuk-grid-column-one-half`, and they'd all have a `-from`? I suppose that's ok?

colinrotherham:
We'd still have the plain ones

But because the mobile/tablet versions aren't modifiers like `--from-tablet` you wouldn't need the default class too

Every grid column class can work on its own:

* `nhsuk-grid-column-one-half` "from tablet" by default
* `nhsuk-grid-column-one-half-from-mobile`
* `nhsuk-grid-column-one-half-from-desktop`

E.g. **One half from mobile**

```html
<div class="nhsuk-grid-row">
  <div class="nhsuk-grid-column-one-half-from-mobile">
    <!-- Left side at 320px -->
  </div>
  <div class="nhsuk-grid-column-one-half-from-mobile">
    <!-- Right side at 320px -->
  </div>
</div>
```

anandamaryon1:
Ah yes ok, so we omit `-from-tablet` because we expect that to be the default?

colinrotherham:
>Ah yes ok, so we omit `-from-tablet` because we expect that to be the default?

Yeah, in the same way `nhsuk-grid-column-one-half` implies "from desktop" currently in v10

The default grid column breakpoint will be changing from desktop to tablet

frankieroberto:
I'm finding this all a bit confusing and hard to follow, sorry! I might well be misunderstanding.

Are we saying that there are essentially 3 options for when something should be "one half"?

|     | Mobile | Tablet | Desktop |
|-----|--------|--------|---------|
| `nhsuk-grid-column-one-half-always` | 50% | 50% | 50% |
| `nhsuk-grid-column-one-half-from-tablet` | 100% | 50% | 50% |
| `nhsuk-grid-column-one-half-from-desktop` | 100% | 100% | 50% |

If so, how about this naming scheme:

|     | Mobile | Tablet | Desktop |
|-----|--------|--------|---------|
| `nhsuk-grid-column-one-half-always` | 50% | 50% | 50% |
| `nhsuk-grid-column-one-half` (default option) | 100% | 50% | 50% |
| `nhsuk-grid-column-one-half-desktop-only` | 100% | 100% | 50% | 


However, I can't really see any use case for using `nhsuk-grid-column-one-half-always` and having a 50% width column on mobile? That could be end up super narrow?!

colinrotherham:
We're scrapping `-always` now

Only two extra ways

* Adding `-from-desktop` from GOV.UK Frontend
* Adding `-from-mobile` for completeness

anandamaryon1:
Started to update the guidance to help get my head around it, does this look OK? 
image

colinrotherham:
@anandamaryon1 Would you mind if we didn't mix the two?

Have a look at the [Layout: **Desktop specific grid classes** guidance](https://design-system.service.gov.uk/styles/layout/#desktop-specific-grid-classes) on the GOV.UK Design System

I'd have one section for desktop, one for mobile

If advanced users want to combine the two they can?

colinrotherham:
Code is ready for review again

@frankieroberto We're now using your suggestion but with `-always` changed to `-mobile`

|                                               | Mobile | Tablet | Desktop |
| --------------------------------------------- | ------ | ------ | ------- |
| `nhsuk-grid-column-one-half-from-mobile`      | 50%    | 50%    | 50%     |
| `nhsuk-grid-column-one-half` (default option) | 100%   | 50%    | 50%     |
| `nhsuk-grid-column-one-half-from-desktop`     | 100%   | 100%   | 50%     |

But below 320px column will never apply

We'll always use 100% as noted by @anandamaryon1 in [this comment](https://github.com/nhsuk/nhsuk-frontend/pull/1296#issuecomment-3606395574):

>And yeah, I don't mind forcing stacked on watch / sub-mobile size.

frankieroberto:
@colinrotherham I still find it kinda confusing, but maybe some good guidance would help? (Could even do some user testing on it?!?)

`-from-desktop` feels confusing to me, as that basically means it's a column on desktop only? (there's nothing bigger than a desktop?)

`from-mobile` feels a bit more intuitive to me, but I wouldn't have guessed that < 320px width wouldn't count - I'd still describe a 300px wide screen as 'mobile', even if there aren't many mobile screens that narrow!

I'm just 1 user though!

colinrotherham:
>@colinrotherham I still find it kinda confusing, but maybe some good guidance would help? (Could even do some user testing on it?!?)

Yeah, we're assuming `-from-desktop` has worked well for GOV.UK Frontend since [it was added in 2019](https://github.com/alphagov/govuk-frontend/pull/1123)

I don’t think we want to go in a competing direction really 😔

We can review [their guidance](https://design-system.service.gov.uk/styles/layout/#desktop-specific-grid-classes) though?

>### Desktop specific grid classes
>
>To specify a width at the desktop breakpoint you can use the desktop specific grid classes. For example `govuk-grid-column-two-thirds-from-desktop` will set your column width to be two-thirds width at the desktop breakpoint only.

Appreciate they only implemented `-from-desktop` so feedback on `-from-mobile` is very much welcome

Would it help to clarifiy that the GOV.UK Frontend naming convention is actually `-from-[breakpoint]`?

Omitting the new default (tablet) we're left with:

* **320px+** `-from-mobile`
* ~**641px+** `-from-tablet`~
* **769px+** `-from-desktop`
* **990px+** `-from-large-desktop`

I've not added `-from-large-desktop` to this PR as it's probably unnecessary for now

packages/nhsuk-frontend/src/nhsuk/core/objects/_grid.scss
```diff
@@ -8,16 +8,57 @@
...
+          ".nhsuk-grid-column-#{$width}",
+          ".nhsuk-grid-column-#{$width}-from-#{$breakpoint}"
+        );
+
+        #{$grid-column-selector} {
+          @include nhsuk-grid-column-width($width);
+        }
+      }
     }
   }
```
colinrotherham, line 63: This work did flag one problem with the current GOV.UK Frontend approach

By looping widths-then-breakpoints, Sass generates some quite repetitive output:

```scss
// These *must* be defined in a separate loop as they have the same
// specificity as the non-breakpoint specific classes, so need to appear after
// them in the outputted CSS
@each $width in map-keys($govuk-grid-widths) {
  .govuk-grid-column-#{$width}-from-desktop {
    @include govuk-grid-column($width, $at: desktop);
  }
}
```

To fix this, I added an `@extend %nhsuk-grid-column;` for all column defaults (no breakpoints)

Followed by this new loop of ~widths-then-breakpoints~ breakpoints-then-widths

## Reduced CSS output

Here's a diff of the new CSS output:

```patch
-.nhsuk-grid-column-one-quarter-from-mobile {
   box-sizing: border-box;
   width: 100%;
   padding: 0 16px;
 }
+.nhsuk-grid-column-full-from-desktop, .nhsuk-grid-column-full, .nhsuk-grid-column-full-from-mobile, .nhsuk-grid-column-three-quarters-from-desktop, .nhsuk-grid-column-three-quarters, .nhsuk-grid-column-three-quarters-from-mobile, .nhsuk-grid-column-two-thirds-from-desktop, .nhsuk-grid-column-two-thirds, .nhsuk-grid-column-two-thirds-from-mobile, .nhsuk-grid-column-one-half-from-desktop, .nhsuk-grid-column-one-half, .nhsuk-grid-column-one-half-from-mobile, .nhsuk-grid-column-one-third-from-desktop, .nhsuk-grid-column-one-third, .nhsuk-grid-column-one-third-from-mobile, .nhsuk-grid-column-one-quarter-from-desktop, .nhsuk-grid-column-one-quarter, .nhsuk-grid-column-one-quarter-from-mobile, .nhsuk-grid-column {
+
 @media (min-width: 20em) {
   .nhsuk-grid-column-one-quarter-from-mobile {
     width: 25%;
     float: left;
   }
-}
-
-.nhsuk-grid-column-one-third-from-mobile {
-  box-sizing: border-box;
-  width: 100%;
-  padding: 0 16px;
-}
-@media (min-width: 20em) {
   .nhsuk-grid-column-one-third-from-mobile {
     width: 33.3333333333%;
     float: left;
   }
-}
-
-.nhsuk-grid-column-one-half-from-mobile {
-  box-sizing: border-box;
-  width: 100%;
-  padding: 0 16px;
-}
-@media (min-width: 20em) {
   .nhsuk-grid-column-one-half-from-mobile {
     width: 50%;
     float: left;
   }
-}
-
-.nhsuk-grid-column-two-thirds-from-mobile {
-  box-sizing: border-box;
-  width: 100%;
-  padding: 0 16px;
-}
-@media (min-width: 20em) {
   .nhsuk-grid-column-two-thirds-from-mobile {
     width: 66.6666666667%;
     float: left;
   }
-}
-
-.nhsuk-grid-column-three-quarters-from-mobile {
-  box-sizing: border-box;
-  width: 100%;
-  padding: 0 16px;
-}
-@media (min-width: 20em) {
   .nhsuk-grid-column-three-quarters-from-mobile {
     width: 75%;
     float: left;
   }
-}
-
-.nhsuk-grid-column-full-from-mobile {
-  box-sizing: border-box;
-  width: 100%;
-  padding: 0 16px;
-}
-@media (min-width: 20em) {
   .nhsuk-grid-column-full-from-mobile {
     width: 100%;
     float: left;
   }
 }
-
-.nhsuk-grid-column-one-quarter {
-  box-sizing: border-box;
-  width: 100%;
-  padding: 0 16px;
-}
 @media (min-width: 40.0625em) {
   .nhsuk-grid-column-one-quarter {
     width: 25%;
     float: left;
   }
-}
-
-.nhsuk-grid-column-one-third {
-  box-sizing: border-box;
-  width: 100%;
-  padding: 0 16px;
-}
-@media (min-width: 40.0625em) {
   .nhsuk-grid-column-one-third {
     width: 33.3333333333%;
     float: left;
   }
-}
-
-.nhsuk-grid-column-one-half {
-  box-sizing: border-box;
-  width: 100%;
-  padding: 0 16px;
-}
-@media (min-width: 40.0625em) {
   .nhsuk-grid-column-one-half {
     width: 50%;
     float: left;
   }
-}
-
-.nhsuk-grid-column-two-thirds {
-  box-sizing: border-box;
-  width: 100%;
-  padding: 0 16px;
-}
-@media (min-width: 40.0625em) {
   .nhsuk-grid-column-two-thirds {
     width: 66.6666666667%;
     float: left;
   }
-}
-
-.nhsuk-grid-column-three-quarters {
-  box-sizing: border-box;
-  width: 100%;
-  padding: 0 16px;
-}
-@media (min-width: 40.0625em) {
   .nhsuk-grid-column-three-quarters {
     width: 75%;
     float: left;
   }
-}
-
-.nhsuk-grid-column-full {
-  box-sizing: border-box;
-  width: 100%;
-  padding: 0 16px;
-}
-@media (min-width: 40.0625em) {
   .nhsuk-grid-column-full {
     width: 100%;
     float: left;
   }
 }
-
-.nhsuk-grid-column-one-quarter-from-desktop {
-  box-sizing: border-box;
-  width: 100%;
-  padding: 0 16px;
-}
 @media (min-width: 48.0625em) {
   .nhsuk-grid-column-one-quarter-from-desktop {
     width: 25%;
     float: left;
   }
-}
-
-.nhsuk-grid-column-one-third-from-desktop {
-  box-sizing: border-box;
-  width: 100%;
-  padding: 0 16px;
-}
-@media (min-width: 48.0625em) {
   .nhsuk-grid-column-one-third-from-desktop {
     width: 33.3333333333%;
     float: left;
   }
-}
-
-.nhsuk-grid-column-one-half-from-desktop {
-  box-sizing: border-box;
-  width: 100%;
-  padding: 0 16px;
-}
-@media (min-width: 48.0625em) {
   .nhsuk-grid-column-one-half-from-desktop {
     width: 50%;
     float: left;
   }
-}
-
-.nhsuk-grid-column-two-thirds-from-desktop {
-  box-sizing: border-box;
-  width: 100%;
-  padding: 0 16px;
-}
-@media (min-width: 48.0625em) {
   .nhsuk-grid-column-two-thirds-from-desktop {
     width: 66.6666666667%;
     float: left;
   }
-}
-
-.nhsuk-grid-column-three-quarters-from-desktop {
-  box-sizing: border-box;
-  width: 100%;
-  padding: 0 16px;
-}
-@media (min-width: 48.0625em) {
   .nhsuk-grid-column-three-quarters-from-desktop {
     width: 75%;
     float: left;
   }
-}
-
-.nhsuk-grid-column-full-from-desktop {
-  box-sizing: border-box;
-  width: 100%;
-  padding: 0 16px;
-}
-@media (min-width: 48.0625em) {
   .nhsuk-grid-column-full-from-desktop {
     width: 100%;
     float: left;
```

PR #1309, opened 2025-05-20
Add button group and full width button styles

colinrotherham:
## Description

This PR aligns with GOV.UK Frontend by including:

1. Switch to full width buttons on mobile
2. Add new "button group" styling

The styles for `.nhsuk-button-group` were initially copied from GOV.UK Frontend:

* [govuk-frontend/src/govuk/**objects/_button-group.scss**](https://github.com/alphagov/govuk-frontend/blob/v5.10.2/packages/govuk-frontend/src/govuk/objects/_button-group.scss)

But have now been updated to match NHS App Frontend spacing:

* [nhsapp-frontend/src/components/**button/_index.scss**](https://github.com/nhsuk/nhsapp-frontend/blob/v3.1.0/src/components/button/_index.scss#L94-L162)

<br>

Button group

<br>
<br>

From [NHS App Frontend **Button** component](https://design-system.nhsapp.service.nhs.uk/components/buttons/) research:

<br>

>### Full width buttons
>Testing on the NHS App has found that full width buttons are not a problem for users with smaller screens. But we will continue to test if:
>
>* full width buttons are problematic for users with low digital skills
> * users don’t actually see them as actionable things to click

## Checklist

- [ ] CHANGELOG entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -278,9 +278,9 @@ It was already possible to set the page app width with the `$nhsuk-page-width` v
 
 This change was introduced in [#1412: Allow creating custom width containers](https://github.com/nhsuk/nhsuk-frontend/pull/1412)
 
-#### Buttons are now full-width on mobile
+#### Buttons are now full width on mobile
 
-All buttons now display as full-width at mobile size. This does not require any changes, but you should check that it does not break any of your layouts.
+All buttons are now displayed at full width on mobile. This may not require any changes, but you should check that it does not break any of your layouts.
 
 This change was introduced in [pull request #1309: Add button group and full width button styles](https://github.com/nhsuk/nhsuk-frontend/pull/1309)
 


packages/core/objects/_button-group.scss
```diff
@@ -0,0 +1,78 @@
...
+    $link-spacing: nhsuk-spacing(1);
+
+    // Flexbox is used to center-align links on mobile, align everything along
+    // the baseline on tablet and above, and to removes extra whitespace that
+    // we'd get between the buttons and links because they're inline-blocks.
+    //
+    // Ideally we'd use `gap` with flexbox rather than having to do it all with
+    // margins, but unfortunately the support isn't there (yet) and @supports
+    // doesn't play nicely with it
+    // (https://github.com/w3c/csswg-drafts/issues/3559)
```
frankieroberto, line 30: We're using `gap` for the new Header account section, so maybe support is ok for it now? https://github.com/nhsuk/nhsuk-frontend/pull/1058/files#diff-9872b34d22a8af3575af4913b03966d1994a4c9e966fddbaa977458158892ca1R119 cc @paulrobertlloyd

paulrobertlloyd: We use `gap` in the header, but not as much as we once did; the navigation used it, but it meant items were bunched up on older versions of Safari locked-to-the-last-support-version-of-iOS. Probably something we’d want to test across a number of different devices.

frankieroberto:
One factor to consider is the mobile view. In govuk-frontend, buttons go full-width on mobile, and so it makes sense to also vertically stack buttons group at mobile widths.

However NHS frontend does not make buttons full-width on mobile views, so we _could_ continue to allow buttons to be side-by-side in a button group on mobile, however they might not always fit.

Here's how it currently looks at 640 width vs 641:

| 640 width | 641 width |
|--------------|--------------|
| Screenshot 2025-05-20 at 16 12 31 | Screenshot 2025-05-20 at 16 12 46 |

Tricky!

frankieroberto:
The NHS App team have already added button groups: https://design-system.nhsapp.service.nhs.uk/components/buttons/#grouping-buttons - they also have a `nhsapp-button` modifier which makes buttons full-width on mobile.

We could consider something similar and have the behaviour opt-in, perhaps via `nhsuk-button--full-width-on-mobile` or similar?

paulrobertlloyd:
Please, please, please can we replicate the GOV.UK behaviour and have buttons go full width on mobile by default? (I've been using my button group wrapper as a bit of a hack to enable this)

paulrobertlloyd:
Please, please, please can we replicate the GOV.UK behaviour and have buttons go full width on mobile by default? (I've been using my button group wrapper as a bit of a hack to enable this)

frankieroberto:
> Please, please, please can we replicate the GOV.UK behaviour and have buttons go full width on mobile by default? (I've been using my button group wrapper as a bit of a hack to enable this)

@paulrobertlloyd Which breakpoint would you pick? 640 has always felt a bit too wide to me, but maybe its fine? 🤷 

Screenshot 2025-05-20 at 18 11 34

How about making it opt-in or opt-out? 🤔

paulrobertlloyd:
Be bold @frankieroberto! < 640px is perfectly fine! (What do they do on the app?)

frankieroberto:
> Be bold @frankieroberto! < 640px is perfectly fine! (What do they do on the app?)

The app has a modifier to make them full-width on mobile: https://github.com/nhsuk/nhsapp-frontend/blob/main/src/components/button/_index.scss#L14-L23 - I think @davidhunter08 and co did some research on it.

frankieroberto:
Happy to go with option 1 - full width buttons on mobile. Maybe it doesn't even need an opt-out?

Would we consider it a breaking change though? 🤔

paulrobertlloyd:
Yaaaaay!!

colinrotherham:
Updated to match the spacing in NHS App Frontend

* [src/components/**button/_index.scss**](https://github.com/nhsuk/nhsapp-frontend/blob/v3.1.0/src/components/button/_index.scss#L94-L162)

//cc @davidhunter08

frankieroberto (approved):
Nice! Looks good to me.

Here's how I'm planning to use it:

Screenshot 2025-06-12 at 17 14 29

frankieroberto:
One thing this pull request doesn't currently do is let you add non-linked text next to the button:

Should we support that too? 🤔 

Screenshot 2025-06-13 at 12 44 35

colinrotherham:
Thanks @frankieroberto

Ahh in GOV.UK Frontend `.govuk-button-group .govuk-link` gets a bottom margin like `govuk-body`

E.g. You'd want to wrap your "or" text into `<span class="nhsuk-body-m">or</span>`

But looks like we need to iterate the NHS App spacing a little more for that

colinrotherham:
Thanks @anandamaryon1, let me turn the review app back on

anandamaryon1:
Happy with the new default of full width buttons, since they've been used already on GOV, NHS App, NHS Login and 111Online, and others, for a while now. 

Added a few more examples to test wording and links beside buttons.

frankieroberto (approved):

CHANGELOG.md
```diff
@@ -272,6 +278,12 @@ It was already possible to set the page app width with the `$nhsuk-page-width` v
 
 This change was introduced in [#1412: Allow creating custom width containers](https://github.com/nhsuk/nhsuk-frontend/pull/1412)
 
+#### Buttons are now full-width on mobile
+
+All buttons now display as full-width at mobile size. This does not require any changes, but you should check that it does not break any of your layouts.
```
colinrotherham: Minor tweaks, don't think we hyphenate full-width?

Probably change "does not" to "may not" because we do want users to check just in case

```suggestion
#### Buttons are now full width on mobile

All buttons are now displayed at full width on mobile. This may not require any changes, but you should check that it does not break any of your layouts.
```

colinrotherham:
Brilliant, shall we merge this and do testing as part of the pre-release?

Noting that the styles were largely taken from the NHS App

Issue #1311, opened 2025-05-20
Breadcrumb and back link margin collapses into header menu


colinrotherham:
# Bug Report

## What is the issue?

1. Back links use fixed `margin-top` within `nhsuk-width-container`
2. Breadcrumbs use fixed `margin-top` within `nhsuk-width-container`
3. Header menu uses a dynamic `margin-bottom` to reserve space for menu items

We should use ~`margin-top`~ `padding-top` to prevent margin collapse:

<img width="615" alt="Image" src="https://github.com/user-attachments/assets/0b95264f-8ad1-4bc8-8ad5-748d03b6b7ba" />

## What steps are required to reproduce the issue?

Open the header menu when the back link or breadcrumbs are visible.

PR #1312, opened 2025-05-20
Resolve inconsistent link styles

colinrotherham:
## Description

Fixed a few link consistency issues whilst investigating https://github.com/nhsuk/nhsuk-frontend/issues/1311

For example, the breadcrumbs component as a back link on mobile:

### Before
No underline on `:hover` with mismatched icon colour

Breadcrumbs back link before

### After
Underline on `:hover` with matching icon colour

Breadcrumbs back link after

## Checklist

- [ ] CHANGELOG entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,6 +63,7 @@ This change was made in [pull request #1259: Review legacy Nunjucks params](http
 
 :wrench: **Fixes**
 
+- [#1312: Review and fix inconsistent link styles](https://github.com/nhsuk/nhsuk-frontend/pull/1312)
 - [#1318: Consistent padding for card and expander](https://github.com/nhsuk/nhsuk-frontend/pull/1318)
 - [#1324: Show border on summary row cells with no action link](https://github.com/nhsuk/nhsuk-frontend/pull/1324)
 - [#1324: Increase spacing between paragraphs in summary list values](https://github.com/nhsuk/nhsuk-frontend/pull/1324)


frankieroberto:
@colinrotherham the regular back links aren't underlined, as [has been previously noticed](https://github.com/nhsuk/nhsuk-service-manual-community-backlog/issues/32#issuecomment-2405546950) but I’m not sure we know why.

MatMoore:
I noticed that in the GOV.UK design system, there is a `text-underline-offset` for all links (presumably to help with legibility) but the NHS design system doesn't have that...

Not a blocker for this PR as it's the same for regular links, but just wondering if this is something that has been discussed before?

chrimesdev (approved):

paulrobertlloyd:
Not sure if this is the place to make such a change, but it’s always confused me why NHS.UK Frontend uses purple for the hover state, not the visited state (which is typically a browser’s default, and the style used by GOV.UK Frontend). Would we be open to changing this?

frankieroberto:
> Not sure if this is the place to make such a change, but it’s always confused me why NHS.UK Frontend uses purple for the hover state, not the visited state (which is typically a browser’s default, and the style used by GOV.UK Frontend). Would we be open to changing this?

Yeah, I think we could explore doing something similar to GOV.UK and going a slightly darker blue on hover and with a thicker underline? Open a separate issue for this though?

paulrobertlloyd:
Same also for which links get underlined, and which don’t. Also a bit inconsistent. But perhaps something to be addressed in a separate PR. This is a good first step, however.

packages/components/header/_header-base.scss
```diff
@@ -78,8 +78,9 @@
 }
 
 .nhsuk-header__link {
-  @include nhsuk-logo-size;
   display: block;
+  @include nhsuk-logo-size;
+  @include nhsuk-link-style-white;
```
paulrobertlloyd, line 83: Might need to check how this lands when rebased into the `header-breaking-changes` branch; suspect this might not be needed, but can’t remember.

packages/components/header/_header-base.scss
```diff
@@ -344,6 +345,7 @@
   text-decoration: underline;
   white-space: nowrap;
   @include nhsuk-font(16);
+  @include nhsuk-link-style-white;
```
paulrobertlloyd, line 348: Again, might need to see how this lands in the `header-breaking-changes` branch. That uses a new, separate mixin for link styles as they need to inherit the text colour used in the white and blue header variants.

paulrobertlloyd (approved):
The header changes will need review when merged into the `header-breaking-changes` branch which, among other things, uses a new mixin which may no longer be needed given the new one provided by this PR.

paulrobertlloyd:
> Which bits are still inconsistent @paulrobertlloyd?

Not underlined:

- Action link
- Pagination (Previous/Next)
- Back link
- Skip link

Underlined:

Prose styles and most other components including:

- Breadcrumbs
- Pagination (descriptions below Previous/Next labels)

Issue #1169, opened 2025-03-05
Align visually hidden behaviour with GOV.UK Frontend


colinrotherham:
There are a number of issues against `visually-hidden` behaviour:

* https://github.com/nhsuk/nhsuk-frontend/issues/704
* https://github.com/nhsuk/nhsuk-frontend/issues/760
* https://github.com/nhsuk/nhsuk-frontend/issues/1168

We should see if these are resolved by GOV.UK Frontend

## Visually hidden

For example, we can see various `visually-hidden` changes in:

* https://github.com/alphagov/govuk-frontend/pull/1037
* https://github.com/alphagov/govuk-frontend/pull/1109
* https://github.com/alphagov/govuk-frontend/pull/3836
* https://github.com/alphagov/govuk-frontend/pull/5063

## Visually shown

But we should likely deprecate and remove `visually-shown` (see https://github.com/nhsuk/nhsuk-frontend/issues/1168) to prevent conflicting styles.

I'd recommend a targetted `visually-hidden` instead:

### Before

```scss
@include visually-hidden; // [1]

@include mq($from: desktop) {
  @include visually-shown($display: table-header-group); // [2]
}
```

### After

```scss
@include mq($until: desktop) {
  @include visually-hidden; // [1]
}
```

Issue #1096, opened 2025-01-09
Summary list row dividers are not full width if some rows do not have actions


frankieroberto:
If some rows don’t have actions, the horizontal divider between the rows does not go across the full width:

<img width="664" alt="Screenshot 2025-01-09 at 14 56 31" src="https://github.com/user-attachments/assets/afd0e885-6057-45a7-a02e-df1a6815f95c" />

Issue #934, opened 2024-03-25
Improve line spacing on summary list


anandamaryon1:
## What
Summary list rows with multiple values can be difficult to differentiate when text wraps onto multiple lines. 

For example:
<img width="585" alt="image" src="https://github.com/nhsuk/nhsuk-service-manual-community-backlog/assets/159916913/6a89fd31-5b0a-4df4-a944-c39a750ddeac">

Our summary list is based on GOV.UK, with our styling applied. Compared to GOV.UK, our styling has slightly more line-height, and slightly less margin between paragraphs. 

Therefore, to enhance readability, I suggest that we increase the margin slightly, from 8px to 12px. 

**Proposed design example:**
<img width="586" alt="image" src="https://github.com/nhsuk/nhsuk-service-manual-community-backlog/assets/159916913/65d1d71c-b784-46ca-baaa-2ec944c8f041">

## Why
To improve readability of summary lists that run onto multiple lines, making lists easier to read and scan, therefore improving the usability of the summary list component.

Issue #866, opened 2023-03-31
Can't apply ids to row keys, values and anchor links on summary list component


adeel718:
# Bug Report

On the summary list component it is currently not possible to add ids to rows.keys, rows.values or the action item hrefs.

## What is the issue?

There should be the ability to add ids

## What steps are required to reproduce the issue?

Create a summary list component

## Is there anything else you think would be useful in recreating the issue?

![image](https://user-images.githubusercontent.com/6770058/229145272-83403906-945a-46a6-927e-47a093a8fa94.png)

GWRowley:
Suggested fix for _actions_ in **components/summary-list/template.njk** lines 1-8

`{%- macro _link(action) %}
  <a href="{{ action.href }}" {%- if action.classes %} class="{{ action.classes }}"{% endif %} {%- if action.id %} id="{{ action.id }}"{% endif %}>
    {{ action.html | safe if action.html else action.text }}
    {%- if action.visuallyHiddenText -%}
      <span class="nhsuk-u-visually-hidden"> {{ action.visuallyHiddenText }}</span>
    {% endif -%}
  </a>
{% endmacro -%}`

colinrotherham:
This has been completed in:

* https://github.com/nhsuk/nhsuk-frontend/pull/2047

With the following options:

* Summary list rows now support `id` and `attributes` options
* Summary list key, value and actions now support `id`, `width` and `attributes` options

PR #1324, opened 2025-05-23
Show border on summary row cells with no action link

paulrobertlloyd:
## Description

In a summary list where some rows have actions and other don’t, those that don’t have an action show without a bottom border.  

This PR downstream the fixes from GOV.UK Frontend that add an empty pseudo element to any summary list row that doesn’t contain any actions, in a summary list where one or more the rows does contain an action.

I’ve updated the visual regression tests, so that the example with action links includes a row without an action link, so as to validate this change:

Screenshot of a summary list with one row without an action link.

## Checklist

- [ ] CHANGELOG entry


paulrobertlloyd:
@colinrotherham This PR:

- Fixes #1096 
- Fixes #934 as an extra addition

As to the others

- #504 was fixed by #1008
- #704 was possibly fixed #1169 (but likely needs retesting)

As for #866, that sounds like a separate PR, and is not something supppored by the equivalent GOV.UK Frontend components.

PR #1349, opened 2025-06-02
Add select component `value`, `disabled` and `formGroup` params

### LINKED ISSUE #1347: Support value param for select
The select component is currently missing the `value` option, unlike radios in https://github.com/nhsuk/nhsuk-frontend/issues/988

For parity with other components we should likely add:

* `disabled`
* `formGroup`
* `value`

With the param `describedBy` to be considered in another issue similar to:

* https://github.com/alphagov/govuk-frontend/pull/1347

colinrotherham:
## Description

Closes #1347 and updates all select component examples

## Checklist

- [x] CHANGELOG entry

### DIFF
--- a/app/components/select/hint.njk
+++ b/app/components/select/hint.njk
@@ -14,11 +14,11 @@
         "hint": {
           "text": "This can be different to where you went before"
         },
+        "value": "choose",
         "items": [
           {
             "value": "choose",
-            "text": "Choose location",
-            "selected": true
+            "text": "Choose location"
           },
           {
             "value": "eastmidlands",
--- a/app/components/all.njk
+++ b/app/components/all.njk
@@ -832,6 +832,7 @@
           "hint": {
             "text": "If you have more than 1 nationality, select all options that are relevant to you"
           },
+          "values": ["british"],
           "items": [
             {
               "value": "british",
@@ -860,6 +861,7 @@
         "hint": {
           "text": "This includes changing your last name or spelling your name differently"
         },
+        "value": "no",
         "items": [
           {
             "value": "yes",
@@ -867,8 +869,7 @@
           },
           {
             "value": "no",
-            "text": "No",
-            "checked": true
+            "text": "No"
           }
         ]
       }) }}
@@ -929,6 +930,7 @@
         "label": {
           "text": "Sort by"
         },
+        "value": "updated",
         "items": [
           {
             "value": "published",
@@ -936,8 +938,7 @@
           },
           {
             "value": "updated",
-            "text": "Recently updated",
-            "selected": true
+            "text": "Recently updated"
           },
           {
             "value": "views",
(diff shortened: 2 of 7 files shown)


packages/components/select/macro-options.json
```diff
@@ -70,6 +82,20 @@
...
+    "name": "formGroup",
+    "type": "object",
+    "required": false,
+    "description": "Additional options for the form group containing the select component.",
+    "params": [
+      {
+        "name": "classes",
+        "type": "string",
+        "required": false,
+        "description": "Classes to add to the form group (for example to show error state for the whole group)."
```
frankieroberto, line 95: The example in this description, whilst matching the govuk one, doesn't make sense to me, as for the error state you’d use the `errorMessage` option which sets the classes automatically? And what does "whole group" mean here? 🤔

app/components/select/hint.njk
```diff
@@ -0,0 +1,64 @@
...
+          "text": "Choose location"
+        },
+        "hint": {
+          "text": "This can be different to where you went before"
+        },
+        "items": [
+          {
+            "value": "choose",
+            "text": "Choose location",
+            "selected": true
```
frankieroberto: Worth moving this to be `value: "choose"` at the top level, now we have that option?

packages/components/select/macro-options.json
```diff
@@ -49,6 +49,18 @@
       }
     ]
   },
+  {
+    "name": "value",
+    "type": "string",
+    "required": false,
+    "description": "Value for the option which should be selected. Use this as an alternative to setting the `selected` option on each individual item."
```
frankieroberto, line 56: Could this maybe be a bit clearer? It's potentially a bit confusing, as both things are called `value`. 😬  Maybe we should spell out that this needs to exactly match the value of the option you want to be initially selected?

```suggestion
    "description": "Use this to make an option initially selected when the page loads. The value should exactly match the value of the option to be selected. This is an alternative to setting the `selected` boolean option on each individual item."
```

frankieroberto (approved):
Nice! Left some minor comments

packages/components/select/macro-options.json
colinrotherham: Not sure really, we use it everywhere `formGroup` exists 😬

One to fix in another issue?

packages/components/select/macro-options.json
colinrotherham: Another one where I've copied the GOV.UK Frontend text

Might be a nice example where we can remove our component (in place of theirs) so I've kept it as-is

app/components/select/hint.njk
colinrotherham: Made sense to change them all, pushed in https://github.com/nhsuk/nhsuk-frontend/pull/1349/commits/9c3ea43b406bea2b8f058b9272991f0513f7beb7

Thanks!

Issue #1357, opened 2025-06-04
Spacing between button background and box shadow


adamwhitingnhs:
At various screen sizes there is a small space between the button background and the box shadow of a button using the nhsuk-button class.

I have reproduced the issue in JSFiddle [here](https://jsfiddle.net/L8ha5qdt/1/)

Browser: Google Chrome
NHS.UK frontend package version: 9.6.1

![Image](https://github.com/user-attachments/assets/29201804-aced-46d3-a390-cf842a875954)

I have found the breaking change to have been added as part of this PR https://github.com/nhsuk/nhsuk-frontend/pull/1077/files#diff-47500cb7b50a4c38bd5df27220dac558b5a4daad155542488b051fb27769e991

The box shadow was removed from the button itself and added in the &:after selector with some absolute spacing.
Not sure why this was changed as the PR itself is referring to fixing an issue with the secondary button.

anandamaryon1:
Hi there, thanks for reporting this. We're looking into it now. 

Are you using any page or text zoom? I'm struggling to recreate it unless I zoom the page to 110%.

adamwhitingnhs:
It should be visible @ 100% zoom on a standard 1080p screen. In the JSFiddle I linked it appears to show up when the view box is smaller than 655px on a 1080p resolution, but we're seeing it on larger screen widths than that at 100% zoom.

colinrotherham:
Have you got scaling set above 100%, e.g. 125%, 150%?

I can't reproduce it at 100% but I've got a fix in mind for tomorrow

>The box shadow was removed from the button itself and added in the &:after selector with some absolute spacing.
>Not sure why this was changed as the PR itself is referring to fixing an issue with the secondary button.

We moved `box-shadow` to `:after` so we can inset it slightly for secondary buttons

Otherwise the button shadow radius is largely hidden behind the 2px border

<img width="537" alt="Button psuedo element shadow trick " src="https://github.com/user-attachments/assets/4b58a0e6-3456-4a3f-9866-81c6a7fe679c" />

<br>
<br>

You can see the original shadow position on the right, versus the slightly inset version moved left from the border:

<img width="673" alt="Button psuedo element shadow trick (zoom)" src="https://github.com/user-attachments/assets/cdce031d-3f3f-4280-9f61-0f08f5b0adde" />

colinrotherham:
I've opened a workaround for the browser pixel snapping issue in:

[...]

PR #1358, opened 2025-06-05
Fix spacing between button background and box shadow

colinrotherham:
## Description

This PR fixes https://github.com/nhsuk/nhsuk-frontend/issues/1357 when either:

1. Device DPI scaling is set above 100%
2. Certain page zoom percentages are applied

### Border pixel snapping

The issue appears to be intentional due to the CSS specification:
https://www.w3.org/TR/css-values-4/#snap-a-length-as-a-border-width

> **CSS Values and Units Module** Level 4
>6. Distance Units: the [<length>](https://www.w3.org/TR/css-values-4/#length-value) type
> …
>If len is greater than 1 [device pixel](https://www.w3.org/TR/css-values-4/#device-pixel), round it down to the nearest integer number of device pixels.

E.g. See how a 2px `border-width` is adjusted to `1.818181px` to snap to the nearest device pixel

<br>

Border pixel snapping to device pixels

<br>
<br>

See these Chrome browser issues for more information:

* **Different zoom ratios result in abnormal layout (element heights experience precision loss)**
  https://issues.chromium.org/issues/328539162
  
* **Fractional devicePixelRatio causes content shifts on border-width changes**
  https://issues.chromium.org/issues/384781869

### Adjustable `box-shadow` position

We moved `box-shadow` to `::after` so we can inset it slightly for secondary buttons

Otherwise the button shadow radius is largely hidden behind the 2px border

Button psuedo element shadow trick

<br>
<br>

You can see the original shadow position on the right, versus the slightly inset version moved left from the border:

Button psuedo element shadow trick (zoom)

To fix the issue this PR sets `::after` for secondary buttons only, where differences in rounding between `border-width`, `top`, `right`, `bottom`, `left` etc are hidden by the 2px outer border set on `::before`.

## Checklist

- [x] CHANGELOG entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -160,7 +160,7 @@ This change was introduced in [#1139: Remove 24px from typography scale](https:/
 - [#1324: Increase spacing between paragraphs in summary list values](https://github.com/nhsuk/nhsuk-frontend/pull/1324)
 - [#1325: Ensure summary list keys are aligned across lists](https://github.com/nhsuk/nhsuk-frontend/pull/1325)
 - [#1325: Add `nhsuk-summary-list__row--no-border` modifier to remove border on an individual summary list row](https://github.com/nhsuk/nhsuk-frontend/pull/1325)
-- [#1345: Fix primary card chevron position without description](https://github.com/nhsuk/nhsuk-frontend/pull/1345)
+- [#1346: Fix primary card chevron position without description](https://github.com/nhsuk/nhsuk-frontend/pull/1346)
 - [#1349: Add select component `value`, `disabled` and `formGroup` params](https://github.com/nhsuk/nhsuk-frontend/pull/1349)
 - [#1358: Fix spacing between button background and box shadow](https://github.com/nhsuk/nhsuk-frontend/pull/1358)
 


anandamaryon1:
Thanks for looking at this @colinrotherham 

Introduces a slight issue with the bottom border radius:
image

Seems to be fixed by:
```
.nhsuk-button--secondary-solid:not(:focus):after, .nhsuk-button--secondary:not(:focus):after {
    border-radius: 2px;
    left: 0;
    right: 0;
}
```

What was the purpose of:
```
    left: -.25px;
    right: -.25px;
```

Perhaps I'm undoing something important with this suggestion? But it my quick testing it seems OK.

colinrotherham:
@anandamaryon1 Do you mind tweaking the values to find what works?

Might get away with increasing the `border-radius: 2px` again

I'm using `-.25px` to avoid another pixel snapping issue at 110% in Safari:

Safari pixel snapping issue

anandamaryon1:
Best I can find is sticking with:

```
.nhsuk-button--secondary-solid:not(:focus):after, .nhsuk-button--secondary:not(:focus):after {
    border-radius: 2px;
    left: 0;
    right: 0;
}
```
High zoom on Chrome:
image
image


110% Safari: (worst it looks)
image
image



It seems the lessor of two evils, since it only looks slightly odd at 110% zoom on Safari. As opposed to always looking a little off, at default zoom.

colinrotherham:
@anandamaryon1 I've updated the fix so we preserve the current radius:

```css
left: 0;
right: 0;
border-radius: 2px;
```

Looks like [setting the `box-shadow` twice](https://github.com/nhsuk/nhsuk-frontend/pull/1358/commits/7ae5ca4441b19700bbb80e98064e8e56ea4f052d#diff-382ededc3f8ad291d75fd83056eb98004a1768bf97492fe6f1d6e423061133cdR165-R170) like [in the image](https://private-user-images.githubusercontent.com/415517/451603809-cdce031d-3f3f-4280-9f61-0f08f5b0adde.png?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3NDkxMjMyNjIsIm5iZiI6MTc0OTEyMjk2MiwicGF0aCI6Ii80MTU1MTcvNDUxNjAzODA5LWNkY2UwMzFkLTNmM2YtNDI4MC05ZjYxLTBmMDhmNWIwYWRkZS5wbmc_WC1BbXotQWxnb3JpdGhtPUFXUzQtSE1BQy1TSEEyNTYmWC1BbXotQ3JlZGVudGlhbD1BS0lBVkNPRFlMU0E1M1BRSzRaQSUyRjIwMjUwNjA1JTJGdXMtZWFzdC0xJTJGczMlMkZhd3M0X3JlcXVlc3QmWC1BbXotRGF0ZT0yMDI1MDYwNVQxMTI5MjJaJlgtQW16LUV4cGlyZXM9MzAwJlgtQW16LVNpZ25hdHVyZT0zODkwNzZjMzJiYTQ1N2MxOTkwYzFmMGI2OGJkMjdjMjBjYzY3YzIwOWJhZDkyM2Y2ZjVjYjhiNDNjNDJiNDQ1JlgtQW16LVNpZ25lZEhlYWRlcnM9aG9zdCJ9.LbBltMZZCs1-QVQvNtCKV0tBwdP8TfTaB7LNNrQF67M) is what's closing that 110% zoom gap in Safari

Ready for review again

anandamaryon1:
Nice. This fixes all of the solid buttons and _mostly_ fixes the secondary border/shadow, albeit with some slight imperfections at certain zooms. But on balance, I think we should put this fix out asap and hope that in future we can think of a perfect fix for the secondary button border/shadow at zoom in all browsers.

PR #1361, opened 2025-06-05
Fix appearance of focused logo and service name in header

### LINKED ISSUE #1369: NHS logo hover state not consistent in Safari
The hover state for the NHS logo is currently not consistent in desktop Safari.

It’s meant to look like this (see Google Chrome):

Image

but instead it looks like this:

Image

or on the NHS service manual website, it looks like this:

Image

This will be fixed by the upcoming NHS header changes in version 10.

paulrobertlloyd:
## Description

Fixes #1369.

Reviewing the focused header states, a few issues become apparent:

- The focused NHS logo looks broken, with the bottom shadow merging with the logo, and only the top part of the logo showing a yellow outline
- A linked service name stretches the full width of the header

> [!NOTE]
> The commentary below is out of date. I’m now using a different approach by adding a mixin to use for focusing images. See the updated fix described in https://github.com/nhsuk/nhsuk-frontend/pull/1361#issuecomment-2946847275

This PR fixes these issues by:

- Incorporating padding into the NHS logo SVG (this ensures consistent rendering, and saves having to make multiple adjustments in CSS)
- Removing the top yellow box shadow on the focused state for a linked service logo to ensure the yellow outline around the logo is evenly spaced
- Adding hover and active background colour to a linked service logo (except for in the white header variant) – these match the blue/darker blue hover/active colours used by the search submit button
- Ensuring the NHS logo is resized with the correct aspect ratio when resized for use alongside an organisation name
- Updating the padding for the organisation name and descriptor to match that of the NHS logo

and finally:

- Preventing a service name from growing to fill any remaining space in the header

## Focused NHS logo 

Before, focused:

Screenshot of default header before.

After, focused:

Screenshot of default header after, focused.

After, hover:

Screenshot of default header after, hover.

After, active:

Screenshot of default header after, active.

## Focused organiaation logo

Before:

Screenshot of organisation white header before.

After:

Screenshot of organisation white header after.

## Focused service name

Before:

Screenshot of header with focused service name before.

After:

Screenshot of header with focused service name after.

## Checklist

- [ ] CHANGELOG entry


paulrobertlloyd:
A few notes:

- Adding padding to the NHS logo directly is one approach, downside being that when it is scaled down the padding decreases too… think we just about get away with it, but it does mean the padding is not guaranteed to always be 2px.
- The focus ring for form elements (including the search input) is 4px, not 2px. However, 2px matches the space we’ve added inside navigation items. So… 🤷🏼‍♂️
- The alternative approach I tried got complicated quickly, namely because it involves trying to override mixins for link/focus styles, in certain contexts only. Maybe there’s an approach where we lean on our good friend the pseudo element, and style that, independently of any link styles. Lots of different ways to skin this particular cat.

paulrobertlloyd:
> It pushes the logo right/down a little bit—do we need to offset that?

It does, this is more of an issue for the organisational logos perhaps than when we use the NHS logo on its own. Open to other approaches to fix this!

colinrotherham:
Would the [GOV.UK Frontend `@include govuk-link-image` mixin](https://github.com/alphagov/govuk-frontend/blob/v5.10.2/packages/govuk-frontend/src/govuk/helpers/_links.scss#L367-L387) help?

Have a look at the [image links example](https://govuk-frontend-review.herokuapp.com/examples/links) with double box shadow:

Image links with double box shadow

paulrobertlloyd:
Ooooh, I’ll give it a try.

paulrobertlloyd:
Updated this PR to add a new `@nhsuk-link-image` mixin (copied over from GOV.UK Frontend). This simplifies things greatly, and no longer requires messing about with the SVG.

I’ve removed the hover and active styles from the NHS logo however, as it gets weird when this is combined with a service name.

New comparison images below.

## Focused NHS logo

Before:

Screenshot of default header before.

After:

Screenshot of default header after.

## Focused NHS logo and service name combined

Before:

Screenshot of combined header before.

After:

Screenshot of combined header after.

## Focused NHS organisation header

No change, baring one pixel increase in gap between logo and organisation name (thus the updated visual regression images, now only 22 down from 72 in the initial PR!)

Before:

Screenshot of organisation header before.

After:

Screenshot of organisation header after.

## Focused service name

This remains unchanged from the initial commit for this PR.

Before:

Screenshot of header with focused service name before.

After:

Screenshot of header with focused service name after.

***

I have also added some extra visual regression tests for the focused NHS/service name/organisation name link.

packages/nhsuk-frontend/src/nhsuk/components/header/_index.scss
```diff
@@ -60,14 +61,22 @@ $_header-item-padding: 12px;
...
+  flex-grow: 0;
+  margin-right: auto;
   text-wrap: balance;
   @include nhsuk-font(19, $line-height: 1.1);
 }
 
+.nhsuk-header__service-logo[href] {
+  @include nhsuk-link-image;
+
+  display: inline-flex;
```
paulrobertlloyd: I thought this might trigger a `mixed-dclr` warning, but I need to override the display property used by `nhsuk-link-image`, and got no linting warnings, so maybe this is fine? @colinrotherham

colinrotherham:
Forgot to mention, we need `:not(:active)` on any focus styles (buttons included)

Until we align more with GOV.UK Frontend at least

packages/nhsuk-frontend/src/nhsuk/components/header/_index.scss
colinrotherham: Yeah all good, we have no nested rules or media queries added

colinrotherham:
I've fixed `@include nhsuk-link-image` being overridden in the white header

Think we need a design opinion on whether the "expanded focus box" should have a solid outline all around

colinrotherham:
> Nice!
> 
> There’s no hover state for the logo, do we think that’s ok? Currently (in production) there’s a dark blue border, although I’ve just noticed that this doesn’t work in Safari properly for some reason...
> 
> Screenshot 2025-06-06 at 13 07 27

@frankieroberto I gave this a try by adding:

```patch
  &:hover,
  &:focus,
  &:active {
    .nhsuk-header__organisation-name,
    .nhsuk-header__organisation-name-descriptor,
    .nhsuk-header__service-name {
      color: inherit;
    }
  }
+
+ &:hover {
+   .nhsuk-header__logo {
+     outline: $nhsuk-focus-width solid transparent;
+     box-shadow: 0 0 0 4px #003d78;
+   }
+ }
```

Do you know where `#003d78` comes from, or what we'd use for the white header?

See what you mean about the Safari `:hover` issue though

anandamaryon1:
Happy with the focus states, thanks for these @paulrobertlloyd 

Seems a pity to drop the hover state, even though it was low-contrast. Could do a subtle colour or opacity change on hover, similar to buttons?

paulrobertlloyd:
Added a hover state that attempts to replicate the underline used in the service/organisation name. Some notes:

- Requires using a pseudo element because Safari refuses to render anything outside the bounding box of the SVG on hover
- Uses `currentcolor` to improve maintainability
- The metrics for the underline can be tweaked later to match the decision made in #1366

## Combined logo and service name

Screenshot of logo and service name with hover.

## Organisation name

Screenshot of organisation name with hover.

packages/nhsuk-frontend/src/nhsuk/components/header/_index.scss
```diff
@@ -60,14 +61,68 @@ $_header-item-padding: 12px;
...
+
+  &:not(:focus):hover {
+    text-decoration: underline;
+  }
+
+  // It should be possible to add a box shadow directly to .nhsuk-header_logo
+  // however Safari doesn’t render anything beyond the bounding box of the SVG.
+  // By adding a pseudo element with the same dimensions, we style this instead.
+  &:hover:not(:focus) {
+    position: relative;
```
colinrotherham:
```suggestion
  // It should be possible to add a box shadow directly to .nhsuk-header_logo
  // however Safari doesn’t render anything beyond the bounding box of the SVG.
  // By adding a pseudo element with the same dimensions, we style this instead.
  &:not(:focus):hover {
    position: relative;
    text-decoration: underline;
```

colinrotherham: Sorry, ended up adding a commit (did the same to organisation styles too)

Helps for find and replace when skipping between the same selectors in both blocks

colinrotherham:
@anandamaryon1 Can you check the SVG logo underline in [the comment above](https://github.com/nhsuk/nhsuk-frontend/pull/1361#issuecomment-2963192358)

Seems like a solid workaround to avoid the Safari issue

But should it inherit the text colour or stay blue?

paulrobertlloyd:
> But should it inherit the text colour or stay blue?

This might be a moot point once we change the link hover colour to dark blue?

colinrotherham:
Obviously @paulrobertlloyd found a brilliant workaround, but!

What if we use his same `::before` workaround to stretch a border around the SVG?

Just pushed the old hover style back up, but feel free to drop the commit if we prefer the underline

Screenshot 2025-06-12 at 22 21 06

<br>
<br>

Screenshot 2025-06-12 at 22 20 42

paulrobertlloyd:
I tried to recreate the previous hover state originally once I figured out the work around, but then decided on the underline because it:

1. is visually distinguishable (squint and you can see the underline whereas the dark border looks blurred)
2. requires less CSS
3. mimics the hover style used elsewhere in NHS frontend, but also mimics that used for the logo hover state on GOV.UK

colinrotherham:
No worries, it's gone. Like it never happened 😶

paulrobertlloyd:
All I’ll add is that, when the link underline stuff gets merged, we should update the line-weight for the logo hover to match.

PR #1366, opened 2025-06-06
Add link underline offset and make underline thicker on hover

frankieroberto:
## Description

This PR updates links (and link-like elements such as details) so that the underline consistently thinner and offset further away from the link text, improving readability.

It also introduces new mixins and modifier classes for:

- text colour links
- muted colour links
- links without underlines

| Before | After |
|----|-----|
| Links before  | Links after |

## Checklist

- [x] CHANGELOG entry

### DIFF
--- a/packages/nhsuk-frontend/src/nhsuk/components/back-link/_index.scss
+++ b/packages/nhsuk-frontend/src/nhsuk/components/back-link/_index.scss
@@ -48,6 +48,6 @@
   }
 
   .nhsuk-back-link--reverse {
-    @include nhsuk-link-style-white;
+    @include nhsuk-link-style-reverse;
   }
 }
--- a/packages/nhsuk-frontend/src/nhsuk/components/breadcrumb/_index.scss
+++ b/packages/nhsuk-frontend/src/nhsuk/components/breadcrumb/_index.scss
@@ -89,7 +89,7 @@
 
     .nhsuk-breadcrumb__link,
     .nhsuk-back-link {
-      @include nhsuk-link-style-white;
+      @include nhsuk-link-style-reverse;
     }
 
     .nhsuk-breadcrumb__list-item::before {
--- a/packages/nhsuk-frontend/src/nhsuk/core/tools/_links.scss
+++ b/packages/nhsuk-frontend/src/nhsuk/core/tools/_links.scss
@@ -192,16 +192,32 @@
   }
 }
 
-/// White link styling, used in hero
-/// component, and in the `.nhsuk-link--reverse`
-/// modifier class.
+/// Reverse link styles
+///
+/// Makes links white, in all states. Use this mixin if you're displaying links
+/// against a dark background.
+///
+/// If you use this mixin in a component, you must also include the
+/// `nhsuk-link-style-default` mixin to get the correct focus and hover states.
 ///
 /// @example scss
-///   @include nhsuk-link-style-white;
+///   .nhsuk-component__link {
+///     @include nhsuk-link-style-default;
+///     @include nhsuk-link-style-reverse;
+///   }
 ///
 
-@mixin nhsuk-link-style-white {
+@mixin nhsuk-link-style-reverse {
   @include nhsuk-link-style-text($override-colour: $nhsuk-reverse-text-colour);
+}
+
+/// White link styles (deprecated)
+///
+/// @alias nhsuk-link-style-reverse
+/// @deprecated To be removed in v12.0, replaced by nhsuk-link-style-reverse
+
+@mixin nhsuk-link-style-white {
+  @include nhsuk-link-style-reverse;
 }
 
 /// Default link visited only styling
--- a/packages/nhsuk-frontend/src/nhsuk/components/action-link/_index.scss
+++ b/packages/nhsuk-frontend/src/nhsuk/components/action-link/_index.scss
@@ -60,7 +60,7 @@
   }
 
   .nhsuk-action-link--reverse {
-    @include nhsuk-link-style-white;
+    @include nhsuk-link-style-reverse;
 
     &:not(:focus) .nhsuk-icon {
       // stylelint-disable-next-line declaration-no-important
--- a/packages/nhsuk-frontend/src/nhsuk/components/card/_index.scss
+++ b/packages/nhsuk-frontend/src/nhsuk/components/card/_index.scss
@@ -253,7 +253,7 @@ $card-border-hover-colour: $nhsuk-card-border-hover-colour;
       background-color: nhsuk-colour("black");
 
       a {
-        @include nhsuk-link-style-white;
+        @include nhsuk-link-style-reverse;
       }
 
       @include nhsuk-print-colour {
--- a/packages/nhsuk-frontend/src/nhsuk/components/hero/_index.scss
+++ b/packages/nhsuk-frontend/src/nhsuk/components/hero/_index.scss
@@ -22,7 +22,7 @@
 
     .nhsuk-link,
     a:not(.nhsuk-button) {
-      @include nhsuk-link-style-white;
+      @include nhsuk-link-style-reverse;
     }
 
     .nhsuk-hero--border {
--- a/packages/nhsuk-frontend/src/nhsuk/core/elements/_links.scss
+++ b/packages/nhsuk-frontend/src/nhsuk/core/elements/_links.scss
@@ -38,7 +38,7 @@ a {
 }
 
 .nhsuk-link--reverse {
-  @include nhsuk-link-style-white;
+  @include nhsuk-link-style-reverse;
 
   @include nhsuk-media-query($media-type: print) {
     &::after {


paulrobertlloyd:
Are the before/after images for the first row the correct way round?

frankieroberto:
> Are the before/after images for the first row the correct way round?

Yep - the only difference in the first row is a 1px offset for the underline.

paulrobertlloyd:
Oh. Is that enough differentiation? Should the underline not get thicker? Should it not be consistent with the other link example shown? (On my phone, so can't test right now, so apologies for perhaps silly questions!)

frankieroberto:
> Oh. Is that enough differentiation? Should the underline not get thicker? Should it not be consistent with the other link example shown? (On my phone, so can't test right now, so apologies for perhaps silly questions!)

Check when you’re back at a laptop! It’s offset by 1px, and gets thicker (from 1px to 3px) on hover.

Basically, the exact same as GOVUK (whose code I copied).

paulrobertlloyd:
🤐

paulrobertlloyd:
- Did you mean to have navigation links in the header underlined? Are there other places this adds underlines where we don’t currently have them?
- It looks as though on GOV.UK the hover state doesn’t add any text decoration gaps between descenders, which makes for a less distracting transition. We should probably do the same.

frankieroberto:
> Did you mean to have navigation links in the header underlined? Are there other places this adds underlines where we don’t currently have them?

I hadn't quite realised how widely (although inconsistently) the link mixin was used. I’ve now added a `$link-underlines` boolean option to all the link mixins, so that we can explicitly specify which components should use underlines for the links (or link-like components) and which ones shouldn't.

For now, I've added/kept underlines to Back links, Breadcrumbs, Cards, Contents list, Details, Error summary, Expander, Footer, Hero, Summary list, Pagination

There's no underlines on: Action link, Header

All up for debate though!

> It looks as though on GOV.UK the hover state doesn’t add any text decoration gaps between descenders, which makes for a less distracting transition. We should probably do the same.

Fixed in 15a8d0d35a179a20db7bc593528c174be37de22b.

packages/nhsuk-frontend/src/nhsuk/core/tools/_links.scss
```diff
@@ -79,16 +91,29 @@
...
-    text-decoration: none;
+
+    @if $link-underlines {
+      text-decoration: underline;
+      text-decoration-thickness: $nhsuk-link-hover-underline-thickness;
+      text-underline-offset: $nhsuk-link-underline-offset;
+
+      // Disable ink skipping on underlines on hover. Browsers haven't
+      // standardised on this part of the spec yet, so set both properties
+      text-decoration-skip-ink: none; // Chromium, Firefox
```
anandamaryon1: I'm not seeing this be applied in browser. (Chrome) 

Agree on GOV that it helps the underline to look less 'busy'. But not sure for us, since our font has lower descenders. Meaning they go slightly below the line, which looks awkward to me. To avoid we'd have to push the offset even more, probably too much?

1. as-is
2. text-decoration-skip-ink: none;
3. ink-skip + offset increased
image

anandamaryon1: Maybe increasing it slightly, with ink-skip, is better actually…

As-is
image

Increased
image

frankieroberto: @anandamaryon1 in the commit above it’s only disabling ink skipping on hover - as with the thicker underline on hover, browsers seem to increase amount of ink skipping, which makes for a bit of a jarring transition. With the thicker underline, ink skipping seems less useful?

Whether we should have ink skipping or not in the regular state, and how much the offset should be, I’m not sure!

anandamaryon1: Ahh, yes it's right there in the comment 😮‍💨 …

It's a moot point on theirs because the offset clears the descenders.

I'll try some options and share around for feedback.

packages/nhsuk-frontend/src/nhsuk/components/back-link/_index.scss
```diff
@@ -48,19 +48,4 @@
...
-
-  &:hover,
-  .nhsuk-back-link--reverse &:hover {
-    text-decoration: underline;
-  }
-}
-
-// stylelint-disable-next-line selector-no-qualifying-type
-button.nhsuk-back-link__link {
-  text-decoration: none;
```
colinrotherham: Love seeing all this go! Exactly as I'd hoped

packages/nhsuk-frontend/src/nhsuk/components/breadcrumb/_index.scss
```diff
@@ -112,16 +112,6 @@
...
-  &:link,
-  .nhsuk-breadcrumb--reverse &:link {
-    text-decoration: none;
-  }
-
-  &:hover,
-  .nhsuk-breadcrumb--reverse &:hover {
-    text-decoration: underline;
-  }
-
```
colinrotherham: 🙌

packages/nhsuk-frontend/src/nhsuk/components/details/_index.scss
```diff
@@ -25,15 +25,11 @@
...
-  color: $color_nhsuk-blue; // [1]
   cursor: pointer;
   display: inline-block; // [2]
   padding-left: nhsuk-spacing(4);
   position: relative; // [3]
-
-  &:hover {
-    color: $nhsuk-link-hover-color;
-  }
+  @include nhsuk-link-style($link-underlines: true);
```
colinrotherham: Can we follow GOV.UK Frontend and add `nhsuk-link-style-no-underline` instead?

You can see their version in:
[govuk-frontend/src/govuk/**helpers/_links.scss**](https://github.com/alphagov/govuk-frontend/blob/v5.10.2/packages/govuk-frontend/src/govuk/helpers/_links.scss#L321-L339)

```suggestion
  @include nhsuk-link-style-default;
  @include nhsuk-link-style-no-underline;
```

<br>

With the [associated class `.nhsuk-link--no-underline`](https://github.com/alphagov/govuk-frontend/blob/v5.10.2/packages/govuk-frontend/src/govuk/core/_links.scss#L27-L29) too

colinrotherham:
Fixed the conflicts I caused (back link text change) 👍

packages/nhsuk-frontend/src/nhsuk/components/details/_index.scss
frankieroberto: Err, could do, I hadn’t spotted that.

Our current `nhsuk-link-style` mixin is set up in quite a different way though, and already has 4 keyword arguments, so this seemed consistent with that. We also need the option to use the `nhsuk-link-style-white` mixin with no underlines.

I found the way the link mixins were structured (and how they include each other) all quite confusing too (and govuk's is confusing too, but in a different way), so open to better ideas!

frankieroberto:
@anandamaryon1 @colinrotherham @paulrobertlloyd @michaelgallagher preview of this branch now working here: https://nhsuk-frontend-pr-1366.herokuapp.com/nhsuk-frontend/

paulrobertlloyd:
Have rebased this with `main`, and removed commits relating to the header, as we’ll be dealing with that separately, at least for the next release.

anandamaryon1:
Reviewed at Design System Review Panel. Generally happy with it, but we just need to iron out how it applies to the header and other components, such as the Action link, to ensure we have a clear rationale to what is underlined and what is not. (currently in this PR everything is underlined, which may be best).

I also notice the expander isn't showing a hover colour. 

In chrome (but not safari) I get this slightly ugly rendering of `g` and `y` descenders. (less so on high-res / retina screen).

Chrome
image

Safari
image

I'll do some more detailed review and explore some options.

colinrotherham:
All conflicts now sorted

> I also notice the expander isn't showing a hover colour.

Well spotted! Fixed

> In chrome (but not safari) I get this slightly ugly rendering of g and y descenders. (less so on high-res / retina screen).

If you have a go at tweaking this Sass variable, work your magic:

```scss
$nhsuk-link-underline-offset: 0.1578em !default;
```

For context on where this number came from:

* https://github.com/alphagov/govuk-frontend/issues/2743

colinrotherham:
It's those short stubby Frutiger descenders isn't it?

You can see at 16px (smaller than 19px body text) it's barely poking through on Safari with the 3px offset

What about a super minor change to ~3px~ 2.5px?

### Safari, 3px offset

Header menu link at 3px offset

### Safari, 2.5px offset

Header menu link at 2.5px offset

<br>
<br>

Makes all the difference at low DPI in Chrome:

### Chrome @1x, 3px offset

Header menu link at 3px offset @1x (Chrome)

### Chrome @1x, 2.5px offset

Header menu link at 2 5px offset @1x (Chrome)

frankieroberto:
I'm gonna merge this into #1367 so that we have a combined PR that both changes the colours and the link underlines.

PR #1368, opened 2025-06-09
Add .slugignore file

frankieroberto:
This adds a `.slugignore` file and includes the reference images for the visual regression tests, so that the preview app can be deployed on Heroku (hopefully).


frankieroberto:
Hmm this didn't work - still over the 500MB limit. Might need another approach. Closing for now.

colinrotherham:
Sorry for re-opening, was just doing some testing

The largest disk usage was from `.cache` where Puppeteer downloads Chrome binaries

PR #1382, opened 2025-06-12
Add accessible name to NHS or organisation logo in header

### LINKED ISSUE #1362: NHS logo missing an accessible name
In the new header (and in the current version too I think), the NHS logo doesn’t have an accessible name.

An accessible label (‘NHS homepage’ by default) is added when the logo is linked, but if there is no link, there is no label.

Things get weirder still when the logo is combined with a service name. I’m not sure what would be read out, as the surrounding link has an `aria-label` (defaulting to ‘NHS homepage’); would this override the inner span containing the service name? 

Either way, as neither ‘NHS homepage’ or ‘NHS homepage service manual’ are correct, so you’d have to make sure you updated the `logo.ariaLabel` param.

Might a better approach be to add an `aria-label` and `<title>` to the NHS logo SVG? I note that this is what GOV.UK do:

```svg
<svg
  focusable="false"
  role="img"
  xmlns="http://www.w3.org/2000/svg"
  viewBox="0 0 296 60"
  height="30"
  width="148"
  fill="currentcolor" class="govuk-header__logotype" aria-label="GOV.UK">
  <title>GOV.UK</title>
  …
</svg>
```

Doing so would mean:
- The logo always gets a name, even if it is not linked
- You wouldn’t need to use `logo.ariaLabel` when combined with a service name, as the correct name would come for free

## Organisation logos

If an organisation adds their own bitmap image, we currently don’t allow any alternative text to be added. This is probably okay given you can override the `aria-label` on the link (though it’s important that you do, as ‘NHS homepage’ is not correct). But is changing the `aria-label` the right approach, versus adding alt text to the image?

## Proposal

- Add `<title>` and `aria-label` to the NHS logo
- Add a param to and alt text to an organisational bitmap logo
- Remove `logo.ariaLabel` param

paulrobertlloyd:
## Description

Fixes #1362.

I’ve kept the `logo.ariaLabel` param and with the same name, but now it defaults to ‘NHS’ instead of ‘NHS homepage’, and is applies to the `aria-label` attribute and `<title>` for the NHS SVG logo, or to the `alt` attribute for an organisation’s bitmap logo image.

## Checklist

- [ ] CHANGELOG entry


packages/nhsuk-frontend/src/nhsuk/components/header/template.njk
```diff
@@ -16,10 +16,12 @@
...
     organisation name (with can be split and include a descriptor) -#}
 {%- macro _serviceLogo(logoSrc, organisation) %}
+{% set ariaLabel = params.logo.ariaLabel | default("NHS") %}
 {% if logoSrc %}
-<img class="nhsuk-header__organisation-logo" src="{{ baseUrl }}{{ logoSrc }}" alt="">
+<img class="nhsuk-header__organisation-logo" src="{{ baseUrl }}{{ logoSrc }}" alt="{{ ariaLabel }}">
 {% else %}
-<svg class="nhsuk-header__logo" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 200 80" height="40" width="100">
+<svg class="nhsuk-header__logo" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 200 80" height="40" width="100" focusable="false" role="img" aria-label="{{ ariaLabel }}">
+  <title>{{ ariaLabel }}</title>
```
colinrotherham, line 24: Loving the [insider knowledge](https://github.com/alphagov/govuk-frontend/pull/4497) that we need both an `aria-label` AND a `<title>` 🙌

paulrobertlloyd:
That was my thought too, but thought best to do that in a separate PR. But can add it as part of this one if it makes sense?

colinrotherham:
I'd do it in this PR so we can test it on the review app?

We've not had separate review app assets before, but see https://github.com/nhsuk/nhsuk-frontend/compare/main...file-copy-review-app for how to do so

Thanks again @paulrobertlloyd

packages/nhsuk-frontend/src/nhsuk/components/header/template.njk
paulrobertlloyd: Also known as ‘copy and paste’ 😉

paulrobertlloyd:
Added an example of an organisational header with a custom logo, and added visual regression tests for this logo at different breakpoints, and when focused. The focused style on this custom logo is bobbins, but this should be fixed by #1361.

paulrobertlloyd:
@colinrotherham Urgh, regression tests failing in CI but passing locally 🤷 Any ideas why?

colinrotherham:
@paulrobertlloyd We had a Puppeteer update the other day

It goes **BackstopJS** → **Playwright** → **Puppeteer** → **Chrome for Testing** binary in `.cache`

You can try `npm ci` followed by what we do on GitHub Actions?

1. Delete browsers cached in `.cache`
2. Run npm script `npm run install:playwright`
3. Run npm script `npm run install:puppeteer`

Then run the tests again

backstop.config.js
```diff
@@ -826,7 +826,24 @@ module.exports = {
...
+        viewports.get('mobile'),
+        viewports.get('tablet'),
+        viewports.get('desktop'),
+        viewports.get('large-desktop')
+      ]
+    },
+    {
+      clickSelector: '.nhsuk-header__service-logo',
+      label:
+        'Header organisational with white header and custom logo (focused logo)',
```
colinrotherham, line 842: If it's easier, we can merge without this and add a suite of "focused logo" tests in https://github.com/nhsuk/nhsuk-frontend/pull/1361?

PR #1396, opened 2025-06-13
Add hover style for header navigation

frankieroberto:
This adds a new hover style for the header navigation, if we were to remove the underlines from the links (see #1366).

It makes the 'active' item have a solid white line at the bottom of it, and the hover style to have a light grey white line. The idea is that by using the same border style we avoid a double-underline, and the light-grey version gives you a hint of what the state will look like when clicked.

I think this might also be slightly more obvious if the line thickness was increased from 4px to 6px, but we previously decided against that?

## Screenshot

Screenshot 2025-06-13 at 11 28 14

### DIFF
--- a/packages/nhsuk-frontend/src/nhsuk/components/header/_index.scss
+++ b/packages/nhsuk-frontend/src/nhsuk/components/header/_index.scss
@@ -5,6 +5,8 @@
 
 $_header-item-padding: 12px;
 $_header-link-padding: 2px;
+$_navigation-current-size: 5px;
+$_navigation-hover-size: $nhsuk-link-hover-underline-thickness;
 
 ////
 /// Header component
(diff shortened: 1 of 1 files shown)


paulrobertlloyd:
In 740ebd1773cda3b3b37d5f77af601ba4be433bfc I’ve taken what @frankieroberto started with, and made a few adjustments:

- Rather than needing to constantly redefine or remove `box-shadow`, and applying `box-shadow` to `nhsuk-header__navigation-link` when horizontal but `nhsuk-header__navigation-item` in the overflow menu… I’ve added our friend `::before`, which is absolutely positioned within `nhsuk-header__navigation-link`.
  - For an active or current item, its height is set to `$_navigation-current-size`
  - Upon hover, its height is set to `$_navigation-hover-size`
  - For the overflow menu, we instead manage its width, as it is positioned along the left edge, not the bottom
  - These are both currently the same value (4px), but feel free to have a play with these (I’ve tried multiple variations, between having 6px for selected or hover, but landed on this)
  - This pseudo element uses `currentColor`, and inherits the link hover state (so what is now purple may become dark blue in a later PR)
  - Overall, there is less CSS and less trying to manage these different states and variations
- ~~I’ve then added a subtle transition, so that upon hover the size can grow and shrink~~
- Focus lives in it’s own world, so is not effected by the pseudo element

paulrobertlloyd:
Removed the transition, made hover size 3px, current size 6px.

paulrobertlloyd:
Actually, scratch that, how about:

```scss
$_navigation-current-size: 5px; // Midway between 4px and 6px
$_navigation-hover-size: $nhsuk-link-hover-underline-thickness; // Seems should use this?
```

anandamaryon1:
Not sure I love the differing sizes between hover and current. Seems a bit fussy. 

How about keeping the size and using a lighter tint of the blue? Works for blue and white background which is a bonus. Also tweaked size to 6px and mad the current match page-background colour. 

Header Navigation Screenshot

Header Org White Nav

Side note, the no-visited-state doesn't seem to be working on the white header (click a link and they all go dark), had a play but the mixins obscure it so I couldn't work out how to fix it.

paulrobertlloyd:
> How about keeping the size and using a lighter tint of the blue? Works for blue and white background which is a bonus. Also tweaked size to 6px and mad the current match page-background colour.

I like this! Just to confirm, is the light blue for the hover state (that'd be my preference)? 6px doesn't look so bad here now either… you may have convinced me! 😆

anandamaryon1:
Sorry yes, to confirm, hover is light blue, current is page-background or nhs blue.

frankieroberto:
@anandamaryon1 the blue tint for hover looks good to me!

Any thoughts on the width of the line at the bottom? 4px, 5px or 6px? 😃

paulrobertlloyd:
> Any thoughts on the width of the line at the bottom?

I thought @anandamaryon1 was suggesting 6px. Worth nothing that the focused state will use 4px.

frankieroberto:
@paulrobertlloyd could tweak the focused border to use 6px in this context too?

paulrobertlloyd:
You’re talking me back into 4px again ;-)

(I saw 6px on my phone and it looked fine, but now, when I see 6px on desktop, the vertical spacing between the top and the bottom of a current/hovered item looks uneven with 6px).

Issue #1390, opened 2025-06-13
Add Gulp build output for Nunjucks templates, macro options and fixtures


colinrotherham:
We should ensure all remaining static source files are included in our build output

For example, each component has:

* README files
* Nunjucks files
* Macro options

See existing approach in GOV.UK Frontend via:

* [govuk-frontend/**tasks/fixtures.mjs**](https://github.com/alphagov/govuk-frontend/blob/main/packages/govuk-frontend/tasks/fixtures.mjs)
* [govuk-frontend/**tasks/templates.mjs**](https://github.com/alphagov/govuk-frontend/blob/main/packages/govuk-frontend/tasks/templates.mjs)

## Macro options format

We should process component `macro-options.json` files from another source format, ideally so we can add examples (and their predefined macro options) in future.

I've added some examples for comparison here:
https://gist.github.com/colinrotherham/cbc2fd7e0f0d672f04b3268315fc5c11

* **macro-options.json**
* **macro-options.mjs**
* **macro-options.yaml**

NHS.UK frontend currently use `.json` for every component.

But GOV.UK Frontend uses `.yaml` and converts to `.json` via **js-yaml** in a build task.

After a discussion with @frankieroberto and @paulrobertlloyd we're leaning towards `.mjs` as it can be exported to `.json` at build time (and published into the npm package) and continue to be imported into the review app.

JavaScript can additionally be checked by ESLint and code autocomplete works in editors:

```mjs
import { examples } from './macro-options.mjs'

// Code autocomplete works

examples[0]
examples[0].options
examples[0].options.text
```

colinrotherham:
Closed by https://github.com/nhsuk/nhsuk-frontend/pull/1402 and https://github.com/nhsuk/nhsuk-frontend/pull/1403

PR #1402, opened 2025-06-16
Add Gulp build output for macro options JSON

colinrotherham:
## Description

This PR includes changes to:

1. Move all `macro-options.json` files into ES module format
2. Add helpers `getListing()` and `getDirectories()` from GOV.UK Frontend
3. Add new Gulp task `gulp fixtures` to build `macro-options.json` files

These changes complete 50% of issue https://github.com/nhsuk/nhsuk-frontend/issues/1390

I'll be adding `fixtures.json` from GOV.UK Frontend in another PR so we can share component examples

## Checklist

- [ ] CHANGELOG entry

### DIFF
--- a/shared/lib/components.mjs
+++ b/shared/lib/components.mjs
@@ -50,7 +50,7 @@ export async function getNames() {
 }
 
 /**
- * Convert macro option params to JSON format
+ * Convert macro option params to array of options
  *
  * @param {{ [param: string]: MacroParam }} [params] - Nunjucks macro option params
  * @returns {MacroOption[] | undefined} Nunjucks macro options


shared/lib/components.mjs
```diff
@@ -24,6 +24,40 @@ export function nunjucksEnv(searchPaths = [], nunjucksOptions = {}) {
...
+ * @returns {MacroOption[] | undefined} Nunjucks macro options
+ */
+export function getComponentMacroOptions(params) {
+  if (!params) {
+    return
+  }
+
+  // Format nested params
+  return Object.entries(params).map(([name, param]) => {
+    const option = /** @type {MacroOption} */ ({
```
MatMoore, line 65: I'm not understanding why we have both `options` and `params` in each of the component files. Is `options` just here for backwards compatibility while we migrate to `params`?

shared/lib/components.mjs
```diff
@@ -24,6 +24,40 @@ export function nunjucksEnv(searchPaths = [], nunjucksOptions = {}) {
   })
 }
 
+/**
+ * Convert macro option params to JSON format
```
MatMoore: "JSON format" here is a bit confusing - it makes it sounds like its generating a string, rather than an array. The actual JSON stringify happens elsewhere

shared/lib/components.mjs
colinrotherham: The exported `params` is an object that (one day) can be used to [validate a component's config via schema](https://github.com/alphagov/govuk-frontend/blob/v5.10.2/packages/govuk-frontend/src/govuk/components/character-count/character-count.mjs#L407-L467)

But the exported `options` is just the params in an array instead

Yeah might be good to settle on the params syntax one day, but for now it's hanging around for backwards compatibility. For example, so we can [import them into the service manual website](https://github.com/nhsuk/nhsuk-service-manual/blob/fb0c3aaa00b43a02021ec6a8f422680fbf96e547/lib/file-helper.js#L42-L64)

shared/lib/components.mjs
colinrotherham: Changed it, thanks @MatMoore

PR #1403, opened 2025-06-16
Add Gulp build output for component fixtures JSON 

colinrotherham:
## Description

This PR updates all component `macro-options.mjs` files to include all examples

These are built to `fixtures.json` and we'll automatically use them to render the review app examples in future.

By copying GOV.UK Frontend, we can verify HTML output in forks like [nhsuk-frontend-jinja](https://github.com/NHSDigital/nhsuk-frontend-jinja)

Following https://github.com/nhsuk/nhsuk-frontend/pull/1402, this nearly completes https://github.com/nhsuk/nhsuk-frontend/issues/1390

### One example per component

To support future review app changes, each example now only renders one component.

Complex examples with multiple or duplicate components have moved:

Example pages

## Checklist

- [ ] CHANGELOG entry

### DIFF
--- a/shared/lib/components.mjs
+++ b/shared/lib/components.mjs
@@ -31,7 +31,7 @@ export function nunjucksEnv(searchPaths = [], nunjucksOptions = {}) {
  * @param {string} component - Component directory name
  * @returns {Promise<ComponentData>} Component data
  */
-export async function getData(component) {
+export async function load(component) {
   const optionsPath = join(
     paths.pkg,
     `src/nhsuk/components/${component}/macro-options.mjs`
@@ -69,7 +69,7 @@ export async function getData(component) {
 /**
  * Load all component data (from source)
  */
-export async function getDataList() {
+export async function loadAll() {
   const listing = await getDirectories('nhsuk/components', {
     cwd: join(paths.pkg, 'src')
   })
@@ -80,7 +80,7 @@ export async function getDataList() {
     .sort()
 
   // Load component data per directory
-  return Promise.all(components.map(getData))
+  return Promise.all(components.map(load))
 }
 
 /**
--- a/shared/tasks/components.mjs
+++ b/shared/tasks/components.mjs
@@ -10,7 +10,7 @@ import { files } from './index.mjs'
  * @param {Pick<AssetOptions, "destPath">} options - Asset options
  */
 export async function generateFixtures({ destPath }) {
-  const list = await components.getDataList()
+  const list = await components.loadAll()
 
   // Loop component names
   const fixtures = list.map(async (data) => {
@@ -39,7 +39,7 @@ export async function generateFixtures({ destPath }) {
  * @param {Pick<AssetOptions, "destPath">} options - Asset options
  */
 export async function generateMacroOptions({ destPath }) {
-  const list = await components.getDataList()
+  const list = await components.loadAll()
 
   // Loop component names
   const macroOptions = list.map(async (data) => {


shared/lib/components.mjs
```diff
@@ -28,13 +28,39 @@ export function nunjucksEnv(searchPaths = [], nunjucksOptions = {}) {
...
+  )
+
+  // Add component directory name to options
+  return { component, ...options }
+}
+
+/**
+ * Load all component data (from source)
+ */
+export async function getDataList() {
```
MatMoore: [very minor] could these two function names be more descriptive? Perhaps `loadComponent` / `loadComponentList` ?

MatMoore (approved):
🎉  nice job.

I was wondering - will these fixtures include the HTML equivalent of the examples? (In the service manual we render both in the docs side by side.)

If this is available via fixtures, then I will write the tests in nhsuk-frontend-jinja to just take the nunchucks example, render it through Jinja, and compare to the html. But if the html is not easily accessible, then there will be an extra step of running it through nunchucks as well.

shared/lib/components.mjs
colinrotherham: Glad you pointed this out, I did struggle with these names

The API is namespaced so it'll need to work well as:

```js
components.getDataList()
components.getData('header')
```

How about:

```js
components.loadAll()
components.load('header')
```

colinrotherham:
Thanks @MatMoore

>I was wondering - will these fixtures include the HTML equivalent of the examples? (In the service manual we render both in the docs side by side.)

Yeah there's a `html` key in **fixtures.json**, same format used by GOV.UK Frontend

I imagine you might hit differences in empty space (trailing, indentation, new lines) so I wasn't sure if we should run the output through a formatter like Prettier but perhaps have a look after we publish the prerelease?

You'll see **fixtures.json** output like the following for action links:

## **dist/nhsuk/components/action-link/fixtures.json**
```json
{
  "name": "Action link",
  "component": "action-link",
  "fixtures": [
    {
      "name": "default",
      "description": "",
      "options": {
        "text": "Find your nearest A&E",
        "href": "#"
      },
      "html": "<div class=\"nhsuk-action-link\">\n  <a class=\"nhsuk-action-link__link\" href=\"#\">\n    <svg class=\"nhsuk-icon nhsuk-icon__arrow-right-circle\" xmlns=\"http://www.w3.org/2000/svg\" viewBox=\"0 0 24 24\" aria-hidden=\"true\" width=\"36\" height=\"36\">\n      <path d=\"M0 0h24v24H0z\" fill=\"none\"></path>\n      <path d=\"M12 2a10 10 0 0 0-9.95 9h11.64L9.74 7.05a1 1 0 0 1 1.41-1.41l5.66 5.65a1 1 0 0 1 0 1.42l-5.66 5.65a1 1 0 0 1-1.41 0 1 1 0 0 1 0-1.41L13.69 13H2.05A10 10 0 1 0 12 2z\"></path>\n    </svg>\n    <span class=\"nhsuk-action-link__text\">Find your nearest A&amp;E</span>\n  </a>\n</div>"
    }
  ]
}
```

MatMoore:
@colinrotherham perfect 👍🏻

I was planning to fix all the whitespace in the Jinja templates to match the fixtures as part of introducing the tests (as long as you're happy the whitespace is correct here).

colinrotherham:
@MatMoore Yeah all correct and comes from Nunjucks itself

Beyond the defaults, the [Nunjucks environment is configured](https://mozilla.github.io/nunjucks/api.html#configure) with:

```
trimBlocks: true,
lstripBlocks: true,
```

With a final `.trim()` applied to the output

Both to match GOV.UK Frontend

PR #1411, opened 2025-06-17
Reinstate space between card headings and custom HTML content

### LINKED ISSUE #1410: Change in v9.6.2 removed space between card headings and custom HTML content
## What is the issue?

[This commit](https://github.com/nhsuk/nhsuk-frontend/commit/faf56304c90ff4d6097f7cd65955742a52a9549c) removed the bottom margin from `nhsuk-card__heading`, and added a top margin to `nhsuk-card__description`.

This works in the cases where `nhsuk-card__description` is used (i.e. via the `description` param in the Nunjucks macro), but if another class is used (i.e. via `descriptionHtml` param or in the `caller()`), the space between the heading and following content disappears:

Image 1

Image 2

Image 3

NHS.UK frontend package version: 9.6.2

paulrobertlloyd:
## Description

Fixes #1410.

I’ve added an example and updated the backstop config, but not added any visual regression images as I can’t seem to get the `npm run test:visual` task to run anymore (it times out).

## Checklist

- [ ] CHANGELOG entry


packages/nhsuk-frontend/src/nhsuk/components/card/_index.scss
```diff
@@ -57,15 +57,11 @@ $card-border-hover-color: $color_nhsuk-grey-3;
   @include nhsuk-responsive-padding(5);
 }
 
-.nhsuk-card__heading,
+.nhsuk-card__heading:has(+ .nhsuk-icon),
```
paulrobertlloyd, line 60: Only browsers that support `:has()` will remove the trailing margin below orphaned headings on primary cards… I think we can live with that, no?

packages/nhsuk-frontend/src/nhsuk/components/card/_index.scss
```diff
@@ -57,13 +57,16 @@ $card-border-hover-color: $color_nhsuk-grey-3;
   @include nhsuk-responsive-padding(5);
 }
 
-.nhsuk-card__heading,
-.nhsuk-card__description {
-  margin-bottom: 0;
+.nhsuk-card__heading {
+  margin-bottom: nhsuk-spacing(3);
+
+  &:has(+ .nhsuk-icon) {
```
paulrobertlloyd, line 63: Only browsers that support `:has()` will remove the trailing margin below orphaned headings on primary cards… I think we can live with that, no?

anandamaryon1: I'd class this as a progressive enhancement. When this doesn't apply, it'll just give a margin bottom to icon cards, which won't affect the component's usability. 

P.S: TIL we have parent selectors now. How'd I miss this!?

anandamaryon1:
Think this needs a changelog? I'm thinking not worthy but happy to be swayed?

PR #1415, opened 2025-06-19
Don't render heading if one isn't provided

edwardhorsford:
## Description
The current card component will output heading markup even when a heading isn't provided - which also adds excess whitespace.

A user might want to intentionally not provide a heading if they are including it within the `descriptionHtml`. In my service I am going to have several summary lists within the card, each with a heading.

The hacky way to avoid the rendering is to pass `headingHtml: " "` - but this doesn't feel ideal.


paulrobertlloyd:
Several summary lists within a card… sounds like a summary card (which we should add at some point).

colinrotherham:
I accidentally did this in https://github.com/nhsuk/nhsuk-frontend/pull/1684 when splitting out the `_cardHeading()` macro

PR #1428, opened 2025-06-23
Header final code refinements

paulrobertlloyd:
## Description

This PR cherry-picks improvements added in #1419 not tied to hover states. It does have one slight design change; the current item indicator is now the same colour as the text label (so white on blue backgrounds, blue on white backgrounds).

## Checklist

- [ ] CHANGELOG entry


colinrotherham:
I've replaced all the header reference images, see if that's any better

Issue #1354, opened 2025-06-03
Update documentation with new import paths


colinrotherham:
Part of https://github.com/nhsuk/nhsuk-frontend/issues/1237

PR #1436, opened 2025-06-24
Change default JavaScript bundle to ES module format

colinrotherham:
## Description

This change was missed from https://github.com/nhsuk/nhsuk-frontend/pull/1290

It enables ES module imports from our minified JavaScript bundle, for example:

```html
<script type="module">
  import { initAll } from '/javascripts/nhsuk-frontend-10.0.0.min.js'
  initAll()
</script>
```

Changelog entry will be picked up in https://github.com/nhsuk/nhsuk-frontend/issues/1354

## Checklist

- [ ] CHANGELOG entry

### DIFF
--- a/docs/installation/installing-with-npm.md
+++ b/docs/installation/installing-with-npm.md
@@ -112,10 +112,10 @@ initAll()
 Rather than using `initAll`, you can initialise all instances of individual components used by your service. For example:
 
 ```js
-import { initButton, initRadios } from 'nhsuk-frontend';
+import { initButtons, initRadios } from 'nhsuk-frontend';
 
 // Initialise all button components
-initButton();
+initButtons();
 
 // Initialise all radios components
 initRadios();
--- a/packages/nhsuk-frontend/src/nhsuk/components/button/button.jsdom.test.mjs
+++ b/packages/nhsuk-frontend/src/nhsuk/components/button/button.jsdom.test.mjs
@@ -1,7 +1,7 @@
 import { components } from '@nhsuk/frontend-lib'
 import { getByRole } from '@testing-library/dom'
 
-import { initButton } from './button.mjs'
+import { initButtons } from './button.mjs'
 
 describe('Button', () => {
   /** @type {HTMLButtonElement} */
@@ -19,7 +19,7 @@ describe('Button', () => {
 
   describe('Initialisation', () => {
     it('should add event listeners', () => {
-      initButton()
+      initButtons()
 
       expect($button.addEventListener).toHaveBeenNthCalledWith(
         1,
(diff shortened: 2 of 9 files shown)


docs/installation/installing-with-npm.md
```diff
@@ -83,12 +83,16 @@ Add the following JavaScript to the top of the `<body>` section of your page tem
 
 Copy the `/node_modules/nhsuk-frontend/dist/nhsuk/nhsuk-frontend.min.js` script file into your application.
 
-Include the script in the `<head>` of your page using the `type="module"` attribute.
+Then include the script before the closing `</body>` tag of your page using the `type="module"` attribute, and run the `initAll` function to initialise all the components.
```
dnimmo, line 86: ❤️ for moving this to the end of the body

docs/installation/installing-with-npm.md
```diff
@@ -105,17 +109,31 @@ initAll()
...
-import { initSkipLink } from 'nhsuk-frontend/src/nhsuk/components/skip-link/skip-link.mjs';
-
-// Initialise components
-document.addEventListener('DOMContentLoaded', () => {
-  initRadios();
-  initSkipLink();
-});
+import { initButton, initRadios } from 'nhsuk-frontend';
+
+// Initialise all button components
```
dnimmo, line 117: Apologies for the lack of context I have here, so feel free to disregard, but: 

`initButton()` (singular) initialises all buttons, where 
`initRadios()` (pllural) is needed for initialising all radios

How would you feel about updating initButton to be plural? Sorry if this feels super pedantic but feels like a potential source of friction/confusion.

colinrotherham: You're spot on. These were previously default exports so we've got 100% flexibility to change them (thank you)

I did wonder if every component should just `export init`?

Then if you want to rename it, that's up to you

dnimmo: If initialising is necessary and expected per component then yeah tbh, consistency there probably does make more sense. I have to say that personally I don't love named imports like this anyway though - totally a personal preference thing, but for me, I find it a lot easier to parse (especially in longer files with multiple things being imported): 

```
import stuff from 'somewhere';

stuff.doSomething()
```

When you're far enough down a file that you can't see the imports, I find it helps to know where things actually come from - especially with more generic names (like "init"). 

Totally subjective though, of course! Not a hill I'd die on, just felt worth mentioning. :D

colinrotherham: Yeah, appreciate that

It's an NHS 'ism, that historically each component doesn't export itself—it exports an init script

Unfortunately there's no way for us to mix default and named exports without adding a [synthetic `default` and `__esModule` property](https://rollupjs.org/configuration-options/#output-esmodule) but we'd lose `require()` and `import` cross compatibility without Babel etc

We are considering dropping init scripts in v11 once we align more with GOV.UK Frontend

With the above in mind, what do you reckon?

dnimmo: No worries, yeah agree that cross compatibility is more important, and yeah assumed that it was an NHS convention re: inits - all good! 

Yeah with that in mind, given that there's multiple inits from one place, I guess it makes sense just to stick with what's already there, but change `importButton` to `importButtons`.

colinrotherham: Thanks @dnimmo

That's now updated in https://github.com/nhsuk/nhsuk-frontend/pull/1436/commits/0e4a4d7702db09f50b18f73e14f6bda0fb6ccab8 and I've updated skip links to come along for the ride too (multiple are supported on GOV.UK Frontend, unlike here in v9.x)

Over in the documentation PR I've [updated that too](https://github.com/nhsuk/nhsuk-frontend/pull/1437/files/47d08630b5c596da46ac75761b50b021af0c1015..0bcd01160b3a49cab2088bde7094de32bbe81dde)

MatMoore (approved):

PR #1437, opened 2025-06-24
Update documentation with new import paths

colinrotherham:
## Description

This PR closes https://github.com/nhsuk/nhsuk-frontend/issues/1354

## Checklist

- [x] CHANGELOG entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -224,13 +224,13 @@ Page templates now include a new `nhsuk-frontend-supported` class on the `<body>
 
 If you are not using our Nunjucks page template, replace the existing snippet:
 
-```js
+```html
 <script>document.body.className = ((document.body.className) ? document.body.className + ' js-enabled' : 'js-enabled');</script>
 ```
 
 with:
 
-```js
+```html
 <script>document.body.className += ' js-enabled' + ('noModule' in HTMLScriptElement.prototype ? ' nhsuk-frontend-supported' : '');</script>
 ```
 


CHANGELOG.md
```diff
@@ -20,25 +20,191 @@ We’ve updated the header component to support account information and links. A
 
 This was added in [pull request #1058: New header with account section](https://github.com/nhsuk/nhsuk-frontend/pull/1058).
 
+#### Create custom width container classes
```
frankieroberto, line 35: Feels like this section should go below the breaking changes maybe?

CHANGELOG.md
```diff
@@ -20,25 +20,191 @@ We’ve updated the header component to support account information and links. A
...
+
+```scss
+@forward "node_modules/nhsuk-frontend/packages/core";
+```
+
+After:
+
+```scss
+@forward "nhsuk-frontend/dist/nhsuk/core";
+```
```
frankieroberto: Might be better to use the `@import` examples for now? Also maybe drop `/core` from the examples as I think it’s more common to import the whole thing?

CHANGELOG.md
```diff
@@ -20,25 +20,191 @@ We’ve updated the header component to support account information and links. A
...
+const $checkboxes = document.querySelector('.app-checkboxes')
+
+// Initialise single button component
+new Button($button);
+
+// Initialise single checkboxes component
+new Checkboxes($checkboxes);
+```
+
+For precompiled JavaScript, note the following path changes:
```
frankieroberto: Swap the order of these so that it comes before the bundler instructions maybe? (I suspect using the precompiled JavaScript is way more common?)

CHANGELOG.md
```diff
@@ -20,25 +20,191 @@ We’ve updated the header component to support account information and links. A
...
+  'node_modules/nhsuk-frontend/dist/nhsuk/components',
+  'node_modules/nhsuk-frontend/dist/nhsuk/macros',
+  'node_modules/nhsuk-frontend/dist/nhsuk',
+  'node_modules/nhsuk-frontend/dist'
+])
+```
+
+##### If you’re copying or serving assets
+
+Replace `packages/` with `dist/nhsuk` when copying or serving NHS.UK frontend assets:
```
frankieroberto: Is it worth mention that this section also applies to serving static images like logos and favicons? Nearly all services should be doing that.

CHANGELOG.md
anandamaryon1: It's less important but I think we should follow the precedent of having new features first, then changes after.

CHANGELOG.md
anandamaryon1, line 1: > This release stops Internet Explorer 11 and other older browsers from running NHS.UK frontend JavaScript. Your service will not stop working in Internet Explorer 11, but components will look and behave differently without JavaScript.

Perhaps this intro can be extended to give a brief overview of the main changes. 
* IE11 support
* New header
* Path changes

anandamaryon1: Something like:

> This release introduces some breaking changes to file paths, the header component and others. It also stops Internet Explorer 11 and other older browsers from running NHS.UK frontend JavaScript. 
> 
> Read these updates carefully to check what you need to change to ensure your service doesn't break. 

Thoughts @colinrotherham?

colinrotherham: Sounds good, can you have a look at https://github.com/nhsuk/nhsuk-frontend/pull/1437/commits/8dc700e5dba5a485dec99b2f8fe1d376867a6328 ?

CHANGELOG.md
colinrotherham: Definitely, added in https://github.com/nhsuk/nhsuk-frontend/pull/1437/commits/1a423374c29d6b1ff8b8ae7fdc25cc85a4db84b4

CHANGELOG.md
colinrotherham: Makes sense, have a look at https://github.com/nhsuk/nhsuk-frontend/pull/1437/commits/58fa9cc0d1887a7980a9ea2d1f2a0dcec8bdd660

CHANGELOG.md
colinrotherham: I've made Sass tweaks in https://github.com/nhsuk/nhsuk-frontend/pull/1437/commits/5283b98851394886c3edce71b51abb55bfb3631b

frankieroberto (approved):

PR #1440, opened 2025-06-26
Fix path to cached Chrome browser

colinrotherham:
## Description

Looks like Playwright dropped support for `chromePath` so it’s supposed to be `executablePath`

If you've ever had the Playwright CLI install `chromium` in the past, this PR resolves an issue where BackstopJS would use that instead of [Chrome for Testing](https://developer.chrome.com/blog/chrome-for-testing) (which we download into `./cache`)

## Checklist

- [ ] CHANGELOG entry


anandamaryon1:
Cleared `.cache` and reinstalled browsers. Seems to work for me, ran backstop with no fails.

PR #1449, opened 2025-06-30
Fix a visual bug with checkboxes inside a details component

### LINKED ISSUE #929: Ticks on checkboxes inside details element not rendering correctly
# Bug Report

## What is the issue?

If you put a checkbox inside a details element, the tick is too large and not centred correctly.

image

Combine the details and checkbox examples from the design system as per this codepen:

https://codepen.io/andymantell/pen/KKYdjxE

* Device: Windows PC
* Operating System: Windows 11
* Browser: Google Chrome
* Browser version: 122
*
* NHS.UK frontend package version: 8.1.0

frankieroberto:
Fixes #929

## Checklist

- [x] CHANGELOG entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -82,6 +82,7 @@ This change was introduced in pull requests [#1459: Add NHS.UK frontend browser 
 - [#1479: Fix tabs component JavaScript to use `$nhsuk-breakpoints` option](https://github.com/nhsuk/nhsuk-frontend/pull/1479)
 - [#1487: Fix character count not having error border colour when rendered with error message](https://github.com/nhsuk/nhsuk-frontend/pull/1487)
 - [#1489: Apply conditional reveal fixes from GOV.UK Frontend](https://github.com/nhsuk/nhsuk-frontend/pull/1489)
+- [#1449: Fix a visual bug with checkboxes inside a details component](https://github.com/nhsuk/nhsuk-frontend/pull/1449)
 
 ## 10.0.0-internal.0 - 2 July 2025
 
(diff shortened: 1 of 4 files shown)


frankieroberto:
I've added `box-sizing: border-box;` to Checkbox inputs only, but it could be added to `.nhsuk-label` instead, as @andymantell and @Fenwick17 suggested? 🤷

colinrotherham:
Try a search for `box-sizing: border-box` on GOV.UK Frontend for comparison? 😮

frankieroberto:
> Try a search for `box-sizing: border-box` on GOV.UK Frontend for comparison? 😮

Arguably we could add it to everything like this:

```
* {
  box-sizing: border-box;
}
```

I often do that on my own sites.

For now, I think this PR is ok as-is? I've added an example page featuring lots of different components within Details, so that we can tell if this accidentally regresses in future. At the moment, it looks like checkboxes are the only one with a bug.

colinrotherham (approved):

frankieroberto:
@colinrotherham one thing I couldn't quickly work out was how to add this to the visual regression tests (if you think we should)?

colinrotherham:
>@colinrotherham one thing I couldn't quickly work out was how to add this to the visual regression tests (if you think we should)?

Have a look at the [fieldset **macro-options.mjs** 'with inputs' example](https://github.com/nhsuk/nhsuk-frontend/blob/cf46d1c55dfba994769fc312a8c42bd7c30be736/packages/nhsuk-frontend/src/nhsuk/components/fieldset/macro-options.mjs#L94-L147) where `callBlock` is used?

Can do the same with details and use `callBlock` to render child components inside

frankieroberto:
Hmm actually I'm not sure we need to do visual regression testing on this. Those tests are slow enough as it is.

PR #1452, opened 2025-07-01
Footer params

paulrobertlloyd:
## Description

- Update footer component to use `href` and `text` for link parameters instead of `URL` and `label` (matching parameter names for similar links in other components)
- Tidy up footer macro markup
- Refactor template markup to be clearer and simpler

No visual design changes, except for:

- Ensure the copy right statement is always the last item and left aligned (accounts for long copyright statements and produces neater layout)
- Use a dark grey top border for meta items below desktop breakpoint
- Minor spacing adjustments

This is currently a breaking change, but we could easily make it not so by adding fallbacks for `URL` and `label` in the single macro this component uses to generate link lists.

## Checklist

- [ ] CHANGELOG entry


paulrobertlloyd:
Have added more component examples, and changed the default example to only show the copyright statement, with separate examples added with different configurations of links.

packages/nhsuk-frontend/src/nhsuk/components/footer/template.njk
```diff
@@ -1,81 +1,33 @@
...
-        <p class="nhsuk-footer__copyright">{{ copyrightText }}</p>
-      </div>
-      {% endif %}
-    </div> {# close nhsuk-width-container #}
-  </div> {# close nhsuk-footer-container #}
+{% macro _footerList(items, modifier) %}
+<ul class="nhsuk-footer__list{% if modifier %} nhsuk-footer__list--{{ modifier }}{% endif %}">
+{% for item in items %}
+  <li class="nhsuk-footer__list-item">
+    <a class="nhsuk-footer__list-item-link" href="{{ item.href }}">{{ item.text }}</a>
```
frankieroberto: We could support `html` content too?

```suggestion
    <a class="nhsuk-footer__list-item-link" href="{{ item.href }}">{{ item.html | safe if item.html else item.text }}</a>
```

packages/nhsuk-frontend/src/nhsuk/components/footer/template.njk
```diff
@@ -1,81 +1,33 @@
...
+  </li>
+{% endfor %}
+</ul>
+{% endmacro %}
+
+<footer class="nhsuk-footer
+  {%- if params.classes %} {{ params.classes }}{% endif %}" role="contentinfo"
+  {{- nhsukAttributes(params.attributes) }}>
+  <div class="nhsuk-width-container {%- if params.containerClasses %} {{ params.containerClasses }}{% endif %}">
+    <h2 class="nhsuk-u-visually-hidden">Support links</h2>
```
frankieroberto: Allow this to be set too?

```suggestion
    <h2 class="nhsuk-u-visually-hidden">{{ params.visuallyHiddenTitle | default("Support links") }}</h2>
```

packages/nhsuk-frontend/src/nhsuk/components/footer/template.njk
```diff
@@ -1,81 +1,33 @@
...
-            {%- endif %}
-      </div> {# close nhsuk-footer #}
-      <div>
-        <p class="nhsuk-footer__copyright">{{ copyrightText }}</p>
-      </div>
-      {% endif %}
-    </div> {# close nhsuk-width-container #}
-  </div> {# close nhsuk-footer-container #}
+{% macro _footerList(items, modifier) %}
+<ul class="nhsuk-footer__list{% if modifier %} nhsuk-footer__list--{{ modifier }}{% endif %}">
```
frankieroberto: Worth allowing each list to have an optional title (heading) maybe?

paulrobertlloyd:
With @anandamaryon1’s encouragement, and taking on feedback from @frankieroberto and @colinrotherham, I have significantly refactored the footer component so that it is more flexible, as well as being consistent with patterns we use for other components. Here is a rundown of the changes:

## Split links into clearly defined `meta` and `navigation` areas

Inspired by, but by no means replicating how the respective component works on GOV.UK (which has different needs and is far more complex, possibly needlessly so), the previous `links`, `linksColumn2`, `linksColumn3` and `metaLinks` params have been replaced by `meta` and `navigation` params.

The `navigation` param broadly replaces `links`, but allows for multiple (infinite) navigation lists which are displayed as one-quarter grid columns. Each list can optionally display a `title`.

The `meta` param allows authors to show a separate group of links which, above the desktop breakpoint appear inline. This group of links appears below any navigation lists and has a `visuallyHiddenTitle`, that defaults to `"Support links"`.

The `meta` area can additionally take `text` or `html` which appears below the `meta` links and above the copyright statement. This can be used to display extra information, such as a content license, or any bespoke HTML (perhaps this should be the component’s `caller()`?)

Finally, when the footer contains navigation, a grey border is added between this and the meta area below it.

This is possibly all best understood by looking at the updated visual regression images, and the updated macro options.

## Copyright

The copyright param is now an object, which can accept either `text` or `html` (`html` takes precedence otherwise `text`, which defaults to `"© NHS England"`).

## Links

Finally, all footer links can take `text` or `html` values (with `html` taking precedence), and optionally any `attributes`.

colinrotherham:
Rebased and fixed the reference images

Mind checking the [fix for the `meta` params](https://github.com/nhsuk/nhsuk-frontend/pull/1452/commits/cc1f09ed7a720b97598155553d1fa6ed2fb79784)?

packages/nhsuk-frontend/src/nhsuk/components/footer/template.njk
```diff
@@ -1,81 +1,53 @@
...
+{% if params.meta %}
+    {% if params.meta.items %}
+      <h2 class="nhsuk-u-visually-hidden">{{ params.meta.visuallyHiddenTitle | default("Support links") }}</h2>
+      {{ _footerList(params.meta.items) | indent(6) }}
+    {% endif %}
+    {% if params.meta.text %}
+      <p class="nhsuk-body-s">{{ params.meta.text }}</p>
+    {% elif params.meta.html %}
+      {{ params.meta.html | safe }}
+    {% endif %}
```
colinrotherham: Thanks for all this @paulrobertlloyd, looks ace

Mind if we flip this bit so `meta.html` takes precedence over `meta.text`?

Nunjucks caller idea sounds great, maybe something like?

```suggestion
    {% if caller or params.meta.html %}
      {{ params.meta.html | safe }}
      {{ caller() if caller else params.meta.html | safe | trim | indent(6) }}
    {% elif params.meta.text %}
      <p class="nhsuk-body-s">{{ params.meta.text }}</p>
    {% endif %}
```

paulrobertlloyd:
@colinrotherham Updated the meta area in ec83459 to ensure `meta.html` takes precedence, and can now take the component’s `caller()`.

frankieroberto:
@paulrobertlloyd nice, love this! 💙

A couple of questions:

* as I understand it, if you have 3 sets of navigation links plus meta links, they just appear as 4 columns on desktop with no visual distinction? I wonder if it’d make sense for meta links to always appear _under_ the navigation links (maybe with another line to separate), and perhaps still appearing horizontally rather than vertically on desktop? ie more like [GOV.UK footer with links and secondary navigation](https://design-system.service.gov.uk/components/footer/full/). It would be a bigger change to what we have now though.
* meta links have a visually-hidden title whereas navigation links have a visible title - should we allow both options for both (maybe `{title: "", visuallyHidden: true/false}`) or are there good reasons for the difference?

Other than that, just needs a changelog entry I think?

paulrobertlloyd:
On your first point, that’s how it now works – see https://nhsuk-frontend-pr-1452.herokuapp.com/nhsuk-frontend/components/footer/with-meta-and-navigation/

Om the second, maybe… I feel the difference is a useful point of differentiation.

frankieroberto:
@paulrobertlloyd aha, nice, I'd missed that! 🎉 

The guidance might be a good opportunity to suggest links that most services ought to have in the footer. ie things like Accessibility statement, Contact us, Terms of use, Cookies, Privacy policy, etc.

Maybe "Give feedback" should be there too (instead of in a phase banner)?

colinrotherham:
Hope this is alright, but I noticed @anandamaryon1 assumed `navigation.items` in the changelog

I've pushed up a change to use `if params.navigation is mapping` so arrays/objects are interchangeable:

```patch
- navigation: {
+ navigation: [{
    items: [
      {
        text: "List 1, item 1",
        href: "/example-1-1"
      },
      {
        text: "List 1, item 2",
        href: "/example-1-2"
      }
    ]
  },
+ {
+   items: [
+     {
+       text: "List 2, item 1",
+       href: "/example-2-1"
+     },
+     {
+       text: "List 2, item 2",
+       href: "/example-2-2"
+     }
+   ]
+ }],
```

CHANGELOG.md
```diff
@@ -2,6 +2,124 @@
 
 ## Unreleased
 
+:boom: **Breaking changes**
+
+You must make the following changes when you migrate to this release, or your service might break.
+
+#### Update footer component
+
+The footer component's Nunjucks macros and HTML have been updated for better consistency and flexiblity.
```
frankieroberto:
```suggestion
The footer component's Nunjucks macros and HTML have been updated for better consistency and flexibility.
```

packages/nhsuk-frontend/src/nhsuk/components/footer/macro-options.mjs
```diff
@@ -8,66 +8,112 @@ export const name = 'Footer'
...
+          href: {
+            type: 'string',
+            required: true,
+            description: 'Footer navigation link `href` attribute.'
+          },
+          text: {
+            type: 'string',
+            required: true,
+            description: 'Footer navigation link text.'
+          },
```
frankieroberto: `html` is also now supported for the footer links, so that should be mentioned here too?

packages/nhsuk-frontend/src/nhsuk/components/footer/macro-options.mjs
```diff
@@ -8,66 +8,112 @@ export const name = 'Footer'
...
+        type: 'array',
+        required: false,
+        description: 'Contains the array of key policy footer link items.',
+        params: {
+          href: {
+            type: 'string',
+            required: true,
+            description: 'Footer meta link `href` attribute.'
+          },
+          text: {
```
frankieroberto, line 114: `html` is also supported for meta link items.

packages/nhsuk-frontend/src/nhsuk/components/footer/template.njk
```diff
@@ -1,81 +1,53 @@
...
-    </div> {# close nhsuk-width-container #}
-  </div> {# close nhsuk-footer-container #}
+        {{ _footerList(navigation.items) | indent(10) }}
+      </div>
+    {% endfor %}
+    </div>
+  {% endfor %}
+{% endif %}
+    <div class="nhsuk-footer__meta">
+{% if params.meta %}
```
frankieroberto, line 54: indent?

packages/nhsuk-frontend/src/nhsuk/components/footer/macro-options.mjs
colinrotherham: Great spot, added in https://github.com/nhsuk/nhsuk-frontend/pull/1452/commits/dfce80ff9e0c3a87ff3576a96f90d9d3005e418d

packages/nhsuk-frontend/src/nhsuk/components/footer/macro-options.mjs
colinrotherham: Thanks, another one in https://github.com/nhsuk/nhsuk-frontend/pull/1452/commits/dfce80ff9e0c3a87ff3576a96f90d9d3005e418d

packages/nhsuk-frontend/src/nhsuk/components/footer/template.njk
colinrotherham: I thought the same but it's aligned with the if/else above

It's a bit jarring, but if we un-indent the nested if/else then the HTML looks double-indented 😣

Can solve it with more macros like `_footerList()` but one for another day?

colinrotherham:
Do we need the `contentLicence` param like the GOV.UK Footer?

Otherwise for the service manual website are we alright with a different order?

## Before

Footer before

## After

Ignore the list item, compare the OGL content

Footer after

colinrotherham (approved):

colinrotherham:
Also a thought, do we want to support text only list items, without links?

For example, the 4th column as a bulleted list

packages/nhsuk-frontend/src/nhsuk/components/footer/template.njk
paulrobertlloyd: Spent quite a bit of time looking at the HTML output on this update ☺️

paulrobertlloyd:
> Do we need the contentLicence param like the GOV.UK Footer?

I don’t know what value it adds given we have `meta.text|html` (I spend most of my time with the GOV.UK footer trying to removing the OGL license, which also has the logo horribly baked in). I think the order is fine… the copyright statement is no longer right aligned, so it should be the last thing on the page.

paulrobertlloyd:
> Also a thought, do we want to support text only list items, without links?

I can’t think of any examples where this might be used. We could allow `html|text` for the navigation area?

colinrotherham:
> I can’t think of any examples where this might be used. We could allow `html|text` for the navigation area?

Adding `text`, `html` and column `width` does the job

Example like this:

Footer with custom HTML

colinrotherham:
I've added the `columns` per row count from GOV.UK Frontend (previously hard coded to 4)

Pretty flexible now, here's `"columns": 3` with a custom `width` on each:

Footer with columns and custom HTML

paulrobertlloyd:
> I've added the `columns` per row count from GOV.UK Frontend (previously hard coded to 4)

Nice! I wasn’t sure how far to go with the customisation options, but the example you’ve shown and the implemented look ace. ❤️

PR #1462, opened 2025-07-07
Add table cell modifier for missing data

frankieroberto:
This modifier makes the text dark grey for table cells which contain a short label explaining why the data is missing.

The [GOV.UK content guide for tables](https://www.gov.uk/guidance/content-design/tables) suggests not having empty cells in tables, but instead including a short label to explain why the data is missing:

> Screenshot 2025-07-07 at 14 41 15

This pull requests adds a table cell modifier to make that text secondary colour (dark grey) so that it is easier to perceive as being different from the cells containing actual data.

## Screenshot

Screenshot 2025-07-07 at 14 40 20

## Checklist

- [ ] CHANGELOG entry

### DIFF
--- a/packages/nhsuk-frontend/src/nhsuk/components/tables/_index.scss
+++ b/packages/nhsuk-frontend/src/nhsuk/components/tables/_index.scss
@@ -179,3 +179,9 @@
 .nhsuk-table__cell--numeric {
   text-align: right;
 }
+
+// Cells which contain a short label to explain
+// why they are otherwise empty
+.nhsuk-table__cell--missing-reason {
+  color: $nhsuk-secondary-text-color;
+}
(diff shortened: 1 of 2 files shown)


edwardhorsford (changes requested):
I'm unsure about bespoke styles like this just for one component - what about other things like summary lists? Would a generic secondary text style thing be better?

If we do keep here, I'd suggest renaming to a less use case specific name - why is it currently called 'reason'?

frankieroberto:
Closing this as we already have `nhsuk-u-secondary-text-color` which can be used (that I didn’t know about).

I'll open PR on the service manual repo to suggest some guidance.

frankieroberto:
PR for the guidance: https://github.com/nhsuk/nhsuk-service-manual/pull/2228

Issue #1348, opened 2025-06-02
Support initial `aria-describedby` on all form fields


colinrotherham:
Placeholder issue to bring in `params.describedBy` from GOV.UK Frontend:

* https://github.com/alphagov/govuk-frontend/pull/1347

PR #1468, opened 2025-07-08
Support initial `aria-describedby` on all form fields

colinrotherham:
## Description

This PR adds a feature from GOV.UK Frontend to close https://github.com/nhsuk/nhsuk-frontend/issues/1348

* https://github.com/alphagov/govuk-frontend/pull/1347

### Components with fieldsets

The following components were documented to support `params.fieldset.describedBy` but it was missing:

* Checkboxes
* Radios
* Date input

### Documented but not supported

The following components were documented to support `params.describedBy` but it was missing:

* ~Character count~
* Textarea
* Text input

**Note:** Character count should not support `params.describedBy` as it must be set to the "count message" hint ID instead

## Checklist

- [x] CHANGELOG entry


colinrotherham:
@edwardhorsford We might not need this now, but is it helpful for our wrapping fieldset approach?

PR #1469, opened 2025-07-08
Remove deprecated code and legacy feature detection

colinrotherham:
## Description

Following https://github.com/nhsuk/nhsuk-frontend/issues/1236, this PR removes various features no longer needed, for example:

* Add common helper `getFragmentFromUrl()` to replace `SkipLink.getHref()` IE workaround
* Skip unnecessary [`pageshow` feature](https://caniuse.com/mdn-api_window_pageshow_event) detection for IE8–10
* Remove [deprecated `event.keyCode`](https://caniuse.com/mdn-api_keyboardevent_keycode)
* Remove tab component unused custom events

## Checklist

- [x] CHANGELOG entry

### DIFF
--- a/packages/nhsuk-frontend/src/nhsuk/common.mjs
+++ b/packages/nhsuk-frontend/src/nhsuk/common.mjs
@@ -1,4 +1,21 @@
 /**
+ * Get hash fragment from URL
+ *
+ * Extract the hash fragment (everything after the hash) from a URL,
+ * but not including the hash symbol
+ *
+ * @param {string} url - URL
+ * @returns {string | undefined} Fragment from URL, without the hash
+ */
+export function getFragmentFromUrl(url) {
+  if (!url.includes('#')) {
+    return undefined
+  }
+
+  return url.split('#').pop()
+}
+
+/**
  * Toggle a boolean attribute on a HTML element
  *
  * @param {HTMLElement} element
--- a/packages/nhsuk-frontend/src/nhsuk/components/checkboxes/checkboxes.mjs
+++ b/packages/nhsuk-frontend/src/nhsuk/components/checkboxes/checkboxes.mjs
@@ -38,7 +38,7 @@ export class Checkboxes extends Component {
 
     // Attach handleClick as click to inputs
     this.$inputs.forEach((checkboxButton) => {
-      checkboxButton.addEventListener('change', this.handleClick.bind(this))
+      checkboxButton.addEventListener('click', this.handleClick.bind(this))
     })
   }
 
--- a/packages/nhsuk-frontend/src/nhsuk/components/checkboxes/checkboxes.mjs
+++ b/packages/nhsuk-frontend/src/nhsuk/components/checkboxes/checkboxes.mjs
@@ -28,17 +28,8 @@ export class Checkboxes extends Component {
 
     // When the page is restored after navigating 'back' in some browsers the
     // state of form controls is not restored until *after* the DOMContentLoaded
-    // event is fired, so we need to sync after the pageshow event in browsers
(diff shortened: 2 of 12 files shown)


packages/nhsuk-frontend/src/nhsuk/components/checkboxes/checkboxes.mjs
```diff
@@ -38,7 +38,7 @@ export class Checkboxes extends Component {
 
     // Attach handleClick as click to inputs
     this.$inputs.forEach((checkboxButton) => {
-      checkboxButton.addEventListener('change', this.handleClick.bind(this))
+      checkboxButton.addEventListener('click', this.handleClick.bind(this))
```
MatMoore, line 41: What's the reason for using click over change here? Are there cases where change won't fire?

packages/nhsuk-frontend/src/nhsuk/components/tabs/tabs.mjs
```diff
@@ -55,9 +55,6 @@ export class Tabs extends Component {
 
     this.jsHiddenClass = 'nhsuk-tabs__panel--hidden'
 
-    this.showEvent = new CustomEvent('tab.show')
```
MatMoore, line 58: Could these events be useful to external code? I'd consider it a breaking change to remove them, so maybe worth a callout in the changelog?

packages/nhsuk-frontend/src/nhsuk/common.mjs
```diff
@@ -1,3 +1,20 @@
...
+ *
+ * @param {string} url - URL
+ * @returns {string | undefined} Fragment from URL, without the hash
+ */
+export function getFragmentFromUrl(url) {
+  if (!url.includes('#')) {
+    return undefined
+  }
+
+  return url.split('#').pop()
```
MatMoore, line 15: I noticed this implementation looks like it returns everything after the last # whereas the other implementation looks like it returns everything after the first one. I doubt it's an issue in practice but I think you could technically have a frament of "#foo#bar"

packages/nhsuk-frontend/src/nhsuk/components/checkboxes/checkboxes.mjs
colinrotherham: Yeah, back in the day Internet Explorer would only fire the `change` event once focus was lost

It likely has no difference now but we match GOV.UK Frontend by using `click` instead

packages/nhsuk-frontend/src/nhsuk/components/tabs/tabs.mjs
colinrotherham: I searched both **NHSDigital** and **nhsuk** GitHub orgs and couldn't find any usage

We're assuming it was used by the NHS website where this code was customised

packages/nhsuk-frontend/src/nhsuk/common.mjs
colinrotherham: Thanks Mat. It's the implementation from GOV.UK Frontend so we'll have to keep an eye on that

packages/nhsuk-frontend/src/nhsuk/components/tabs/tabs.mjs
colinrotherham: I've added a changelog entry anyway [in the last push](https://github.com/nhsuk/nhsuk-frontend/compare/dce87147b6e237bf7a4ec7a5ab370df092b9009d..ab12f2cfbe8e6ff2a901a6e4f09d8ba17eb2c28a), just in case

PR #1471, opened 2025-07-10
Fix styles for button as link :visited when :active and :focus

### LINKED ISSUE #1372: Buttons as links, when active and focus applied, in firefox, have wrong text colour
# Bug Report

## What is the issue?
In Firefox (latest v139) on Mac or Windows, when a button link is pressed, `:active` and `:focus` applied, show with white text, where it should be black. 

Applies to primary, login and warning buttons. 

Image

Different issue for secondary button: 

Image

On Firefox, go here and click and hold buttons as link elements: https://nhsuk.github.io/nhsuk-frontend/examples/all-buttons/

Assume it's to do with CSS specificity.

anandamaryon1:
## Description

To resolve #1372

There may be a smarter way, to not apply focus styles on active buttons, but I couldn't get that working and this seems to work.
 
## Checklist

- [x] CHANGELOG entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,7 @@ This change was introduced in pull requests [#1459: Add NHS.UK frontend browser 
 - [#1467: Update components to set a default `id` based on `name`](https://github.com/nhsuk/nhsuk-frontend/pull/1467)
 - [#1468: Support initial `aria-describedby` on all form fields](https://github.com/nhsuk/nhsuk-frontend/pull/1468)
 - [#1469: Remove deprecated code and legacy feature detection](https://github.com/nhsuk/nhsuk-frontend/pull/1469)
+- [#1471: Fix styles for button as link :visited when :active and :focus](https://github.com/nhsuk/nhsuk-frontend/pull/1471)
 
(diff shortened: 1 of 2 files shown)


colinrotherham:
Nice one! How do you reckon we should test this?

Wondering if BackstopJS's `focusSelector` and our (new) `activeSelector` could be applied at the same time?

colinrotherham:
Do we need to backport this fix to v9.x too?

anandamaryon1:
> Nice one! How do you reckon we should test this?
> 
> Wondering if BackstopJS's `focusSelector` and our (new) `activeSelector` could be applied at the same time?

Good idea, it'd be good to cover it in the tests. I'll have a go now. 

> Do we need to backport this fix to v9.x too?

We've not had any reports but it is out there in the wild so probably good to release the fix for v9 yes

PS. I've no idea what sonar analysis is complaining about.

colinrotherham:
>PS. I've no idea what sonar analysis is complaining about.

Just been on with John Collinson and a couple of tokens had expired (2 years)

All sorted now

colinrotherham (approved):

PR #1467, opened 2025-07-08
Update components to set a default `id` based on `name`

colinrotherham:
## Description

This PR adds the following Nunjucks line to a few missed components:

```njk
{%- set id = params.id if params.id else params.name -%}
```

Notably the [**Select** component](https://service-manual.nhs.uk/design-system/components/select) where it was only partially used

See this issue https://github.com/alphagov/govuk-frontend/issues/3497 for more information

## Checklist

- [ ] Tested against our [testing policy](https://github.com/nhsuk/nhsuk-frontend/blob/master/docs/contributing/testing.md) (Resolution, Browser & Accessibility)
- [x] Follows our [coding standards and style guide](https://github.com/nhsuk/nhsuk-frontend/blob/master/docs/contributing/coding-standards.md)
- [x] CHANGELOG entry

PR #1486, opened 2025-07-14
Various JSDoc + type checking fixes

colinrotherham:
## Description

This PR adds JSDoc blocks to document our public API

It also fixes various issues identified by the TypeScript compiler and ESLint

For example:

### JSDoc blocks

I've added JSDoc blocks as used by GOV.UK Frontend

This enables code autocomplete and improves ESLint checks using typed parameters:

```patch
  /**
+  * Debounce double-clicks
+  *
   * If the click quickly succeeds a previous click then nothing will happen.
   * This stops people accidentally causing multiple form submissions by
   * double clicking buttons.
+  *
+  * @param {MouseEvent} event - Mouse click event
+  * @returns {undefined | false} Returns undefined, or false when debounced
   */
  debounce(event) {
    const { target } = event
```

### DOM selector guards

DOM selectors such as `getQuerySelector()` and `getElementById` return `null` when elements are not found

Similarly the [HTMLInputElement API `input.form`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/form) will return `null` without a `<form>` owner

I've copied GOV.UK Frontend to add simple guards such as:

```patch
  const $panel = this.getPanel($tab)
+  if (!$panel) {
+    return
+  }

  $panel.setAttribute('role', 'tabpanel')
  $panel.setAttribute('aria-labelledby', $tab.id)
```

## Checklist

- [ ] CHANGELOG entry


shared/tasks/scripts.mjs
```diff
@@ -66,8 +66,7 @@ export function compile(
 
     // Add minifier plugin (optional)
     if (output.compact) {
-      output.plugins ??= []
-      output.plugins.push(
+      output.plugins = [
```
MatMoore, line 69: this is now overriding output.plugins if already set, is that ok?

packages/nhsuk-frontend-review/src/examples/all.njk
```diff
@@ -800,7 +800,6 @@
         "label": {
           "text": "NHS number"
         },
-        "id": "input-example",
```
MatMoore, line 803: are these unnecessary because they're not used in the template, or because you want a minimal example?

if the latter it would be good to make sure it is included in at least one example per component. I'm hoping to rely on this to exercise all the params in the tests for the jinja port.

packages/nhsuk-frontend/src/nhsuk/components/button/button.jsdom.test.mjs
```diff
@@ -87,14 +99,72 @@ describe('Button', () => {
...
+
+      fireEvent($root, event)
+      expect(event.preventDefault).not.toHaveBeenCalled()
+
+      fireEvent($root, event)
+      expect(event.preventDefault).toHaveBeenCalled()
     })
+  })
 
+  describe('Accessibility', () => {
```
MatMoore: Not sure how strict we normally are about the test names, but in similar frameworks there is a convention of having the `describe`/`it` text create readable sentences. This is the case for the double click tests, but not the accessibility ones, e.g. "Accessibility should not trigger the click event[...]"

Is it worth reframing this?

packages/nhsuk-frontend/src/nhsuk/components/button/button.jsdom.test.mjs
```diff
@@ -87,14 +99,72 @@ describe('Button', () => {
...
     it('should have accessible name and role', () => {
       expect($root).toHaveAccessibleName('Save and continue')
       expect($root).toHaveRole('button')
     })
+
+    it('should not trigger the click event when the space key is pressed', async () => {
```
MatMoore, line 155: I was confused as to why these two tests seemed to be asserting the opposite of each other until I realised they are referring to a different "it"

This one uses the default example, and the one below uses the link example. Just wondering if this could be clearer from the name.

packages/nhsuk-frontend/src/nhsuk/components/button/button.jsdom.test.mjs
colinrotherham: Yeah probably!

It's more like this at the moment, like a subheading or category:

>**Character count** (Accessibility) should not trigger the click event[…]

packages/nhsuk-frontend-review/src/examples/all.njk
colinrotherham: Mainly to better inform users of this GOV.UK Frontend change we added:

* https://github.com/nhsuk/nhsuk-frontend/pull/1467

i.e. That `params.name` will populate the ID by default so it's not necessary unless it needs overriding

Should I add a new example per component named "with custom ID" or similar?

shared/tasks/scripts.mjs
colinrotherham: Yeah, it was a mistake to let the Gulp task configure Rollup plugins

Thanks for being thorough

packages/nhsuk-frontend-review/src/examples/all.njk
MatMoore: I think that would be helpful yeah, or an example that sets everything that can possibly be set?

packages/nhsuk-frontend/src/nhsuk/components/button/button.jsdom.test.mjs
MatMoore: ah ok, that's better than what I was assuming then. Feel free to ignore this suggestion

packages/nhsuk-frontend-review/src/examples/all.njk
colinrotherham: I've brought back IDs for all the form examples and had a big spring clean in https://github.com/nhsuk/nhsuk-frontend/pull/1486/commits/ff4aaba86540eaffce960af1b701c319fb8246e8

I've added a few missing examples (e.g. dates with only two fields) but can add extra "everything" examples perhaps in future. We've got enough for the test coverage now

packages/nhsuk-frontend/src/nhsuk/components/button/button.jsdom.test.mjs
colinrotherham: Good feedback, I've addressed this in https://github.com/nhsuk/nhsuk-frontend/pull/1486/commits/bd26df41b2a9ce3ac2f9f4ec4b4b39126bbad74d

PR #1487, opened 2025-07-15
Fix character count not having error border colour when rendered with error message

colinrotherham:
## Description

This PR fixes an issue in the character count component where the red error border is removed, for example:

1. Textarea error class removed (on init) when already under `maxlength` or `maxwords`
2. Textarea error class removed (on keyup) when moving under `maxlength` or `maxwords`

Already fixed on GOV.UK Frontend in https://github.com/alphagov/govuk-frontend/pull/5896

### Before
 
Character count before, without red border

### After

Character count after, with red border

## Checklist

- [x] CHANGELOG entry


packages/nhsuk-frontend/src/nhsuk/components/character-count/template.njk
```diff
@@ -15,7 +15,7 @@
     spellcheck: params.spellcheck,
     value: params.value,
     formGroup: params.formGroup,
-    classes: 'nhsuk-js-character-count' + (' nhsuk-textarea--error' if params.errorMessage) + (' ' + params.classes if params.classes),
+    classes: 'nhsuk-js-character-count' + (' ' + params.classes if params.classes),
```
colinrotherham, line 18: The error class was added by both `characterCount()` and `textarea()`

I've removed it here as only `textarea()` needs `params.errorMessage`

PR #1506, opened 2025-07-24
Add JavaScript configuration support to components

colinrotherham:
## Description

This PR ensures all configurable components support both Nunjucks and JavaScript configurations

* **Button component**
  Supports option `preventDoubleClick`

* **Error summary component**
  Supports options `disableAutoFocus` to replace (deprecated) `focusOnPageLoad`

* **Notification banner component**
  Supports option `disableAutoFocus`

**Note:** Character count needs [`%{count} characters` placeholders](https://frontend.design-system.service.gov.uk/localise-govuk-frontend/#localise-gov-uk-frontend) so is not yet supported

---

A new base class `ConfigurableComponent` and utilties have been ported from GOV.UK Frontend

* [**Passing configuration using data attributes in HTML**](https://frontend.design-system.service.gov.uk/v4/configure-components-with-javascript/#setting-nunjucks-macro-options)
* [**Passing configuration to a new instance of a component in JavaScript**](https://frontend.design-system.service.gov.uk/v4/configure-components-with-javascript/#passing-configuration-using-data-attributes-in-html)

Which provides support for the following:

### Component configuration via `initAll()`

```js
import { initAll } from 'nhsuk-frontend'

initAll({
  button: {
    preventDoubleClick: true
  }
})
```

### Component configuration via `createAll()`

```js
import { createAll, Button } from 'nhsuk-frontend'

// Initialise all button components
createAll(Button, {
  preventDoubleClick: true
})
```

### Component configuration via classes

```js
import { Button } from 'nhsuk-frontend'

const $button = document.querySelector('.app-button')

// Initialise single button component
new Button($button, {
  preventDoubleClick: true
})
```

## Checklist

- [x] CHANGELOG entry

### DIFF
--- a/docs/installation/installing-with-npm.md
+++ b/docs/installation/installing-with-npm.md
@@ -125,31 +125,31 @@ if ($element) {
 
 #### Initialise individual components
 
-Rather than using `initAll`, you can initialise all instances of individual components used by your service. For example:
+Rather than using `initAll`, you can initialise all instances of individual components using `createAll`. For example:
 
 ```js
-import { initButtons, initRadios } from 'nhsuk-frontend';
+import { createAll, Button, Checkboxes } from 'nhsuk-frontend'
 
 // Initialise all button components
-initButtons();
+createAll(Button)
 
-// Initialise all radios components
-initRadios();
+// Initialise all checkboxes components
+createAll(Checkboxes)
 ```
 
 Or where necessary, single instances of individual components only:
 
 ```js
-import { Button, Checkboxes } from 'nhsuk-frontend';
+import { Button, Checkboxes } from 'nhsuk-frontend'
 
 const $button = document.querySelector('.app-button')
 const $checkboxes = document.querySelector('.app-checkboxes')
 
 // Initialise single button component
-new Button($button);
+new Button($button)
 
 // Initialise single checkboxes component
-new Checkboxes($checkboxes);
+new Checkboxes($checkboxes)
 ```
 
 ## Importing assets (optional)
--- a/docs/installation/installing-with-npm.md
+++ b/docs/installation/installing-with-npm.md
@@ -107,6 +107,18 @@ import { initAll } from 'nhsuk-frontend'
 initAll()
 ```
 
+You can pass configuration to components by including key-value pairs of camel-cased component names with their options:
+
+```js
+import { initAll } from 'nhsuk-frontend'
+
+initAll({
+  button: {
+    preventDoubleClick: true
+  }
+})
+```
+
 #### Initialise only part of a page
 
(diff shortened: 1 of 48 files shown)


packages/nhsuk-frontend/src/nhsuk/configurable-component.mjs
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+ * Configurable base component class
+ *
+ * @abstract
+ * @template {Partial<Record<keyof ConfigurationType, unknown>>} [ConfigurationType=ObjectNested]
+ * @template {HTMLElement} [RootElementType=HTMLElement]
+ * @augments Component<RootElementType>
+ */
+export class ConfigurableComponent extends Component {
```
MatMoore, line 17: Just wondering - is there a reason why the functionality in this class is a separate subclass, rather than being part of `Component` itself? (i.e. a non-configurable component would just have an empty config)

docs/installation/installing-with-npm.md
```diff
@@ -107,6 +107,18 @@ import { initAll } from 'nhsuk-frontend'
 initAll()
 ```
 
+Or you can pass configuration to components by including key-value pairs of camel-cased component names with their options:
```
MatMoore: Think the "or" might be unnecessary in this sentence - this sounds more like a continuation of the previous para rather than an alternative way of initialising?

MatMoore (approved):
There's lots of new code here, but going commit by commit it broadly looks good to me. Are there any areas where we have diverged from what GOV.UK frontend is doing that might have maintenance implications later? If not, happy to merge :shipit:

packages/nhsuk-frontend/src/nhsuk/configurable-component.mjs
colinrotherham: It's primarily for [tree shaking](https://rollupjs.org/introduction/#tree-shaking), so unused JavaScript can be dropped

For projects that never import a configurable component, all the config-related utilities are omitted too

colinrotherham:
>There's lots of new code here, but going commit by commit it broadly looks good to me. Are there any areas where we have diverged from what GOV.UK frontend is doing that might have maintenance implications later? If not, happy to merge :shipit:

@MatMoore Only two areas where we've diverged:

1. We support both `initAll($scope)` and `initAll({ scope: $scope })` for _our_ backwards compatibility
2. We're free to use [separate helpers](https://github.com/nhsuk/nhsuk-frontend/tree/configurable-component/packages/nhsuk-frontend/src/nhsuk/common/configuration) versus [everything in **configuration.mjs**](https://github.com/alphagov/govuk-frontend/blob/main/packages/govuk-frontend/src/govuk/common/configuration.mjs) for _their_ backwards compatibility

Thanks again for the review

PR #1511, opened 2025-07-31
Use pseudo element for ticks on contents list items

paulrobertlloyd:
## Description

This PR uses pseudo element to generate the ticks for contents list items. By using a border, the added benefit is that this line will also show for users using a high contrast mode.

Partly addresses #1028 by removing the `emdash` and `emdash-small` SVG icons, which aren’t listed in the service manual.

I also updated the design to use the same length regardless of breakpoint; the 3px difference doesn’t seem worth the extra KBs. This means the width of the tick now uses a standard size.

(I ran visual regression tests locally and these all passed… might need to try that again when I get into the office as they should fail given the above mentioned changed)

## Checklist

- [x] CHANGELOG entry


paulrobertlloyd:
Actually, come of think it, this should probably use `::marker`…

colinrotherham:
>(I ran visual regression tests locally and these all passed… might need to try that again when I get into the office as they should fail given the above mentioned changed)

We dropped the diff threshold down to 0.1% but the [changes for antialiasing](https://github.com/nhsuk/nhsuk-frontend/commit/019990618f8fae0c92f586133bda9fc39972f5e7) made it more lenient

Try temporarily adding `misMatchThreshold: 0` and removing these lines?

```patch
  report: ['browser'],
- resembleOutputOptions: {
-   ignoreAntialiasing: true
- },
```

It's a good trick to update the reference images when the tiniest changes would otherwise be skipped

paulrobertlloyd:
> Actually, come of think it, this should probably use `::marker`…

Ah damn, [Safari is a blocker](https://bugs.webkit.org/show_bug.cgi?id=204163) here

paulrobertlloyd:
Visual regression changes added (see if you can spot the difference!) and added item to changelog.

packages/nhsuk-frontend/src/nhsuk/components/contents-list/_index.scss
```diff
@@ -20,14 +20,17 @@
...
 
-  @include nhsuk-media-query($from: tablet) {
-    background: url("data:image/svg+xml,%3Csvg class='nhsuk-icon nhsuk-icon__emdash' xmlns='http://www.w3.org/2000/svg' fill='%23aeb7bd' width='16' height='1' aria-hidden='true'%3E%3Cpath d='M0 0h19v1H0z'%3E%3C/path%3E%3C/svg%3E")
-      left nhsuk-px-to-rem(14px) no-repeat;
+  &::before {
+    content: "";
+    display: block;
+    left: 0;
+    position: absolute;
+    top: 0.75em;
```
paulrobertlloyd, line 31: @colinrotherham Should this perhaps use `nhsuk-px-to-rem`?

colinrotherham: You read my mind, I wrote this very comment out but deleted it 😊

It's a little more verbose but does protects us from font sizes potentially getting out of sync in future

Up to you

paulrobertlloyd: Actually, just tried this and it doesn’t work. Let’s leave it as it is.

PR #1515, opened 2025-08-01
Update back link and breadcrumb components

paulrobertlloyd:
## Description

Partially addresses #1028, and removes the final set of SVG files inlined within our CSS. It also does the following:

- simplifies back link by removing the inlined SVG and aligning markup with GOV.UK Frontend
- hides the back link for print media
- uses consistent class naming for breadcrumb items (renaming `nhsuk-breadcrumb__item` to `nhsuk-breadcrumb__list-item`)
- uses the back-link component for the breadcrumb mobile presentation (where a back link is shown instead) saving us from repeating styles and markup
- tidies up markup in the templates for these 2 components
- adds a `nhsuk-shape-chevron` mixin to generate chevrons (can adjust size, colour, direction and display property if needed)
- adds a `nhsuk-chevron-size()` function to the get the calculated distance between a chevron’s base and it’s tip
- removes the `chevron-left` and `chevron-left` SVG icons and their associated styles

### Back link component

Before:

```html
<div class="nhsuk-back-link"> <a class="nhsuk-back-link__link" href="#">
  <svg class="nhsuk-icon nhsuk-icon__chevron-left" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" aria-hidden="true" height="24" width="24">
    <path d="M8.5 12c0-.3.1-.5.3-.7l5-5c.4-.4 1-.4 1.4 0s.4 1 0 1.4L10.9 12l4.3 4.3c.4.4.4 1 0 1.4s-1 .4-1.4 0l-5-5c-.2-.2-.3-.4-.3-.7z"></path>
  </svg>
  Back</a>
</div>
```

After:

```html
<a class="nhsuk-back-link" href="#">Back</a>
```

### Breadcrumbs component

Before:

```html
<nav class="nhsuk-breadcrumb" aria-label="Breadcrumb">
  <ol class="nhsuk-breadcrumb__list">
    <li class="nhsuk-breadcrumb__item"><a class="nhsuk-breadcrumb__link" href="#">Level one</a></li>
    <li class="nhsuk-breadcrumb__item"><a class="nhsuk-breadcrumb__link" href="#">Level two</a></li>
    <li class="nhsuk-breadcrumb__item"><a class="nhsuk-breadcrumb__link" href="#">Level three</a></li>
  </ol>
  <p class="nhsuk-breadcrumb__back">
    <a class="nhsuk-breadcrumb__backlink" href="#">
      <span class="nhsuk-u-visually-hidden">Back to &nbsp;</span>
      Level three
    </a>
  </p>
</nav>
```

After:

```html
<nav class="nhsuk-breadcrumb" aria-label="Breadcrumb">
  <ol class="nhsuk-breadcrumb__list">
    <li class="nhsuk-breadcrumb__list-item">
      <a class="nhsuk-breadcrumb__link" href="#">Level one</a>
    </li>
    <li class="nhsuk-breadcrumb__list-item">
      <a class="nhsuk-breadcrumb__link" href="#">Level two</a>
    </li>
    <li class="nhsuk-breadcrumb__list-item">
      <a class="nhsuk-breadcrumb__link" href="#">Level three</a>
    </li>
  </ol>
  <a class="nhsuk-back-link" href="#"><span class="nhsuk-u-visually-hidden">Back to&nbsp;</span>Level three</a>
</nav>
```

## Checklist

- [x] CHANGELOG entry


packages/nhsuk-frontend/src/nhsuk/components/breadcrumb/template.njk
```diff
@@ -1,27 +1,31 @@
+{% from "nhsuk/components/back-link/macro.njk" import backLink %}
 {% from "nhsuk/macros/attributes.njk" import nhsukAttributes %}
 
 <nav class="nhsuk-breadcrumb{% if params.classes %} {{ params.classes }}{% endif %}" aria-label="{{ params.labelText | default("Breadcrumb") }}"
-{{- nhsukAttributes(params.attributes) }}>
+  {{- nhsukAttributes(params.attributes) }}>
```
colinrotherham, line 5: 🙌

packages/nhsuk-frontend/src/nhsuk/components/breadcrumb/template.njk
```diff
@@ -1,27 +1,31 @@
+{% from "nhsuk/components/back-link/macro.njk" import backLink %}
```
colinrotherham, line 1: This is a lovely workaround, and keeps NHS behaviour (thank you)

**One catch:** Can you make sure the back link Sass is forwarded by the breadcrumb Sass please?

Otherwise, services that selectively import _only_ the breadcrumb Sass will have missing styles

packages/nhsuk-frontend/src/nhsuk/components/breadcrumb/template.njk
```diff
@@ -1,27 +1,31 @@
...
   </ol>
-  <p class="nhsuk-breadcrumb__back">
-    <a class="nhsuk-breadcrumb__backlink" href="{{ lastHref }}" {{- nhsukAttributes(lastItem.attributes) }}>
-      <span class="nhsuk-u-visually-hidden">Back to &nbsp;</span>
-      {{ lastText }}
-    </a>
-  </p>
+  {{ backLink({
+    href: lastHref,
+    html: '<span class="nhsuk-u-visually-hidden">Back to&nbsp;</span>' + lastText,
```
colinrotherham, line 28: There was previously a line break before `{{ lastText }}`, do we need it?

tests/backstop/bitmaps_reference/Back_link_as_a_button_active_mobile.png
colinrotherham, line 1: Vertical alignment looks a bit off here?

tests/backstop/bitmaps_reference/Back_link_reverse_focus_mobile.png
colinrotherham, line 1: This might get fixed when we review link styles, but the GOV version appears to have a couple of pixels padding before the arrow hits the left hand side of the focus box

CHANGELOG.md
```diff
@@ -4,6 +4,80 @@
 
 :boom: **Breaking changes**
 
+#### Update back link component
+
+HTML markup for the back link component has been updated to align with that used on GOV.UK.
```
colinrotherham: Flagging this as I've previously said "GOV.UK Frontend" rather than "GOV.UK"

Not sure which one is better 🤷‍♂️ 

```suggestion
HTML markup for the back link component has been updated to align with GOV.UK Frontend.
```

packages/nhsuk-frontend/src/nhsuk/components/breadcrumb/template.njk
paulrobertlloyd: I don’t think so… this is how it renders without CSS:

Screenshot of un-styled breadcrumbs.

CHANGELOG.md
paulrobertlloyd: Updated.

packages/nhsuk-frontend/src/nhsuk/components/breadcrumb/template.njk
paulrobertlloyd: Done!

CHANGELOG.md
colinrotherham, line 1: Shall we continue to use the diff format instead of "Before" and "After" HTML?

CHANGELOG.md
```diff
@@ -4,6 +4,80 @@
...
+HTML markup for breadcrumbs component has been updated to better align with that used on GOV.UK and use back link styles for presentation on mobile.
+
+Before:
+
+```html
+<nav class="nhsuk-breadcrumb" aria-label="Breadcrumb">
+  <ol class="nhsuk-breadcrumb__list">
+    <li class="nhsuk-breadcrumb__item"><a class="nhsuk-breadcrumb__link" href="#">Level one</a></li>
+    <li class="nhsuk-breadcrumb__item"><a class="nhsuk-breadcrumb__link" href="#">Level two</a></li>
+    <li class="nhsuk-breadcrumb__item"><a class="nhsuk-breadcrumb__link" href="#">Level three</a></li>
```
edwardhorsford: Suggest putting the anchors on new lines like the `after` example so it's easier to compare the differences.

edwardhorsford:
Does removing the svg have any implications for users in forced-colour modes of browsers?

paulrobertlloyd:
> Does removing the svg have any implications for users in forced-colour modes of browsers?

Shouldn't do as we're using borders to create the chevrons (as they do in GOV.UK Frontend)

tests/backstop/bitmaps_reference/Back_link_as_a_button_desktop.png
anandamaryon1, line 1: I find the new chevron a bit too light/thin, in comparison to the weight of the text. Is it possible to thicken it slightly, somewhere between what we had originally and what we have here? 

Eg. 
From this: 
nhsuk-frontend-pr-1515 herokuapp com_nhsuk-frontend_components_back-link_default_ (1)

To something like: 
localhost_3000_nhsuk-frontend_components_back-link_default_ (3)

(tweaked to make `nhsuk-chevron-border` 2px min instead of 1.5px, and added 1px to `box-size` and added 1px `border-radius` to _imperfectly_ soften it.)

anandamaryon1:
> > Does removing the svg have any implications for users in forced-colour modes of browsers?
> 
> Shouldn't do as we're using borders to create the chevrons (as they do in GOV.UK Frontend)

Did a quick test on Windows + Edge and looks works fine 👍 

Thanks for this @paulrobertlloyd, love how it uses the same chevron for back link, breadcrumbs + mobile view. 

Only quibble about the visual weight of the icon.

CHANGELOG.md
paulrobertlloyd: Done!

CHANGELOG.md
paulrobertlloyd: Updated (and now using `diff` syntax highlighting)

tests/backstop/bitmaps_reference/Back_link_as_a_button_desktop.png
paulrobertlloyd: Have updated the design to use a 2px border width with 1px border radius.

However, making this change introduced an ever so slight (yet annoying very visible) hairline gap where the two borders overlap in Safari (in both the current version and in Safari Technical Preview):

Screenshot 2025-08-05 at 17 23 32

I’ve managed to fix this by using an inset outline instead, for which I’ve then hidden two sides of using a `clip-path()`. We may want to test that this approach also works in high contrast mode and in other browsers/platforms.

anandamaryon1: Yep, still works in high-contrast 👍 

Playing with your new approach, I'd like to suggest another tiny tweak:

`border-radius: 2px;` on chevron and `padding-left: nhsuk-em(nhsuk-spacing(2), $_font-size) + ($_chevron-size + 0.125); // [1]` to slightly increase the padding of back link. 

Radius only really noticeable on high-res screens but softens it a bit more. 

Before: 
nhsuk-frontend-pr-1515 herokuapp com_nhsuk-frontend_components_back-link_default_ (2)

After:
localhost_3000_nhsuk-frontend_components_back-link_default_ (4)


Happy for me to commit? Or is there a tidier way than `($_chevron-size + 0.125)` to slightly bump up the padding?

paulrobertlloyd: I’ve updated the chevron mixin so that it now uses a 2px border radius. I’ve also updated it so that this radius only gets applied to the chevron’s tip.

As to the padding, I’ve changed this so that it’s (essentially) 1em, which is less of a magic number, and meaning the padding stays the same as it is currently.

anandamaryon1: Nice, thanks

With the border radius, what you've done makes logical sense but I'd still prefer the very slight softening that just `border-radius: 2px;` gives to the top and bottom. (which is admittedly only visible on high density screens).

image
image

paulrobertlloyd: If only for the sake of comparison, these are the available options…

2px border radius on the tip only:

2px-0px

2px border radius on the tip, 1px on the two ‘feathers’:

2px-1px

2px border radius on the tip and the two ‘feathers’:

2px-2px


The later is a bit of an odd shape, but it does help the chevron better align with the text. And of course, these are zoomed in, how they render at 100% size is negligible.

paulrobertlloyd: Have updated PR to use 2px on all sides.

packages/nhsuk-frontend/src/nhsuk/components/breadcrumb/_index.scss
```diff
@@ -28,102 +36,59 @@
...
-  .nhsuk-breadcrumb--reverse &:active::before {
-    background-image: url($icon-background-reverse);
-  }
-
-  &:focus::before,
-  &:active::before,
-  .nhsuk-breadcrumb--reverse &:focus::before {
-    background-image: url($icon-background-focus);
+  .nhsuk-breadcrumb__list-item::before {
+    color: rgba($color_nhsuk-white, 0.5);
```
anandamaryon1:
```suggestion
    color: rgba($color_nhsuk-white, 0.6);
```
To get them to over 3:1 contrast ratio (on NHS blue at least)

paulrobertlloyd: Good catch, fixed!

CHANGELOG.md
```diff
@@ -30,6 +30,64 @@ This was added in [pull request #1506: Add JavaScript configuration support to c
...
+
+HTML markup for breadcrumbs component has been updated to better align with that used on GOV.UK and use back link styles for presentation on mobile:
+
+Before:
+
+```diff
+  <nav class="nhsuk-breadcrumb" aria-label="Breadcrumb">
+    <ol class="nhsuk-breadcrumb__list">
+-     <li class="nhsuk-breadcrumb__item">
++     <li class="nhsuk-breadcrumb__item-item">
```
colinrotherham: Are these supposed to be "list-item" @paulrobertlloyd?

```suggestion
+     <li class="nhsuk-breadcrumb__list-item">
```

paulrobertlloyd: Fixed!

Issue #1407, opened 2025-06-17
Remove deprecated features marked for removal in v11


colinrotherham:
In the later v9.x releases we marked various features with:

```scss
/// @deprecated To be removed in v11.0, replaced by xxxx
```

Similar to https://github.com/nhsuk/nhsuk-frontend/issues/1406 we should remove these before the v11.0.0 release

colinrotherham:
We also fixed the BEM syntax for icons in https://github.com/nhsuk/nhsuk-frontend/pull/1521

Backwards compatibility was maintained, but should we also remove this in v11?

```patch
- .nhsuk-icon__arrow-right-circle,
  .nhsuk-icon--arrow-right-circle {
```

Issue #662, opened 2020-11-30
Back link looks broken with increased text size


davidhunter08:
# Bug Report

## What is the issue?

When you increase the text size the back button text and icon become misaligned.

<img width="137" alt="Screenshot 2020-11-30 at 10 48 36" src="https://user-images.githubusercontent.com/14331000/100603415-fa743800-32fc-11eb-99f1-2ee7ca9c3df2.png">

## What steps are required to reproduce the issue?

Increase the browser text size and view a page with a back link on.

## What was the environment where this issue occurred?

* Device: Macbook Pro
* Operating System: Big Sur
* Browser: Chrome
* Browser version: Version 87.0.4280.67 (Official Build) (x86_64)
* NHS.UK frontend package version: 4.0.0

### Possible solution

Adding `margin: auto`, `bottom: 0` and `top: 0` looks to solve the issue. Needs more testing though.

```
.nhsuk-back-link__link .nhsuk-icon__chevron-left {
    bottom: 0;
    height: 24px;
    left: -8px;
    margin: auto;
    position: absolute;
    top: 0;
    width: 24px;
 }
```

<img width="133" alt="Screenshot 2020-11-30 at 10 49 36" src="https://user-images.githubusercontent.com/14331000/100603859-9ef67a00-32fd-11eb-99e8-d926a128fb5c.png">

### Other considerations

It might also be worth reviewing the appearance of the back link, as it currently doesn't match the breadcrumb back link.

Breadcrumb back link example (with increased text size):

<img width="191" alt="Screenshot 2020-11-30 at 10 52 12" src="https://user-images.githubusercontent.com/14331000/100603967-cb11fb00-32fd-11eb-95bf-e97c4778b088.png">

AdamChrimes:
We also need to look at the placement of this within the page, it should be placed outside the `<main>` container. The current styles don't allow for this without causing unintended spacing. A number of applications are currently hot fixing this with;

`<div class="nhsuk-back-link nhsuk-u-margin-top-4 nhsuk-u-margin-bottom-0">`

or

```
.nhsuk-back-link {
  @include nhsuk-responsive-padding(4, 'top');
  margin-bottom: 0;
}
```

This would most likely be a breaking change

AdamChrimes:
Similar issue with the Breadcrumb, Action link and Pagination when increasing the font size

[...]

PR #1521, opened 2025-08-02
Update component icons

### LINKED ISSUE #737: SVG elements receiving focus on IE11
SVG images on IE11 are focusable by default, so this means that when a keyboard user navigates the NHS sites on IE11, all SVG images receive focus automatically. This should not be the case because it is an accessibility fail and it also could confuse users who depend on keyboard for navigation and assistive technologies.

Below is more details on the issue and my resolution

Steps to recreate:
1. Open any nhs.uk site e.g https://www.nhs.uk/better-health/ on IE11
2. Use Tab to navigate through SVG images and icons
3. Observe the hidden selection on SVGs

Environment where this issue occurred:
Operating system - Windows 10
Browser: Internet Explorer 11
NHS.UK frontend package version: 3.1.0

Resolution:
All SVG images/icons should have an attribute _**focusable="false"**_ assigned to them. 
e.g
```
<svg class="nhsuk-icon nhsuk-icon__arrow-left" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" focusable="false">
  <path d="M4.1 12.3l2.7 3c.2.2.5.2.7 0 .1-.1.1-.2.1-.3v-2h11c.6 0 1-.4 1-1s-.4-1-1-1h-11V9c0-.2-.1-.4-.3-.5h-.2c-.1 0-.3.1-.4.2l-2.7 3c0 .2 0 .4.1.6z"></path>
</svg>
```

Below is a live GIF capture of the issue

invisible-links

paulrobertlloyd:
## Description

Partly addresses #1028 by using a macro to provide inline SVG markup for a collection of common icons, ensuring consistent icon sizing and markup, and providing a single source of truth for icons used across NHS.UK frontend.

This PR:

- Adds a new private `nhsukIcon` macro. This takes an icon name and optional accessible title, for example:

  ```njk

  <button type="submit">
    {{ nhsukIcon("search", {
      title: "Search site"
    }) }}
  </button>
  ```

  The following icons are available:

  - `arrow-left`
  - `arrow-right`
  - `arrow-right-circle`
  - `chevron-right-circle`
  - `cross`
  - `tick`
  - `search`
  - `user`

- Updates some of the icons so that all are drawn using a single `path`
- Updates styles relating to the revised icons, ensuring they scale with any associated text labels
- Adds a page to the review app to display all the available icons
- Removes SVG icon files for those now provided by the macro

## Notes

- This PR may also close #737. 
- The icon updates for the header could arguably be pulled into the upcoming v10 release; this would mean one less change is needed when these new icons are merged into the next breaking release.

Separately, this PR would mean the service manual can no longer link to icon files in this repo. If this PR is accepted, the plan would be to update the page in the manual to use the macro to output the SVG code that can be copy and pasted instead.

## Checklist

- [x] CHANGELOG entry


CHANGELOG.md
```diff
@@ -4,6 +4,167 @@
...
+          </a>
+        </li>
+        <li class="nhsuk-pagination-item--next">
+          <a class="nhsuk-pagination__link nhsuk-pagination__link--next" href="#">
+            <span class="nhsuk-pagination__title">Next</span>
+            <span class="nhsuk-u-visually-hidden">:</span>
+            <span class="nhsuk-pagination__page">Symptoms</span>
+  -         <svg class="nhsuk-icon nhsuk-icon__arrow-right" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" aria-hidden="true" width="34" height="34">
+  -           <path d="M19.6 11.66l-2.73-3A.51.51 0 0 0 16 9v2H5a1 1 0 0 0 0 2h11v2a.5.5 0 0 0 .32.46.39.39 0 0 0 .18 0 .52.52 0 0 0 .37-.16l2.73-3a.5.5 0 0 0 0-.64z"></path>
+  _         </svg>
```
colinrotherham: Mind double checking the [rendered HTML](https://github.com/nhsuk/nhsuk-frontend/blob/nhsuk-icon-macro/CHANGELOG.md#update-icons)?

Got this little fella thinking it's a minus:

```suggestion
  -         </svg>
```

paulrobertlloyd: Oops! Fixed.

colinrotherham:
Do we want this in v10 @anandamaryon1?

CHANGELOG.md
```diff
@@ -88,6 +88,153 @@ If you are not using Nunjucks macros, update your HTML markup using the [back li
...
+    <a class="nhsuk-header__account-link" href="#">
+  -   <svg class="nhsuk-icon nhsuk-icon__user" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" aria-hidden="true" focusable="false">
+  -     <path d="M12 1a11 11 0 1 1 0 22 11 11 0 0 1 0-22Zm0 2a9 9 0 0 0-5 16.5V18a4 4 0 0 1 4-4h2a4 4 0 0 1 4 4v1.5A9 9 0 0 0 12 3Zm0 3a3.5 3.5 0 1 1-3.5 3.5A3.4 3.4 0 0 1 12 6Z"></path>
+  -   </svg>
+  +   <svg class="nhsuk-icon nhsuk-icon__user" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" width="34" height="34" focusable="false" aria-hidden="true">
+  +     <path d="M12 1a11 11 0 1 1 0 22 11 11 0 0 1 0-22Zm0 2a9 9 0 0 0-5 16.5V18a4 4 0 0 1 4-4h2a4 4 0 0 1 4 4v1.5A9 9 0 0 0 12 3Zm0 3a3.5 3.5 0 1 1-3.5 3.5A3.4 3.4 0 0 1 12 6Z"/>
+  +   </svg>
+      florence.nightingale@nhs.net
+    </a>
+  ```
```
paulrobertlloyd: If we decide to include this change in v10, we won’t need to add this to the change log.

paulrobertlloyd:
@anandamaryon1 Primary card chevron icon updated.

packages/nhsuk-frontend/src/nhsuk/core/styles/_icons.scss
```diff
@@ -1,60 +1,18 @@
...
-  fill: $color_nhsuk-blue;
-}
-
-.nhsuk-icon__minus {
-  fill: $color_nhsuk-blue;
-}
-
 /// Icon size adjustments
 
 .nhsuk-icon--size-25 {
```
paulrobertlloyd, line 18: Can we remove these icon size classes? They don’t seem to be used anywhere, or documented in the service manual.

paulrobertlloyd:
Have further tweaked the styles associated with these icons to address the issue raised in #662. Be good to get a re-review on these more substantial icon style updates @colinrotherham.

Action link, pagination and do/don’t list icons should now scale relative to associated text. While the alignment starts to go slightly off, I think that’s less important than the fact the icons share the same size as the text:

## Action link

Before (px-based)

Action link before.

After (rem-based)

Action link after.

## Do/don’t list

Before (px-based)

Do list before.

After (rem-based)

Do list after.

## Header account

Before (px-based)

Header account before.

After (rem-based)

Header-account after.

## Pagination

Before (px-based)

Pagination before.

After (em-based)

Pagination after.

packages/nhsuk-frontend/src/nhsuk/core/styles/_icons.scss
paulrobertlloyd, line 20: Can we remove these icon size classes? They don’t seem to be used anywhere, or documented in the service manual.

colinrotherham: Depending on the answer, do they still work if we keep them?

CHANGELOG.md
colinrotherham: Had a chat with @anandamaryon1 yesterday and he's going to review too

Mentioned in https://github.com/nhsuk/nhsuk-frontend/pull/1521#pullrequestreview-3115512731 but do we have context on where the icon changes have come from?

Got a bit more info in https://github.com/nhsuk/nhsuk-frontend/pull/1026 and https://github.com/nhsuk/nhsuk-frontend/pull/1027 but let's create or link to an issue

packages/nhsuk-frontend/src/nhsuk/core/styles/_icons.scss
paulrobertlloyd: Yes, but the number doesn’t even relate to the icon size (perhaps it once did but not currently where the icon size is 34px; `nhsuk-icon--size-100` multiplies that number by 2, so the actual size is 68px). 🤪

colinrotherham: Ahh yeah, looks like the unit is "percentage increase"

```scss
.nhsuk-icon--size-25 {} // 25% bigger
.nhsuk-icon--size-50 {} // 50% bigger
.nhsuk-icon--size-75 {} // 75% bigger
.nhsuk-icon--size-100 {} // 100% bigger
```

There's reasonable usage over on the [NHSDigital organisation (search results)](https://github.com/search?q=org%3ANHSDigital+nhsuk-icon--size+language%3AHTML&type=code&l=HTML)

What do you think @anandamaryon1?

packages/nhsuk-frontend/src/nhsuk/core/styles/_icons.scss
```diff
@@ -1,60 +1,18 @@
 @use "../settings" as *;
+@use "../tools" as *;
```
colinrotherham, line 2: Do we need tools here?

packages/nhsuk-frontend/src/nhsuk/core/tools/_links.scss
```diff
@@ -125,10 +113,6 @@
   &:active,
   &:active:visited {
     color: $link-active-colour;
-
-    .nhsuk-icon {
-      fill: $link-active-colour;
-    }
```
colinrotherham: Ah sorry @paulrobertlloyd 

I hated adding these in https://github.com/nhsuk/nhsuk-frontend/commit/cf291cd649246bbb98c2aa198165f3fe4e656b52 but `fill: currentcolor` isn't applied for `:visited` links etc

Action links and cards are hard coded but pagination used these lines 😔

Pagination arrow, visited state

<br>
<br>

The mixins are repetitive but reliable

paulrobertlloyd: Is it just `:visited` that needs an explicit `fill`?

colinrotherham: Sadly not, think it's lots of things (potentially)

It was this [specification change](https://github.com/w3c/csswg-drafts/issues/5419) and issues with SVGs in https://github.com/w3c/csswg-drafts/issues/6310 and a comment here:

* https://github.com/alphagov/govuk-frontend/issues/2272#issuecomment-876577113

Looks like the `:visited` colour was related but not necessarily fixed

Could give a couple of solutions a try from this article:

**CurrentColor SVG in forced colors modes**
https://melanie-richards.com/blog/currentcolor-svg-hcm/

1. `fill: LinkText`
2. `forced-color-adjust: preserve-parent-color`

Otherwise might be best keeping our existing mixin "clobber the icon colour" approach for now, but I'm very aware lots of components aren't supporting forced colours well enough currently

packages/nhsuk-frontend/src/nhsuk/core/styles/_icons.scss
paulrobertlloyd: No, nixed.

packages/nhsuk-frontend/src/nhsuk/core/tools/_links.scss
paulrobertlloyd: Ah, sod it, let’s keep this PR scoped to replacing icons with the macro and ensuring that these icons scale correctly. Link styling is a whole other kettle of fish, which we will be returning to again soon. Reverted.

colinrotherham:
I've pushed a few updates

1. Added `rem` sizes by default + size modifier classes
2. Added support for `params.classes` on `nhsukIcon()`
3. Default to ~`__${name}`~ `--${name}` for BEM (backwards compatible)

Updated the [**All icon examples** preview page](https://nhsuk-frontend-pr-1521.herokuapp.com/nhsuk-frontend/examples/all-icons/) too:

Icon sizes

packages/nhsuk-frontend/src/nhsuk/core/styles/_icons.scss
colinrotherham: I've made sure they're all supported ([see comment](https://github.com/nhsuk/nhsuk-frontend/pull/1521#issuecomment-3191239980)) and include @paulrobertlloyd's icon scaling fixes

paulrobertlloyd:
Nice set of updates. Would we remove the older and incorrectly BEM’d icon class names in v11? Do we have a list somewhere of things to be removed, and has that list been updated?

packages/nhsuk-frontend/src/nhsuk/components/header/_index.scss
```diff
@@ -167,10 +167,11 @@ $_header-item-active-size: $nhsuk-focus-width;
   margin-top: nhsuk-spacing(3);
   overflow: hidden;
 
-  .nhsuk-icon__user {
+  .nhsuk-icon__user,
```
paulrobertlloyd: We don’t need a fallback class for the user icon, as it’s not been put into general release yet.

packages/nhsuk-frontend/src/nhsuk/components/header/_index.scss
```diff
@@ -246,9 +247,10 @@ $_header-item-active-size: $nhsuk-focus-width;
     margin: 0;
   }
 
-  .nhsuk-icon__search {
-    height: 28px;
-    width: 28px;
+  .nhsuk-icon__search,
```
paulrobertlloyd: Do we need to keep the fallback class for the search icon either, given we’re essentially saying all the markup for the header is changing in v10?

colinrotherham:
>Nice set of updates. Would we remove the older and incorrectly BEM’d icon class names in v11? Do we have a list somewhere of things to be removed, and has that list been updated?

We've got an issue open and I've added https://github.com/nhsuk/nhsuk-frontend/issues/1407#issuecomment-3191319216

Typically we'll just search for `@deprecated` but probably need a few others adding

Legacy support for point `pt` font sizes is another good one

packages/nhsuk-frontend/src/nhsuk/components/card/_index.scss
```diff
@@ -261,7 +261,8 @@ $card-border-hover-color: $card-border-hover-colour;
   }
 
   .nhsuk-action-link__link {
-    .nhsuk-icon__arrow-right-circle {
+    .nhsuk-icon__arrow-right-circle,
+    .nhsuk-icon--arrow-right-circle {
```
paulrobertlloyd: Can we get away with using `.nhsuk-icon` here, given there is only ever one icon (and heck, you may even want to change it)?

packages/nhsuk-frontend/src/nhsuk/components/action-link/_index.scss
```diff
@@ -5,70 +5,50 @@
...
-    color: $nhsuk-print-text-colour;
-
-    &:visited {
-      color: $nhsuk-print-text-colour;
-    }
-  }
-
-  .nhsuk-icon__arrow-right-circle {
+  .nhsuk-icon__arrow-right-circle,
+  .nhsuk-icon--arrow-right-circle {
```
paulrobertlloyd: Can we get away with using `.nhsuk-icon` here, given there is only ever one icon (and heck, you may even want to change it)?

packages/nhsuk-frontend/src/nhsuk/components/header/_index.scss
colinrotherham: Good point. Same with its "optical alignment" too

```scss
// Adjust optical alignment due to the handle appearing
// to push the magnifying glass circle to the top left
margin: 0 -2px -2px 0;
```

Should we patch this in the SVG icon instead?

packages/nhsuk-frontend/src/nhsuk/components/card/_index.scss
colinrotherham: 100% agree for any size or positioning

Could always do something like this:

```suggestion
    .nhsuk-icon[class*="arrow-right-circle"] {
```

packages/nhsuk-frontend/src/nhsuk/components/header/_index.scss
paulrobertlloyd: Changed to use only `.nhsuk-icon--user`

packages/nhsuk-frontend/src/nhsuk/components/header/_index.scss
paulrobertlloyd: Good idea, lets!

packages/nhsuk-frontend/src/nhsuk/components/header/_index.scss
colinrotherham: Brill. I suppose `.nhsuk-icon` works well here too? No modifier needed

i.e. Apply the styles to any icon within `.nhsuk-header__account`

packages/nhsuk-frontend/src/nhsuk/components/action-link/_index.scss
colinrotherham: Same as above, 100% agree for any size or positioning

But we don't want to give the other icons a green fill do we?

packages/nhsuk-frontend/src/nhsuk/components/header/_index.scss
paulrobertlloyd: Done!

packages/nhsuk-frontend/src/nhsuk/components/card/_index.scss
paulrobertlloyd: Maybe keep with `.nhsuk-icon` for now, and add such a restriction if needed down the road?

PR #1522, opened 2025-08-04
Use em helper for contents list tick placement

paulrobertlloyd:
## Description

Forgot we had `nhsuk-em` to help with this…

## Checklist

- [ ] CHANGELOG entry


packages/nhsuk-frontend/src/nhsuk/components/contents-list/_index.scss
```diff
@@ -28,7 +28,7 @@
     display: block;
     left: 0;
     position: absolute;
-    top: 0.75em;
+    top: nhsuk-em(12, $nhsuk-base-font-size);
```
colinrotherham: I think I prefer the units here, based on GOV.UK usage

```suggestion
    top: nhsuk-em(12px, $nhsuk-base-font-size);
```

<br>

Are we definitely sure `nhsuk-px-to-rem()` won't work?

paulrobertlloyd: `rem` is relative to the base font size which is always 16px, but we need it to be relative to the list item’s font size, which can be 16px or 19px, depending on the breakpoint.

colinrotherham: Yeah, as in should we set both default (mobile) and tablet sizes like we did before?

https://github.com/nhsuk/nhsuk-frontend/blob/v10.0.0-internal.2/packages/nhsuk-frontend/src/nhsuk/components/contents-list/_index.scss#L22-L32

Appreciate we were happy with 2x `nhsuk-px-to-rem()` calls before 😆

But no, I'm happy with `nhsuk-em()` though

paulrobertlloyd: Cool. Have added a line to the CHANGELOG too, happy to approve?

colinrotherham (approved):

PR #1524, opened 2025-08-06
Use CSS-generated chevron for header menu

paulrobertlloyd:
## Description

Depends on #1515. Partly supersedes #1521.

Uses the new chevron shape mixin to create the chevron icons used for the header mobile menu:

<p>Screenshot 2025-08-06 at 17 54 28</p>

We could also do the same for the primary card:

<p>Screenshot 2025-08-06 at 17 55 07</p>

> [!NOTE]
> The visual regression images don’t seem to render the chevron in the header as I’m seeing it displayed. Might want to check that it’s rendering properly in Chrome on different platforms to check if this is a puppeteer issue only.

## Checklist

- [x] CHANGELOG entry


anandamaryon1:
Header more icon shows a little to the top right for me. Feels like it needs moving down and left a couple pixels.

Comparing with current (top left):
image

anandamaryon1:
Card chevron appears slightly weak for me too, on a low density display (better on high density)
image

Having a play and: 

```
  // Chevron
  &::after {
    right: 6px;

    @include nhsuk-shape-chevron(right, $color_nhsuk-white, $font-size: 20);
  }
```

Seems a good balance between both for me.

low density
image

high density
image

paulrobertlloyd:
Have. increased the size of the primary card chevron to 22 (to keep with typography scale) and adjusted the position of the chevron in the header menu toggle button.

anandamaryon1:
> I'm happy with this @anandamaryon1
> 
> It fits in nicely with the breaking header changes we already have
> 
> But what do you want to do about primary cards?
> 
> Should we:
> 
> 1. Detect and/or hide `nhsuk-icon` if the breaking change is missed?
> 2. Make the icon macro change but apply it to the header only for now?

So to avoid having two icons if the `svg` is left in by accident? 

Like this (old markup in with this PR)
image

How would we do option 1? Throw an error if the `svg` is present in the component?

colinrotherham:
>How would we do option 1? Throw an error if the `svg` is present in the component?

Was just thinking a simple `display: none` but I suppose it would never be removed then

colinrotherham:
Mind rebasing this @paulrobertlloyd and I'll give it another look

colinrotherham:
As lovely and compact as this is, I'm actually really loving the soft rounded icon over in https://github.com/nhsuk/nhsuk-frontend/pull/1521 instead

paulrobertlloyd:
Rebased! Am not opposed to keeping the primary card using an inline SVG, by the way. I did wonder if we could use the same icon we use for action links (or for action links use the same chevron in a circle), but perhaps that’s a discussion for another day?

anandamaryon1:
Hmm, I see what you mean @colinrotherham. It's more noticeable in the card, and when comparing with the other icons, the improved rounding fits better. 

To move forward with this let's keep the svg version for the card and leave it with only the back link, breadcrumbs and header, where it's used smaller.

paulrobertlloyd:
Cool. I’ll update this PR to remove the changes relating to the primary card.

paulrobertlloyd:
This PR now only makes the change for the header component.

anandamaryon1 (approved):

PR #1536, opened 2025-08-14
Add support for id param to all components

edwardhorsford:
## Description
Adds support for `id` as a param to the majority of components.

Doing this because:
* I had a use case where I wanted to anchor to a section, and couldn't
* It's good to support as a standard thing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,10 @@ If you are not using Nunjucks macros, update your HTML markup to:
 
 You can now use the action link component on dark backgrounds by using the `.nhsuk-action-link--reverse` class. Added in [pull request #1542: Add reverse action link modifiers and styles](https://github.com/nhsuk/nhsuk-frontend/pull/1542).
 
+:wrench: **Fixes**
+
+- [#1536: Add support for id param to all components](https://github.com/nhsuk/nhsuk-frontend/pull/1536)
+
 ## 10.0.0-internal.3 - 15 August 2025
 
 :new: **New features**
--- a/packages/nhsuk-frontend/src/nhsuk/components/card/macro-options.mjs
+++ b/packages/nhsuk-frontend/src/nhsuk/components/card/macro-options.mjs
@@ -9,6 +9,11 @@ export const name = 'Card'
  * @satisfies {{ [param: string]: MacroParam }}
  */
 export const params = {
+  id: {
+    type: 'string',
+    required: false,
+    description: 'The ID of the card.'
+  },
   heading: {
     type: 'string',
     required: true,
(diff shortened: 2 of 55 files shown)


packages/nhsuk-frontend/src/nhsuk/components/card/template.njk
```diff
@@ -9,6 +9,7 @@
 {%- if params.feature %} nhsuk-card--feature{% endif %}
 {%- if params.topTask %} nhsuk-card--top-task{% endif %}
 {%- if params.classes %} {{ params.classes }}{% endif %}"
+{%- if params.id %} id="{{ params.id }}"{% endif %}
```
colinrotherham: Nice one @edwardhorsford, thanks for this 🙌

Noticed that GOV.UK Frontend almost always puts the ID first (before the class)

```html
<div id="jump-here" class="nhsuk-card">
```

Shall we do the same?

```patch
- <div class="nhsuk-card
+ <div {%- if params.id %} id="{{ params.id }}"{% endif %} class="nhsuk-card
```

CHANGELOG.md
colinrotherham, line 1: We use Markdown's double trailing space syntax in this file to preserve line breaks

Mind reverting the changes further down in this file? (Sorry)

packages/nhsuk-frontend/src/nhsuk/components/card/macro-options.mjs
```diff
@@ -104,6 +104,11 @@ export const params = {
     required: false,
     description: 'Classes to add to the card.'
   },
+  id: {
+    type: 'string',
+    required: false,
+    description: 'Classes to add to the card.'
```
colinrotherham: Mind adjusting this text for ID?

```suggestion
    description: 'The ID of the card.'
```

<br>

We should make it the first param too, for consistency when it's rendered to HTML

It's what we do for character count, text input, select and texture

packages/nhsuk-frontend/src/nhsuk/components/card/template.njk
edwardhorsford: Thanks, fixed.

colinrotherham:
Looks great to me @edwardhorsford

Seeing as that was so quick, shall we do the same to other components?

colinrotherham:
Thanks @edwardhorsford, conflict all sorted

Great work here. What do you want to do about these?

* Action link – Add the `id` to `<div class="nhsuk-action-link…`?
* Checkboxes – ~Stick with `idPrefix`~ Add the `id` alongside `idPrefix` like tabs?
* Radios – ~Stick with `idPrefix`~ Add the `id` alongside `idPrefix` like tabs?
* Fieldset – Add the `id` to `<fieldset class="nhsuk-fieldset…`?
* Tables – Add the `id` to `<table class="nhsuk-table…`?

colinrotherham:
I've completed all the other components @edwardhorsford

Once we apply this PR for character count we'll get ID support there too (otherwise the textarea gets it)

* https://github.com/alphagov/govuk-frontend/pull/4566

colinrotherham (approved):

PR #1556, opened 2025-09-01
File upload component

paulrobertlloyd:
## Description

One of the remaining components that can be found in the GOV.UK Design System but not in the NHS design system is the [file upload component](https://design-system.service.gov.uk/components/file-upload/). This is particularly useful for staff-facing systems. Indeed, Mavis is currently using the (un-enhanced) GOV.UK component for the many places we require teams to upload and import spreadsheets.

This PR largely copies wholesale the component from the GOV.UK Design System, but makes a few changes:

- it does not provide a `javascript` parameter; this makes sense for the GOV.UK Design System where this is a new feature that teams need to opt into. As this is a new component for us, our component can use the enhanced design by default (falling back to the simpler design if JavaScript is not available)
- it does not provide a `value` parameter; this is marked as deprecated in GOV.UK Frontend
- it changes the design to fit in with the NHS design system, and also makes some different choices about empty and selected states, more below.

Otherwise, the HTML markup is unchanged (baring the `nhsuk-*` namespace), as is the component’s JavaScript, beyond updating references to our modules, etc.

### Component states

Here is a matrix of states for the GOV.UK component; empty (and hover), dragged and file selected (and hover):

GOV.UK file upload states.

A lot of research was done for the design for this component, making sure there’s a distinction between file upload components that look empty and filled. However, I find the way various parts of the component change colour, seemingly in an unrelated way, to be a bit maddening. But that could just be me…

Anyway, I’ve add a go at adapting these states for our version of the component:

NHS.UK file upload states.

I’ve tried to be intentional with the use of colour; the component’s background (and drop zone) turns white when in an active state (hover or being dragged onto). When a file is selected, the status area behind the file name goes darker (using a rich grey so as to not look disabled), and the drop zone is filled in with a shade of grey too. I’ve also made sure the focus states work correctly.

### Other notes

By copying the component wholesale, I’ve also brought across related helpers and utilities to support translatable strings for components. We don’t use this for any other components yet (but should!)… we could split this work out into a separate PR, but that may hold up the release of this component.

## Checklist

> [!NOTE]
> Will add visual regression tests and a changelog entry once we are happy with the design. 

- [ ] CHANGELOG entry


anandamaryon1:
This is great, thanks @paulrobertlloyd 

I know what you mean about the many states and colours. I've had a play and have a suggestion to simplify it by:
- triggering the pseudo button states when you interact with the real button / dropzone / big box
- keeping the background white when a file is present

Popped in a PR so we can compare side by side. See here: https://github.com/nhsuk/nhsuk-frontend/pull/1559

anandamaryon1:
I also don't love how the way you remove an added file is by pressing cancel on the dialog but I suppose that's just how `<input type="file">` works. 

And, think I've found a styling bug, which applies to the gov version too. When `disabled` the focus styles are applied if you click on the label. If you concur, I'll raise it over on gov frontend.

paulrobertlloyd:
> And, think I've found a styling bug, which applies to the gov version too. When `disabled` the focus styles are applied if you click on the label. If you concur, I'll raise it over on gov frontend.

Oh yeah! We should fix it on ours and raise it on the GOV.UK version. Nice catch!

anandamaryon1:
> > And, think I've found a styling bug, which applies to the gov version too. When `disabled` the focus styles are applied if you click on the label. If you concur, I'll raise it over on gov frontend.
> 
> Oh yeah! We should fix it on ours and raise it on the GOV.UK version. Nice catch!

Pushed a potential fix. Happy for you to suggest a better / cleaner way to implement it.

packages/nhsuk-frontend/src/nhsuk/components/file-upload/template.njk
```diff
@@ -9,7 +9,7 @@
 {%- set describedBy = params.describedBy if params.describedBy else "" -%}
 {%- set id = params.id if params.id else params.name -%}
 
-<div class="nhsuk-form-group {%- if params.errorMessage %} nhsuk-form-group--error{% endif %} {%- if params.formGroup.classes %} {{ params.formGroup.classes }}{% endif %}"
+<div class="nhsuk-form-group {%- if params.errorMessage %} nhsuk-form-group--error{% endif %} {%- if params.formGroup.classes %} {{ params.formGroup.classes }}{% endif %} {%- if params.disabled %} nhsuk-file-upload--disabled{% endif %}"
```
colinrotherham: @anandamaryon1 @paulrobertlloyd Could we look into the [`disabled` observer code](https://github.com/nhsuk/nhsuk-frontend/blob/1be2dc176951859b1d68f4397b846da420daf13e/packages/nhsuk-frontend/src/nhsuk/components/file-upload/file-upload.mjs#L363-L397)?

Only because [we removed `.nhsuk-button--disabled`](https://github.com/nhsuk/nhsuk-frontend/pull/1075) in v10.0.0

Would be nice to fix this automatically

anandamaryon1: Thanks @colinrotherham I kinda knew there'd be a better way…

Are you able to have a go at this @paulrobertlloyd? Too much like real dev work for me 😵‍💫

paulrobertlloyd: I think this is working as it’s meant to @colinrotherham? It’s looking for `[disabled]` on the `input` element, and if it’s applied, adds a class to the drop zone `div`. Can’t see how else this would work?

packages/nhsuk-frontend/src/nhsuk/components/file-upload/template.njk
```diff
@@ -0,0 +1,85 @@
...
+      key: 'entered-drop-zone',
+      message: params.enteredDropZoneText
+    }) -}}
+    {{- nhsukI18nAttributes({
+      key: 'left-drop-zone',
+      message: params.leftDropZoneText
+    }) -}}
+  >
+  <input class="nhsuk-file-upload {%- if params.classes %} {{ params.classes }}{% endif %} {%- if params.errorMessage %} nhsuk-file-upload--error{% endif %}" id="{{ id }}" name="{{ params.name }}" type="file"
+  {%- if params.disabled %} disabled{% endif %}
```
paulrobertlloyd: @colinrotherham This is the relevant part where the Nunjucks `disabled` param gets set on the base `input`.

packages/nhsuk-frontend/src/nhsuk/components/file-upload/template.njk
colinrotherham: Sorry, as in can we use this approach to prevent the `:focus` styles being applied when disabled too?

packages/nhsuk-frontend/src/nhsuk/components/file-upload/_index.scss
```diff
@@ -0,0 +1,206 @@
+@use "sass:color";
+@use "sass:math";
+@use "../../core/settings" as *;
+@use "../../core/tools" as *;
+@use "../../core/helpers" as *;
```
colinrotherham, line 5: Probably want to copy what I've done in https://github.com/nhsuk/nhsuk-frontend/pull/1567

Making sure we forward all the nested component styles too:

```suggestion
@use "../../core/helpers" as *;
@forward "../error-message";
@forward "../hint";
@forward "../label";
```

packages/nhsuk-frontend/src/nhsuk/components/file-upload/_index.scss
```diff
@@ -0,0 +1,206 @@
...
+      color: inherit;
+      font: inherit;
+    }
+
+    &:focus {
+      // Use `box-shadow` to add border instead of changing `border-width`
+      // (which changes element size) and since `outline` is already used for
+      // the yellow focus state.
+      box-shadow: inset 0 0 0 4px $_button-border-colour;
+      outline: $nhsuk-focus-width solid $nhsuk-focus-colour;
```
colinrotherham: Should this be the focused input mixin?

```suggestion
      @include nhsuk-focused-input;
```

packages/nhsuk-frontend/src/nhsuk/components/file-upload/_index.scss
paulrobertlloyd: Done

packages/nhsuk-frontend/src/nhsuk/components/file-upload/_index.scss
paulrobertlloyd: Done. Also added for `:focus-within`.

packages/nhsuk-frontend/src/nhsuk/components/file-upload/template.njk
paulrobertlloyd: Had a quick Look at this, but for the life of me can’t work out how to make this work. Am a bit confused, because the observer is correctly toggling the `.nhsuk-file-upload--disabled` class, yet even when it is present, without it being set in the template, we get the styles applied when clicking the label.

paulrobertlloyd:
Have rebased `main` and tidied up the commits (would be lovely to have a least one PR added without hundreds of back-and-forth commits, grumble grumble).

frankieroberto:
@anandamaryon1:

> I also don't love how the way you remove an added file is by pressing cancel on the dialog but I suppose that's just how `<input type="file">` works.

That seems to work in Chrome (as a hidden feature), but not in Safari. Not much we can do about that though.

packages/nhsuk-frontend/src/nhsuk/components/file-upload/template.njk
```diff
@@ -0,0 +1,85 @@
...
+  {%- if params.multiple %} multiple{% endif %}
+  {%- if describedBy %} aria-describedby="{{ describedBy }}"{% endif %}
+  {{- nhsukAttributes(params.attributes) }}>
+  </div>
+{% if params.formGroup.afterInput %}
+  {{ params.formGroup.afterInput.html | safe | trim | indent(2) if params.formGroup.afterInput.html else params.formGroup.afterInput.text }}
+{% endif %}
```
colinrotherham, line 107: @paulrobertlloyd Are you alright reviewing https://github.com/nhsuk/nhsuk-frontend/pull/1561 so the before/after sections work?

packages/nhsuk-frontend/src/nhsuk/components/file-upload/file-upload.mjs
```diff
@@ -0,0 +1,492 @@
+import { closestAttributeValue } from '../../common/closest-attribute-value.mjs'
+import { ConfigurableComponent } from '../../configurable-component.mjs'
+import { ElementError } from '../../errors/index.mjs'
+import { I18n } from '../../i18n.mjs'
+
+/**
+ * File upload component
+ *
+ * @preserve
```
colinrotherham: We can drop this. There's a custom [GOV.UK Frontend Babel config](https://github.com/alphagov/govuk-frontend/pull/3958) to preserve or remove unnecessary comments

```suggestion
```

packages/nhsuk-frontend/src/nhsuk/components/file-upload/file-upload.mjs
```diff
@@ -0,0 +1,492 @@
...
+
+  /** @private */
+  i18n
+
+  /** @private */
+  id
+
+  /** @private */
+  $announcements
+
```
colinrotherham: We use a newer TypeScript version so can remove these

Only need optional class fields that aren't added in the constructor (it can infer the others itself)

```suggestion
```

packages/nhsuk-frontend/src/nhsuk/components/file-upload/file-upload.mjs
```diff
@@ -0,0 +1,492 @@
...
+  /**
+   * @private
```
colinrotherham: We don't set `@private` or `@internal`. Mainly because these things are never actually private unless we adopt the [private element syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_elements)

packages/nhsuk-frontend/src/nhsuk/components/file-upload/file-upload.mjs
```diff
@@ -0,0 +1,492 @@
...
+  /**
+   * @param {Element | null} $root - File input element
+   * @param {FileUploadConfig} [config] - File Upload config
+   */
+  constructor($root, config = {}) {
+    super($root, config)
+
+    const $input = this.$root.querySelector('input')
+
+    if ($input === null) {
```
colinrotherham: Interesting that GOV.UK Frontend didn't follow their usual format here to check for incorrect element types. If we can please use:

```suggestion
if (!($input instanceof HTMLInputElement)) {
```

packages/nhsuk-frontend/src/nhsuk/components/file-upload/file-upload.mjs
```diff
@@ -0,0 +1,492 @@
...
+      throw new ElementError({
+        component: FileUpload,
+        identifier: 'File inputs (`<input type="file">`)'
```
colinrotherham: Then can we pass these options in?

This makes sure `ElementError` automatically switches out the text for both unexpected components types and `null` (not found) values

I've used "Form field" because we're only querying for `input` at this point, we haven't expected a specific type attribute

```suggestion
        element: $input,
        expectedType: 'HTMLInputElement',
        identifier: 'Form field (`<input>`)',
```

packages/nhsuk-frontend/src/nhsuk/components/file-upload/file-upload.mjs
```diff
@@ -0,0 +1,492 @@
...
+      })
+    }
+
+    if ($input.type !== 'file') {
+      throw new ElementError({
+        component: FileUpload,
+        identifier:
+          'File input (`<input type="file">`) attribute (`type`) is not `file`'
```
colinrotherham: Thanks for using `ElementError` here instead of `formatErrorMessage()`

Watch out though because `element` hasn't been passed in so it'll automatically append "not found" to the error message. How about something like this so `expectedType` can still be used?

```suggestion
        element: $input,
        identifier: 'Form field (`<input>`)',
        expectedType: 'HTMLInputElement with type="file"',
```

colinrotherham: Or keeping the expected attribute in backticks?

```suggestion
        element: $input,
        identifier: 'Form field (`<input>`)',
        expectedType: 'HTMLInputElement with attribute (`type="file"`)',
```

packages/nhsuk-frontend/src/nhsuk/components/file-upload/file-upload.mjs
```diff
@@ -0,0 +1,492 @@
...
+    // Copy `aria-describedby` if present so hints and errors
+    // are associated to the `<button>`
+    const ariaDescribedBy = this.$input.getAttribute('aria-describedby')
+    if (ariaDescribedBy) {
+      $button.setAttribute('aria-describedby', ariaDescribedBy)
+    }
+
+    // Create status element that shows what/how many files are selected
+    const $status = document.createElement('span')
+    $status.className = 'nhsuk-body-m nhsuk-file-upload-button__status'
```
colinrotherham: Noticed we added `-m` here but aren't `nhsuk-body-m` and `nhsuk-body` essentially the same?

packages/nhsuk-frontend/src/nhsuk/components/file-upload/template.njk
colinrotherham: For completeness with GOV.UK Frontend shall we keep it in the template?

packages/nhsuk-frontend/src/nhsuk/components/file-upload/template.njk
```diff
@@ -0,0 +1,85 @@
+{% from "nhsuk/macros/attributes.njk" import nhsukAttributes %}
+{% from "nhsuk/macros/i18n.njk" import nhsukI18nAttributes %}
+{% from "nhsuk/components/error-message/macro.njk" import errorMessage %}
+{% from "nhsuk/components/hint/macro.njk" import hint %}
+{% from "nhsuk/components/label/macro.njk" import label %}
+
+{#- A record of other elements that we need to associate with the input using
```
colinrotherham: Minor thing. I've been copying from GOV.UK Frontend as-is (almost) in https://github.com/nhsuk/nhsuk-frontend/pull/1567 and resisting things like this. Do you want me to fix it in my PR too or change it here to match?

```suggestion
{#- a record of other elements that we need to associate with the input using
```

packages/nhsuk-frontend/src/nhsuk/components/file-upload/_index.scss
colinrotherham, line 1: @paulrobertlloyd I've done a [really rough diff](https://github.com/nhsuk/nhsuk-frontend/compare/file-upload...file-upload-diff) for comparison with GOV.UK Frontend

There's a mixture of:

1) Improvements to Sass structure
2) Dragging and disabled styles removed
3) Focus styles moved to mixins
4) Button styles adjusted to fit

Whilst 1) is great I think we should keep these lines as-is. For example, where CSS properties have been re-ordered it's hard to see what's the same versus what's changed.

Do we need any of the removed lines I can see for 2) [near the bottom](https://github.com/nhsuk/nhsuk-frontend/compare/file-upload...file-upload-diff)?

packages/nhsuk-frontend/src/nhsuk/components/file-upload/_index.scss
```diff
@@ -0,0 +1,207 @@
+@use "sass:color";
+@use "sass:math";
+@use "../../core/settings" as *;
+@use "../../core/tools" as *;
+@use "../../core/helpers" as *;
+@forward "../error-message";
+@forward "../hint";
+@forward "../label";
+
+@include nhsuk-exports("govuk/component/file-upload") {
```
colinrotherham:
```suggestion
@include nhsuk-exports("nhsuk/components/file-upload") {
```

packages/nhsuk-frontend/src/nhsuk/components/file-upload/_index.scss
colinrotherham: Ease of future maintenance and bug fixes are preferred

packages/nhsuk-frontend/src/nhsuk/components/file-upload/template.njk
paulrobertlloyd: Disagree. We should use capitalisation because we’re not animals. There are already changes in this template (such as the name spacing), so it’s never going to match 100%.

And if that means going through all our components to use capitalisation in comments, I’ll do it.

packages/nhsuk-frontend/src/nhsuk/components/file-upload/_index.scss
paulrobertlloyd: I think you’re fighting a losing battle. Honestly, at some point I’d like to run a linter on our SCSS files to have them use alphabetical order. Again with the capitalisation in comments (which is all over the place in the upstream component). And besides, we’re using different design/behaviour.

Agree that it makes it harder to pull in upstream changes… but while this project remains a distinct fork, that’s always going to be the risk.

packages/nhsuk-frontend/src/nhsuk/components/file-upload/template.njk
colinrotherham: Capitalisation was a valid option too!

>Do you want me to fix it in my PR too or change it here to match?

I'll fix it in https://github.com/nhsuk/nhsuk-frontend/pull/1567

packages/nhsuk-frontend/src/nhsuk/components/file-upload/_index.scss
colinrotherham: As much as A-Z ordering is lovely, adopting [GOV.UK Frontend property order](https://github.com/alphagov/govuk-frontend/blob/main/stylelint.config.js#L52-L58) makes it easier not harder

>Do we need any of the removed lines I can see for 2) [near the bottom](https://github.com/nhsuk/nhsuk-frontend/compare/file-upload...file-upload-diff)?

Can you talk us through the behaviour and non-formatting changes?

packages/nhsuk-frontend/src/nhsuk/components/file-upload/template.njk
paulrobertlloyd: Thank you! And sorry to be a grammar pedant 😊

packages/nhsuk-frontend/src/nhsuk/components/file-upload/_index.scss
paulrobertlloyd: Oh, I don’t mind as long as there’s an agreed order (I’ve come round to alphabetical as it’s easier to understand, but hey ho)… do we have this ordering in our  SCSS linting? We should!

I’ll need to refresh myself to the changes I made. Most of the changes relate to how we treat the pseudo button, and not changing the design of the component across so many different states, I think.

colinrotherham: Thanks @paulrobertlloyd

I'd like to do a review side by side and have all the differences commented

Other than the obvious ones that will always be fine:

* Adjustments for Sass modules
* Namespace as `nhsuk` not `govuk`
* Unnecessary font family or common link mixins
* Necessary NHS focus mixins rather than hard coded

anandamaryon1: In keeping with our higher-level goal of realigning with GOV.UK, we should where possible, defer to how they do things and only make changes where we have clear reason to deviate. Appreciate your strive for continuous  improvement though @paulrobertlloyd but I think sticking close to GOV.UK has higher-level benefits. 

If there's a way to have the linter auto-reorder CSS properties to follow the the rules, that'd be wonderful, or at least shout at us. I can never hold the order in my mind.

packages/nhsuk-frontend/src/nhsuk/components/file-upload/file-upload.mjs
paulrobertlloyd: I looked, and there is no `nhsuk-body` class in our CSS! It’s just that a bare `p` element will get the default `nhsuk-body-m` styles applied. Probably something to fix in another PR?

packages/nhsuk-frontend/src/nhsuk/components/file-upload/file-upload.mjs
paulrobertlloyd: Is support too restricted for us to adopt the private element syntax yet?

packages/nhsuk-frontend/src/nhsuk/components/file-upload/file-upload.mjs
paulrobertlloyd: Removed.

packages/nhsuk-frontend/src/nhsuk/components/file-upload/file-upload.mjs
paulrobertlloyd: Removed.

packages/nhsuk-frontend/src/nhsuk/components/file-upload/file-upload.mjs
paulrobertlloyd: Fixed.

packages/nhsuk-frontend/src/nhsuk/components/file-upload/file-upload.mjs
paulrobertlloyd: Fixed.

packages/nhsuk-frontend/src/nhsuk/components/file-upload/file-upload.mjs
paulrobertlloyd: Fixed.

packages/nhsuk-frontend/src/nhsuk/components/file-upload/_index.scss
paulrobertlloyd: Fixed.

packages/nhsuk-frontend/src/nhsuk/components/file-upload/file-upload.mjs
colinrotherham: True but use it in lots of examples (and sometimes style it just in case)

Just something I spotted in the diff. Seems like this should be `nhsuk-body` too

(Saves having to do a deprecation and breaking release one day)

packages/nhsuk-frontend/src/nhsuk/components/file-upload/_index.scss
paulrobertlloyd: On this, I have (begrudgingly) reinstated the upstream comments in the SCSS file to help better understand where our component differs. I’ve also updated the order of rules to follow those imposed by our linting rules.

However, it’s hard to have a nice diff between our component styles and those of the upstream component because:

- the upstream component’s rules are not (yet) ordered according to the linting rules
- we have a different secondary button design that requires management of a `::before` pseudo element which applies the button’s shadow
- we’ve chosen a different approach for indicating state (both in terms of colours, and also how the ‘pseudo’ button behaves across those states)

We can’t having near parity with the upstream component AND make different design decisions. Of the two, I know which I would pick 😉

packages/nhsuk-frontend/src/nhsuk/components/file-upload/file-upload.mjs
paulrobertlloyd: Updated to use `nhsuk-body`.

packages/nhsuk-frontend/src/nhsuk/components/file-upload/template.njk
paulrobertlloyd: Done.

packages/nhsuk-frontend/src/nhsuk/components/file-upload/file-upload.mjs
colinrotherham: Probably once [Babel 8](https://next.babeljs.io/docs/v8-migration/) is out and we replace `loose: true` with [compiler assumptions](https://next.babeljs.io/docs/assumptions) [`noUninitializedPrivateFieldAccess`](https://next.babeljs.io/docs/assumptions#nouninitializedprivatefieldaccess) and [`privateFieldsAsProperties`](https://next.babeljs.io/docs/assumptions#privatefieldsasproperties)

You can see it was tried in the past https://github.com/alphagov/govuk-frontend/pull/4099/commits/131a0874eb345f8659aef2fc235b4de5c7c334f2 but it did add a reasonable overhead to the compiled code so GOV.UK Frontend just pretends they're private methods

colinrotherham:
Thanks for all this @paulrobertlloyd

The init functions are all deprecated, but do you think we need an `initPasswordInputs()` like the other components?

packages/nhsuk-frontend/src/nhsuk/i18n.jsdom.test.mjs
colinrotherham, line 1: You can drop this commit, updated it in https://github.com/nhsuk/nhsuk-frontend/pull/1565/commits/20efac9e0cf58a7732f625ff2cc0fba157d06cec#diff-74d36de4fcae085fd0346d484c3174f90cc4f318ad24b63c4dd59029f3bdb8e2

paulrobertlloyd: Is this commit in a particular unmerged PR?

colinrotherham: Sorry, SHA has changed

Yeah it's in this one https://github.com/nhsuk/nhsuk-frontend/pull/1565

packages/nhsuk-frontend/src/nhsuk/core/styles/_typography.scss
```diff
@@ -149,6 +149,7 @@ h6,
 }
 
 p,
+.nhsuk-body,
```
paulrobertlloyd, line 152: @colinrotherham The file upload component adds this class to a `span` not `p`, so without adding this alias, the status message does not appear with the correct font size. 

GOV.UK Frontend also adds this alias, but without getting into the weeds, slightly differently (but for which we can address in a separate PR if and when we want to align how we manage out typographic styles).

packages/nhsuk-frontend/src/nhsuk/i18n.jsdom.test.mjs
paulrobertlloyd: Removed my commit in favour of that in the incoming PR.

colinrotherham:
Just flagging some recent PRs we'll want to keep on top of for this one:

* https://github.com/alphagov/govuk-frontend/pull/5311
* https://github.com/alphagov/govuk-frontend/pull/6223
* https://github.com/alphagov/govuk-frontend/pull/6304
* https://github.com/alphagov/govuk-frontend/pull/6337

sarawilcox:
Hi @paulrobertlloyd , just checking that you saw the issue for [file upload in the community backlog](https://github.com/nhsuk/nhsuk-service-manual-community-backlog/issues/93). In case there's anything there we should take into account. 

I'm going to prep a guidance page based on GOV's. Is there anything we want add to the guidance from an NHS point of view?

I'll share the page with you when I've drafted it.

paulrobertlloyd:
Hi @sarawilcox. The main challenge with using guidance for this component from the equivalent on the GOV.UK Design System is that they have the original component with guidance on how (and why) to opt-in to the new improved version. We don’t require that opt-in, we are using the improved design by default.

sarawilcox:
Thanks @paulrobertlloyd , yes, I've based our draft guidance on GOV's improved version. If you have a mo, would you mind reviewing the draft guidance, please? **https://github.com/nhsuk/nhsuk-service-manual/pull/2347**

sarawilcox:
@paulrobertlloyd @anandamaryon1 , have you seen this comment on the community backlog issue for file upload? 
https://github.com/nhsuk/nhsuk-service-manual-community-backlog/issues/93#issuecomment-3426691834

sarawilcox:
Hi, we've had a Dragon user report the following.

> I have a problem using it with Dragon, and this happens with all versions - the GOV.UK improved component, and also with both our default version and the multiple files version.

> I think the problem I'm having might be similar to that reported on the GOV.UK page ("users will first need to perform another action, such as a mouse click."). However, this problem happens the first time I try to use it, not on subsequent interactions.

> I say "choose file" and the component gets highlighted in yellow, but nothing happens. There's no pop-up file explorer window. If I click out of the component so my cursor and the focus is elsewhere, I can then say "choose file" a 2nd time, and this time the component gets highlighted in yellow and the file explorer pop-up does appear.

Is it something that might be fixable? If not, I'll note it in the guidance. 

GOV reports something similar on their guidance page: https://design-system.service.gov.uk/components/file-upload/#research-on-this-component

@anandamaryon1 for info.

colinrotherham:
Rebased with `main`

anandamaryon1:
I've done the browser and assistive technology testing. (excluding JAWS which I don't have access to right now.)

Just one finding:
On Safari (26.1) / VoiceOver the selected filename is not announced, and it instead says 'no file chosen'. 
image

I find the same on the GOV.UK version too.

anandamaryon1:
Some thoughts whilst testing:

1. Should we increase the padding to match other components / our spacing scale? (I think we should, 32px same as cards)
2. For the guidance but, should we explain that some browsers wont let you remove a chosen file
3. Would be a change from GOV, but we could consider changing the button text once a file is chosen. E.g Change/replace file. (most noticeable on screen reader, it reads the filename then says choose file, which gave me pause to think has my file been remembered)

packages/nhsuk-frontend/src/nhsuk/components/file-upload/template.njk
anandamaryon1: What do you mean "keep it in the template" @colinrotherham ?

Happy for you to rejig the code however you see fit.

anandamaryon1:
Sorry, I'll stop looking soon but, also notice that our `disabled` state doesn't have `cursor: not-allowed;`. Which GOV.UK's uses, and we use on checkboxes, but not buttons 🤔.

Maybe a separate PR later to align them?

colinrotherham:
* Rebased with `main` to add **fixtures.mjs**
* Updated [form control states](https://nhsuk-frontend-pr-1556.herokuapp.com/nhsuk-frontend/examples/form-controls-states/) to include file upload

paulrobertlloyd:
@colinrotherham Can we also show focused (and file picked) states for the File upload component on the form control states page?

colinrotherham:
We could do it via JavaScript?

I suppose we should have probably copied the password input approach of rendering the wrapper button (and contents) using Nunjucks

Yet keeping `hidden` set until ready to remove it

That way the fake `:focus` class could be set via `params.button.classes`

colinrotherham:
I've updated this PR to target [~`main`~](https://github.com/nhsuk/nhsuk-frontend/tree/main) [`support/10.x`](https://github.com/nhsuk/nhsuk-frontend/tree/support/10.x) ready for [v10.3.0](https://github.com/nhsuk/nhsuk-frontend/milestone/9)

colinrotherham:
Rebased to pick up the new status check names

paulrobertlloyd:
@anandamaryon1 JSDOM unit and Puppeteer behaviour tests added, and SonarQube code quality issues resolved. Beyond any last-minute design changes, this should be ready to merge.

packages/nhsuk-frontend/src/nhsuk/components/file-upload/template.njk
colinrotherham: Sorry all, I realised it's the opposite here

GOV.UK Frontend doesn't have `--disabled` in the template but we do

Can we remove it?

~The class doesn't do anything with JavaScript disabled (and JavaScript adds it automatically anyway)~

I'm now seeing we've got two disabled classes on the go, I'll give this some thought

colinrotherham: We've also got to make `:not(:disabled)` guards the `:focus`, `:hover` and `:active` states too

E.g. Labels can pass their `:hover` state onto an input causing this flicker even when disabled:

https://github.com/user-attachments/assets/448e1721-a815-45ee-b198-28728960924d

colinrotherham:
I've pushed some more commits:

1. [Uplift changes from GOV.UK Frontend v5.13.0](https://github.com/nhsuk/nhsuk-frontend/pull/1556/commits/4f579fe4192b4152abff4474307d8e605b74c85b)
2. [Fix label-to-input states triggering hover/focus/active when disabled](https://github.com/nhsuk/nhsuk-frontend/pull/1556/commits/41e265a9b76e600776e8224c411dab3fc05e9240)
3. [Move file upload module to form group](https://github.com/nhsuk/nhsuk-frontend/pull/1556/commits/8cbcf53d33efde2faf8d94879815270b362a8259)

Mainly to solve the issue [shown in the video](https://github.com/nhsuk/nhsuk-frontend/pull/1556#discussion_r2615139722) and remove the `--disabled` handling via Nunjucks

See the PR description for an overview of the HTML change

colinrotherham:
@paulrobertlloyd Ready for review again please 🙏

You'll see the [HTML `data-module` move](https://github.com/nhsuk/nhsuk-frontend/pull/1556/commits/8cbcf53d33efde2faf8d94879815270b362a8259) to the form group plus [other Friday changes](https://github.com/nhsuk/nhsuk-frontend/pull/1556#issuecomment-3647617060)

### Before

```html
<div class="nhsuk-form-group">
  <div class="nhsuk-drop-zone" data-module="nhsuk-file-upload"></div>
</div>
```

### After

```html
<div class="nhsuk-form-group nhsuk-file-upload" data-module="nhsuk-file-upload">
  <div class="nhsuk-drop-zone"></div>
</div>
```

I’m wondering if we should now follow BEM now and rename:

* `nhsuk-drop-zone` → `nhsuk-file-upload__drop-zone`
* `nhsuk-file-upload-button` → `nhsuk-file-upload__button`

Over in GOV.UK Frontend they'd already shipped `nhsuk-file-upload` so couldn't follow the same parent component model used by character count and password input etc—but we don't have the same restriction

paulrobertlloyd:
Love it! But also a little torn as I know my team will be using the GOV.UK components gem, so these differences will need a little massaging. I wonder if there’s an opportunity to make a PR against GOV.UK Frontend given they’ll have the JS enhanced version as the default design in the upcoming 6.0 breaking release?

colinrotherham:
>Love it! But also a little torn as I know my team will be using the GOV.UK components gem, so these differences will need a little massaging. I wonder if there’s an opportunity to make a PR against GOV.UK Frontend given they’ll have the JS enhanced version as the default design in the upcoming 6.0 breaking release?

Shall we keep the old classes and I'll harden the checks to skip creating a drop zone etc if it already exists?

We can keep both HTML structures compatible

colinrotherham:
>I wonder if there’s an opportunity to make a PR against GOV.UK Frontend given they’ll have the JS enhanced version as the default design in the upcoming 6.0 breaking release?

Will leave this one to you 😊

colinrotherham:
I've rebased this with https://github.com/nhsuk/nhsuk-frontend/pull/1744

You’ll see I’ve moved all CSS class selectors into the config, allowing for:

```mjs
initAll({
  fileUpload: {
    dropZoneClass: 'govuk-drop-zone'
  }
})
```

colinrotherham:
I've pushed three new commits to highlight:

* [Align file upload properties, i18n keys and CSS classes](https://github.com/nhsuk/nhsuk-frontend/pull/1556/changes/6a86235d4c4b30969662c7cd8bc4e79d58a4421c)
* [Make sure button styles are forwarded](https://github.com/nhsuk/nhsuk-frontend/pull/1556/changes/d98f1f6cdc0f22faffdae9c6d3b33da1d83f1f77)
* [Improve file upload button instruction spacing](https://github.com/nhsuk/nhsuk-frontend/pull/1556/changes/3854f41a1f442e2d65a8774b828154d52a539a93)

I've used our `$nhsuk-button-padding-left-right` variable to balance margin betwen the instruction text

You'll see everything is still configurable but we now use the same naming conventions for:

### CSS classes

```scss
.nhsuk-file-upload__drop-zone
.nhsuk-file-upload__drop-button
.nhsuk-file-upload__choose-files-button
```

### JavaScript properties

```mjs
this.$dropZone
this.$dropButton
this.$chooseFilesButton
```

### i18n keys

```mjs
this.i18n.t('dropInstruction')
this.i18n.t('chooseFilesButton')
```

packages/nhsuk-frontend/src/nhsuk/components/file-upload/_index.scss
```diff
@@ -60,25 +60,26 @@
   // required because disabling pointer events
   // on the button means that the cursor style
   // be applied on the button itself
-  .nhsuk-drop-zone--disabled,
-  .nhsuk-file-upload--disabled .nhsuk-drop-zone {
+  .nhsuk-file-upload--disabled .nhsuk-file-upload__drop-zone,
+   // Deprecated, to be removed in a future release
+  .nhsuk-file-upload__drop-zone--disabled {
```
paulrobertlloyd, line 67: Why do we need this deprecated class if the component has not been released yet?

colinrotherham: Similar to `initFileUploads()` (and the existing init functions) it's supported but immediately deprecated

When disabled, we append `--disabled` to the `$root` element class

* When using HTML from GOV.UK that'll be the drop zone
* When using HTML from NHS.UK that'll be the form group

Probably need you to decide if we keep support to [optionally locate existing drop zone](https://github.com/nhsuk/nhsuk-frontend/pull/1556/changes/0918d420a3b91a9a7b6ac2207d2b2c7939652d81)

paulrobertlloyd: I can see adding immediately deprecated features so that new APIs match existing ones, but this seems like a different case, no?

Let’s keep the robustness of that drop zone approach; not only does it provide compatibility with GOV.UK markup, but also means the JS will work with slightly malformed markup, which I can well imagine being used.

colinrotherham: Ah, as in you want to keep it long term?

I'd assumed we'll be phasing out compatibility with GOV.UK markup

paulrobertlloyd: Does it hurt to leave it in? Or maybe we do for as long as GOV.UK frontend uses the older placement? And if that is the case, maybe we should add a comment to that effect?

colinrotherham: I'm happy leaving it in for now, but it's why I put "in a future release" instead of v11

It's a moving target trying to maintain cross compatibility

colinrotherham (approved):

PR #1021, opened 2024-09-18
Update changelog with description of date input changes

frankieroberto:
Updates the changelog to describe the breaking changes in [Make it easier to set date values](https://github.com/nhsuk/nhsuk-frontend/pull/994].

PR #1570, opened 2025-09-10
Fix description of namePrefix for date input

frankieroberto:
The description of how the `namePrefix` is out of date, as in #994 we updated it to use square brackets instead of hyphens, for easier use with Express.js based servers (eg the NHS Prototype kit).

### DIFF
--- a/packages/nhsuk-frontend/src/nhsuk/components/date-input/macro-options.mjs
+++ b/packages/nhsuk-frontend/src/nhsuk/components/date-input/macro-options.mjs
@@ -18,7 +18,7 @@ export const params = {
     type: 'string',
     required: false,
     description:
-      "Optional prefix. This is used to prefix each `item.name` using `'-'`."
+      "Optional prefix. This is used to prefix each `item.name`, with the name then appearing in square brackets, for example `'date[month]'` ."
   },
   items: {
     type: 'array',


packages/nhsuk-frontend/src/nhsuk/components/date-input/macro-options.mjs
```diff
@@ -18,7 +18,7 @@ export const params = {
     type: 'string',
     required: false,
     description:
-      "Optional prefix. This is used to prefix each `item.name` using `'-'`."
+      "Optional prefix. This is used to prefix each `item.name`, with the name then appearing in square brackets, for example `'date[month]'` ."
```
colinrotherham: Do you think it needs spelling out clearly?

```suggestion
      "Optional prefix. This is used to prefix each `item.name`, with the name then appearing in square brackets, for example `namePrefix[item.name]` ."
```

```suggestion
      "Optional prefix. This is used to prefix each `item.name`, with the name then appearing in square brackets, for example `namePrefix + "[" + item.name + "]"` ."
```

frankieroberto: I think that could be even more confusing? Having seen people using the components trying to interpret these descriptions, I wonder if we should make them clearer to people using the components (eg in prototyping) and less developer-focused?

Maybe something like:

```suggestion
      "Optional. This is used to in the name attribute to group the date input parts together. For example, if the `namePrefix` is `dateOfBirth` the day input will have the `name` attribute `dateOfBirth[day]`, the month input will have the `name` attribute `dateOfBirth[month]` and so on."
```

frankieroberto:
@colinrotherham @anandamaryon1 I've updated this description again, whaddya think?

packages/nhsuk-frontend/src/nhsuk/components/date-input/macro-options.mjs
```diff
@@ -18,7 +18,7 @@ export const params = {
     type: 'string',
     required: false,
     description:
-      "Optional prefix. This is used to prefix each `item.name` using `'-'`."
+      "Optional. This is used to in the name attribute to group the date input parts together. For example, if the `namePrefix` is `dateOfBirth` the day input will have the `name` attribute `dateOfBirth[day]`, the month input will have the `name` attribute `dateOfBirth[month]` and so on."
```
colinrotherham: I'm still worried that this option should be called `name` as it's no longer a prefix

Whilst I'm happy with our decision to change behaviour from a hyphen-separated prefix to a "name" field with nested object values, we unintentionally kept the old name and description.

For context GOV.UK Frontend moved from `name` to `namePrefix` for the reverse reason:

* https://github.com/alphagov/govuk-frontend/issues/903
* https://github.com/alphagov/govuk-frontend/issues/909

We need more than wording really 😔

Our options seem to be:

* Move from `namePrefix` to `name` (breaking change)
* Support both `namePrefix` and `name` (breaking change)

If we keep the `namePrefix` nested object behaviour we're out of step with Vue/React too

frankieroberto: I kinda think `name` would also be slightly misleading and confusing, as the individual date parts also have a `name` param, and the 2 get combined together? 🤷 

Any suggestions on how the wording could be any clearer (regardless of what the param name is)?

colinrotherham: Sorry I don't know how to solve this 😬

>I kinda think `name` would also be slightly misleading and confusing…

Yeah it would be. It's hard to explain why this component doesn't hyphen-separate like the others

Do we have an issue for why we can't do this without losing all the defaults?

```json5
items: [
  { name: "date[day]" },
  { name: "date[month]" },
  { name: "date[year]" }
]
```

Maybe it should be a flag like `isPageHeading: true`

Can you raise it with @sarawilcox and @anandamaryon1?

frankieroberto: I don’t think we have any other components which use hyphen-separation for input names? Date is the only form component with nested components so far?

There’s no universal convention for this - Rails uses parentheses like this: `date_of_birth(1i)`, `date_of_birth(1i)`. Suspect every framework does it a bit differently! We’ve chosen to optimise for how Express.js does it.

colinrotherham: Sorry, that's me not being clear

From a Nunjucks API point of view the starting point was `idPrefix` when [clarifying the name field](https://github.com/alphagov/govuk-frontend/issues/903)

* Checkboxes `idPrefix`
* Tabs `idPrefix`
* Radios `idPrefix`
* Task list `idPrefix`

>…We’ve chosen to optimise for how Express.js does it.

Yeah absolutely, yet we went in a different direction and kept the naming convention for hyphenation

frankieroberto: I still think `namePrefix` is fine, so long as the description is clear - even if it's a different way of prefixing from `idPrefix`? 🤷

Not sure it’s worth a breaking change to change it to something else - would be a lot of work to update prototypes for no real gain.

Maybe aliasing it as `name` or `nameGroup` or something would be ok?

colinrotherham: It would be sensible for all form fields to support a parent/group name, so date inputs aren't the odd one out

Like `address[line1]` and `address[line1]`

I'd happily contribute prototype kit middleware to make any breaking changes backwards compatible

Especially if it means all NHS.UK frontend ports/libraries could eventually sync up

frankieroberto: The address pattern doesn't feel like the same thing to me? Those are separate fields which are just conceptually related (or could be, depending on context). Whereas the date input feels more like a single conceptual thing? 

In any case, you can already do `name: "address[line1]" pretty easily. Whereas for date inputs in govuk-frontend you have to re-specify all of the items, including the classes, which is super painful.

I think the `name` field could be one area where we have to accept where it makes sense for different ports/libraries to deviate based upon their own conventions - eg this is what the Rails port of govuk-frontend does: https://govuk-form-builder.netlify.app/form-elements/date-input/

colinrotherham: If it wasn't for the prototype kit, would you go for consistency instead?

frankieroberto: I don't think so - happy for this to be something that the various ports/framework diverge on, personally. 

Also think it can be quite helpful if the `id` and `name` attributes behave a bit differently, to help differentiate them. eg we generally use camelCase for `name` but hyphens for `id`.

colinrotherham: I'm leaning towards this idea of yours:

>Maybe aliasing it as `name` or `nameGroup` or something would be ok?

But similar to `value` and `values` I'm wondering if we need to explore `name` and `names`?

frankieroberto: Before we make any decision on aliasing the param name, shall we get the existing description updated so that it’s more accurate and a bit clearer? Then aliasing could be a follow-up task in the backlog?

sarawilcox (commented):
Hi @colinrotherham , is there a typo here? "This is used to in the name attribute to group ..."

colinrotherham:
@sarawilcox Yeah there is, do you mind [catching up with the review thread](https://github.com/nhsuk/nhsuk-frontend/pull/1570#pullrequestreview-3243991955)?

sarawilcox:
I'm afraid I'm not understanding this very well, @colinrotherham . Maybe we can discuss with @anandamaryon1 at PR review meeting tomorrow.

frankieroberto:
@sarawilcox I’ve fixed the typo 🤦, thanks.

The short story is that in #994 we (well I) updated the date input component so that the name attributes use the format `dateOfBirth[day]` instead of `dateOfBirth-day` as part of a breaking change release, in order to make them a lot easier to use in the prototype kit - but I forgot to update the description in the Nunjucks macro options.

frankieroberto:
@anandamaryon1 @colinrotherham could you check this correction to the `namePrefix` description makes sense to you, and if so approve it?

I can stick an issue on the backlog for a future debate about whether to rename/alias it from `namePrefix` or not.

I don’t think this needs a changelog entry as it’s purely a documentation wording change, but happy to add one if you think otherwise?

packages/nhsuk-frontend/src/nhsuk/components/date-input/macro-options.mjs
```diff
@@ -18,7 +18,7 @@ export const params = {
     type: 'string',
     required: false,
     description:
-      'Optional prefix. This is used to prefix each item `name`, separated by `-`.'
+      'Optional. This is used in the name attribute to group the date input parts together. For example, if the `namePrefix` is `dateOfBirth` the day input will have the `name` attribute `dateOfBirth[day]`, the month input will have the `name` attribute `dateOfBirth[month]` and so on.'
```
colinrotherham: If we embrace "prefix" but use the [`dob` example from the changelog](https://github.com/nhsuk/nhsuk-frontend/pull/1021)?

```suggestion
      'Optional prefix. This is used to prefix each date input `name` attribute, wrapped in `[` and `]`. For example, setting `namePrefix: "dob"` will output the `name` attributes `dob[day]`, `dob[month]` and `dob[year]` respectively.'
```

<br>

If you can open a GitHub issue to (as a suggestion) make this optional in future though. Due to [RFC3986](https://www.ietf.org/rfc/rfc3986.txt) I'm still not sure how many web servers and proxies allow or understand square brackets when not encoded (e.g. [Chrome GET requests](https://issues.chromium.org/issues/40530289), [Apache Commons HttpClient](https://stackoverflow.com/questions/40568/are-square-brackets-permitted-in-urls), [Apache Tomcat](https://stackoverflow.com/questions/10521274/square-brackets-in-url-causes-exception-in-tomcat), [IBM webMethods](https://community.ibm.com/community/user/discussion/error-creating-rest-api-with-encoded-square-brackets-in-jsonapi-query-parameters), [Nginx](https://serverfault.com/questions/407488/nginx-reverse-proxy-replaces-square-brackets-in-request))

> A host identified by an Internet Protocol literal address, version 6
   [RFC3513] or later, is distinguished by enclosing the IP literal
   within square brackets ("[" and "]").  This is the only place where
   square bracket characters are allowed in the URI syntax.

Otherwise we should look into encoding `[` → `%5B` for better compatibility

colinrotherham: It's a lovely feature that works in Express.js, PHP and Rails though!

frankieroberto: @colinrotherham have committed your suggestion, thanks! 

I’ve not observed any issues with non-escaped `[` in the querystring or POSTed form data, but we can encode if needed. I’ll open a separate issue.

frankieroberto:
@colinrotherham created the issue at #1674.

Ok to approve this now?

colinrotherham (approved):

PR #1596, opened 2025-09-16
Split out GitHub Action linting and test tasks

colinrotherham:
## Description

This PR runs our linting and test tasks concurrently to clearly show which task failed

GitHub Actions concurrent tasks

## Checklist

- [ ] CHANGELOG entry

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -1,4 +1,4 @@
-name: Pull request
+name: Checks
 
 on:
   pull_request:


frankieroberto:
@colinrotherham does this mean that the individual code style tests now come under the 'Pull request / Build' task? 

Screenshot 2025-09-17 at 10 28 48

That seems a bit less visible than having them all reported as separate tasks?  Also not sure why the `Pull request /` prefix is being added?

colinrotherham:
If you scroll down a bit you'll see a separate entry per task, e.g. the JavaScript test projects:

JavaScript test projects split out

<br>
<br>

You can ignore the Code style checks "waiting for status to be reported" etc. They're just placeholders for the currently required status checks and they can be turned off or changed

Yeah the `Pull request /` prefix is annoying—any ideas?

frankieroberto:
@colinrotherham I think the `/` might be a consequence of using the matrix? But the docs aren't super clear.

colinrotherham:
> @colinrotherham I think the / might be a consequence of using the matrix? But the docs aren't super clear.

@frankieroberto But wondered if you had any ideas? I've grown to accept it 😔

colinrotherham:
@frankieroberto Shall we get this merged?

I can rename the status checks to match the new names

frankieroberto:
@colinrotherham nice, ➕ for running the tests concurrently and reporting the status separately!

Could we also change the name of the workflow from `Pull request` to `Tests`, and then the tests would be reported like this:

```
Tests / JavaScript behaviour
Tests / JavaScript component
Tests / JavaScript unit
Tests / Accessibility
```

?

colinrotherham:
Mind if we go with "Checks" instead?

With "Linting" and "Tests" being two of the checks

colinrotherham:
@frankieroberto Done

frankieroberto:
@colinrotherham looks good!

Screenshot 2025-12-09 at 13 09 36

The "Checks" is a bit redundant as the header above the list already refers to them as "checks", but it’s better than "Pull request" I think?

Possibly ideally we could group them sub-categories, something like this:

```
Tests / JavaScript behaviour
Tests / JavaScript component
Tests / JavaScript unit
Tests / Accessibility
Code style / CSS
Code style / JavaScript
Analysis / Sonar
```

...but that’d mean splitting the workflow file up into separate files per group, I think? (Might still be worth doing, but a bigger task - could be done later).

Happy to approve!

colinrotherham:
>...but that’d mean splitting the workflow file up into separate files per group, I think? (Might still be worth doing, but a bigger task - could be done later).

We have a separate file for the Sass checks? But notice the output is still:

```console
Checks / Sass import / Dart Sass 1.80.0 (pull_request)
Checks / Sass import / Dart Sass latest (pull_request)
```

Yeah we can only do what's supported 😔

PR #1590, opened 2025-09-15
Add headings and back links to custom examples

colinrotherham:
## Description

This PR makes sure all custom examples have headings and back links for easy navigation

I've also reviewed and included these changes:

* Remove unnecessary Nunjucks imports
* Use consistent Nunjucks double quotes
* Prefer Nunjucks extends at top
* Consolidate layouts for custom example pages

Screenshot 2025-09-15 at 17 08 16

## Checklist

- [ ] Tested against our [testing policy](https://github.com/nhsuk/nhsuk-frontend/blob/master/docs/contributing/testing.md) (Resolution, Browser & Accessibility)
- [x] Follows our [coding standards and style guide](https://github.com/nhsuk/nhsuk-frontend/blob/master/docs/contributing/coding-standards.md)
- [ ] CHANGELOG entry

PR #1597, opened 2025-09-16
Split out component examples into separate pages, use consistent Nunjucks double quotes etc

colinrotherham:
## Description

This PR follows https://github.com/nhsuk/nhsuk-frontend/pull/1590 and splits out separate pages for each component

I've given each form component example a quick audit, so they now show:

1. Page heading by default
2. Label or legend `--l` (size L) modifier
3. Separate "without page heading" variant

The default examples are now the good ones

**Note:** We can make further improvements once these PRs are complete:

* https://github.com/nhsuk/nhsuk-service-manual/pull/2248
* https://github.com/nhsuk/nhsuk-service-manual/pull/2274
* https://github.com/nhsuk/nhsuk-service-manual/pull/2287

<br>

Home page with component index

<br>
<br>

Component preview pages

## Checklist

- [ ] CHANGELOG entry


packages/nhsuk-frontend/src/nhsuk/components/button/macro-options.mjs
```diff
@@ -154,7 +154,17 @@ export const examples = {
     }
   },
   'reverse': {
-    layout: 'layouts/example-background-blue.njk',
+    pageContext: {
```
paulrobertlloyd: Could we not supply a variable such as `pageContext.reverse` and if it’s `true`, swap the layout? Seems repetitive to set these values on multiple occasions.

packages/nhsuk-frontend-review/src/layouts/listing.njk
```diff
@@ -0,0 +1,71 @@
...
+              <a {{- nhsukAttributes({
+                href: exampleName | slugify ~ "/",
+                class: "nhsuk-link nhsuk-link--no-visited-state" ~ (" " ~ previewLink.classes if previewLink.classes else ""),
+                rel: "noreferrer noopener",
+                target: "_blank"
+              }) -}}>
+                Open this example in a new tab<span class="nhsuk-u-visually-hidden">: {{ name ~ " " ~ exampleName }}</span>
+              </a>
+            </p>
+            {{ render(component, example) if not isFullWidth }}
```
paulrobertlloyd: Not sure where `render` is coming from, but could you not supply `loop.index` to this, and then use that value for a component’s `attributes.id` option?

packages/nhsuk-frontend/src/nhsuk/components/button/macro-options.mjs
colinrotherham: Yeah! This is just me trying to find options that work via both the page template and "on page"

More than happy to narrow this down

packages/nhsuk-frontend-review/src/layouts/listing.njk
colinrotherham: Yeah good shout. What about [other things HTML validate flagged up](https://github.com/nhsuk/nhsuk-frontend/pull/1597/commits/965efd1764485c25859a48fa1b109be66efff216)?

It's `components.render()` from our **macro-options.mjs**

I added it [via `env.addGlobal()`](https://github.com/nhsuk/nhsuk-frontend/blob/965efd1764485c25859a48fa1b109be66efff216/packages/nhsuk-frontend-review/tasks/html.mjs#L43-L45) but it typically [renders components inside other components](https://github.com/nhsuk/nhsuk-frontend/blob/v10.0.0/packages/nhsuk-frontend/src/nhsuk/components/fieldset/macro-options.mjs#L163-L190)

packages/nhsuk-frontend-review/src/layouts/listing.njk
```diff
@@ -0,0 +1,71 @@
+{% extends "layouts/page-full-width.njk" %}
```
paulrobertlloyd, line 1: Perhaps we can be explicit, and have a context variable to override the preview layout, else use the default?

```suggestion
{% set previewLayoutName = previewLayoutName | default("page-full-width") %}
{% extends "layouts/" + previewLayoutName + ".njk" %}
```

packages/nhsuk-frontend-review/src/layouts/listing.njk
paulrobertlloyd: Be good to handle duplicate form field names (pass `loop.index` to `idPrefix` etc.?), not sure I’d worry too much about duplicate landmarks, though.

packages/nhsuk-frontend/src/nhsuk/components/button/macro-options.mjs
colinrotherham: ~Added `flags: { reverse: true }`~

Added an options object to macro option examples (as needed)

```js
options: {
  isFullWidth: true,
  isReverse: true,
  isHidden: true,
  layout: 'example-one-third'
}
```

packages/nhsuk-frontend-review/src/layouts/listing.njk
colinrotherham: Sadly there are just too many (including hard-coded HTML in conditionals)

How this solution in https://github.com/nhsuk/nhsuk-frontend/pull/1597/commits/245e81f6f75e71e12f2ca4f05f381366539c3f0e?

shared/lib/nunjucks/globals/render-example.mjs
```diff
@@ -0,0 +1,60 @@
...
+const nanoId = customAlphabet(lowercase, 10)
+
+/**
+ * Render Nunjucks component with unique attributes
+ *
+ * @param {string} name - Nunjucks component name
+ * @param {MacroExample} example - Nunjucks macro example
+ */
+export function renderExample(name, example) {
+  const uniqueId = nanoId()
```
paulrobertlloyd: In the spirit of trying to keep the number of dependencies low (which we could probably be a lot better at doing), how about using a built-in?

```suggestion
import { randomUUID } from 'node:crypto'
import { components } from '@nhsuk/frontend-lib'
import filters from 'nunjucks/src/filters.js'

/**
 * Render Nunjucks component with unique attributes
 *
 * @param {string} name - Nunjucks component name
 * @param {MacroExample} example - Nunjucks macro example
 */
export function renderExample(name, example) {
  const uniqueId = randomUUID()
```

colinrotherham: You're right. After all this I'm thinking of just appending the loop index 😆

Glad we got there together

paulrobertlloyd (approved):

packages/nhsuk-frontend-review/src/layouts/listing.njk
colinrotherham: Loved this suggestion so I've done it in https://github.com/nhsuk/nhsuk-frontend/pull/1597/commits/4c804eab3fe805a7ad4702c54333a5469d4b54ed

* `options.layout` is the Nunjucks layout AND `app-wrapper--${layout}` modifier
* `options.width` is the grid layout `nhsuk-grid-column-${width}` column width

```json5
options: {
  hidden: true,
  layout: 'background-blue',
  width: 'one-third'
}
```

Makes it nice to see the width overrides alongside others:

Card examples with width overrides

paulrobertlloyd (approved):
This is _sooooo_ much better! These pages will be a fantastic help when adding and updating components; no more opening and switching between multiple tabs. :shipit:

PR #1574, opened 2025-09-12
Uplift GOV.UK Frontend password input component

colinrotherham:
## Description

This PR adds the password input component from GOV.UK Frontend

Mostly to demonstrate `afterInput` in https://github.com/nhsuk/nhsuk-frontend/pull/1561 but using small buttons from https://github.com/nhsuk/nhsuk-frontend/pull/1643

Password input tablet

## Checklist

- [x] Tested against our [testing policy](https://github.com/nhsuk/nhsuk-frontend/blob/master/docs/contributing/testing.md) (Resolution, Browser & Accessibility)
- [x] Follows our [coding standards and style guide](https://github.com/nhsuk/nhsuk-frontend/blob/master/docs/contributing/coding-standards.md)
- [x] CHANGELOG entry

PR #1600, opened 2025-09-20
Refactor buttons, add small button variant

paulrobertlloyd:
## Description

This PR refactors how different button variations are managed via a new `nhsuk-button-style()` mixin.

It also adds a new `nhsuk-button--small` variant, which can be used alongside other small form controls, and which matches the height of text inputs, allowing it to be shown inline (required for the uplifted Password input component, see #1574).

### Refactoring button construction

We currently use a button’s `::after` pseudo element to manage the shadow on the secondary button variant. This means secondary buttons have a different construction to ‘normal’ buttons, and requires setting and unsetting values for different states.

This PR changes this so that all buttons share the same underlying construction: a border and box-shadow which share the same colour.

The `::after` is then used to provide a duplicate border that sits atop the button. For outline buttons, this shares the same design (the border matches the colour of the shadow), but for solid buttons, it’s border colour is changed to match the background colour of the button (thus hiding the button’s border) while its bottom position is adjusted so that the bottom border (aka shadow) of the button below is not hidden.

Perhaps this is best demonstrated with an image:

Button construction.

Colour variations are then managed via a new `nhsuk-button-style()` mixin. This is used to manage how colours are applied, and toggle secondary button behaviour.

This mixin can be used by authors downstream to create their own button styles if needed. Mavis has a secondary warning button variant; I’m unsure whether that should be provided as standard, but if not, it would now be much easier to add. 

### Small button variant

The above makes creating and managing additional button variations much easier, not least adding a new `nhsuk-button-small` variant:

Screenshot of small buttons.

The height of this smaller button is 40px (the standard size being 56px), so that it matches the height of input elements. This is useful when a button needs to sit next to an input, such as for the forthcoming Password input component, but also search forms, or forms that use the forthcoming small checkbox and radio components.

## Checklist

- [ ] CHANGELOG entry


colinrotherham:
Yes @paulrobertlloyd love it 💪

How's it looking with this old issue in mind?

* https://github.com/nhsuk/nhsuk-frontend/issues/1357

Am I seeing all the button screenshots slightly shorter in height?

paulrobertlloyd:
> How's it looking with this old issue in mind?

Haven’t tested (not sure I can test given the environment mentioned in that thread), but not sure it should occur as we are laying a border over a solid background.

There are some artefacts showing in the visual regression tests for the focused button state, which I specifically tried to counter in my code… so that might be something that needs refinement (lower resolution screens are a common factor, I suspect).

> Am I seeing all the button screenshots slightly shorter in height?

Correct, probably due to the shadow size now being 2px (although, thanks to the bottom border of the button, visually remains 4px). Worth noting also that as a consequence of this, the active deprecation is 2px, not 4px. This is also helpful for inline buttons where the amount of depression feels overwrought, and which I’ve had to override on Mavis to make work sensibly.

edwardhorsford:
Great to see this! We've started using small buttons in our service in groups. One thing I also did was make a `button-group--small` modifier so when using button group they are not so spaced.

Screenshot 2025-09-22 at 08 42 04

anandamaryon1:
Brilliant work and a timely addition of smaller buttons.

On the button sizing, shouldn't they be 40px rather than 42px? 

To me, the 'shadow' is part of the button visually, so it looks misaligned when placed next to a input, like the password input. 

image

I prefer this:

image

I also see the slight focus rendering issue, worst in Chrome on small button, I assume it's the overlapping borders, I've had a fiddle but not found a fix…

image

paulrobertlloyd:
@anandamaryon1 On the focus rendering issue, I did experiment with changing the focus state such that buttons get a black border on focus. That would both address the rendering glitches AND make the focus state more distinct. But wasn’t sure if that was desired/out of scope. What do you think?

anandamaryon1:
Not sure that could work with the outline/secondary button? Change would be too subtle between default and focus. 

image
image

paulrobertlloyd:
Indeed, but that is the situation we have with non-outlined buttons currently (focus only changes colour). I did actually have one idea which might be a bit crazy… will add a mock up when I've got a spare minute.

colinrotherham:
>I also see the slight focus rendering issue, worst in Chrome on small button, I assume it's the overlapping borders, I've had a fiddle but not found a fix…
>
>image

Ah sorry @paulrobertlloyd it's https://github.com/nhsuk/nhsuk-frontend/issues/1357 back again 😔

You'll notice when zooming that `bottom: -2px` and `border-width: 2px` start to show different computed values at higher DPIs and zoom levels. Borders will adjust automatically to allow snapping to the nearest pixel when rendered

It looks like this in Safari at my weird resolution when zooming:

Screenshot 2025-09-23 at 14 08 03

Screenshot 2025-09-23 at 14 07 41

<br>
<br>

**Explainer:** See how Safari has switched to a ~2px~ 1.74px border but left/right is still 2px

Screenshot 2025-09-23 at 14 10 16

<br>
<br>

Looks like we've lost the `:after` active state stretching the click target over the shadow too?

This was a feature we kept from GOV.UK Frontend so buttons stay active when mousing into the shadow

Screenshot 2025-09-23 at 14 19 03

frankieroberto:
➕ for the small button variant! That’ll be super useful, and is already something we’re using in Record a vaccination: [`_button.scss`](https://github.com/NHSDigital/record-a-vaccination-prototype/blob/main/app/assets/sass/components/_button.scss#L3-L9)

paulrobertlloyd:
@anandamaryon1 We could add a border and outline to focused button states (option 2 in image below). This would mean:

- all buttons have a different design when focused (not limited to colour changes only)
- buttons on dark backgrounds have a clearer focused appearance
- consistent with focused form field and image states 
- may resolve visual artefacts shown for focus states given above refactor

Focus states

anandamaryon1:
Interesting, I like this @paulrobertlloyd. Keen to see option 2 implemented to have a play with it. 

Don't think I'd go for option 1, since the contrast between the secondary button default and focus would be pretty low, at least currently with the primary buttons it's pretty much an inverse on focus, which is more noticeable.

paulrobertlloyd:
@anandamaryon1 Take a look at the preview deployment now, all buttons have a border and include an extra outline too. I think it could work!

davekingdesign:
This would be something super useful for us on the secure data environment too. I've been looking at including a search on a table with horizontally aligned search button and have come across the same issue with the button being too tall. 

button size small

edwardhorsford:
Anyone have any thoughts on offering a small variant to the button group to go along side this as suggested in [my comment here](https://github.com/nhsuk/nhsuk-frontend/pull/1600#issuecomment-3317389625)?

frankieroberto:
> Anyone have any thoughts on offering a small variant to the button group to go along side this as suggested in [my comment here](https://github.com/nhsuk/nhsuk-frontend/pull/1600#issuecomment-3317389625)?

Is it worth considering avoiding the need for a small variant to the button group by reducing the default margin-right on small buttons in button groups?

colinrotherham:
The refactor is amazing but reintroduced the visual artifacts (and now adds new focus styles)

We could do with splitting out _just_ a small variant

>Is it worth considering avoiding the need for a small variant to the button group by reducing the default margin-right on small buttons in button groups?

Sounds sensible, let's do it. You'd need a button group modifier for links alongside small buttons though?

edwardhorsford:
Happy for it to happen by default - but as Colin says, we'd need to think how links render.

anandamaryon1:
The height of these small buttons, being 40px, makes sense as it matches our inputs but it dips them below WCAG AAA. 

I wonder whether we could/should make them, and text inputs, 44px so that they pass AAA?

colinrotherham:
For context, GOV.UK Frontend are still using ~44px~ 40px but linking the relevant issues:

* https://github.com/alphagov/govuk-frontend/issues/1817
* https://github.com/alphagov/govuk-frontend/issues/2060

But compare [Target Size (Minimum) (Level AA) at 24px](https://www.w3.org/WAI/WCAG22/Understanding/target-size-minimum.html) with [Target Size (Enhanced) (Level AAA) at 44px](https://www.w3.org/WAI/WCAG22/Understanding/target-size-enhanced)

[Go back far enough](https://github.com/alphagov/govuk-frontend/issues/141#issuecomment-316941253) and @edwardhorsford can advise too 😊

anandamaryon1:
Yeah interesting to consider that target size and visual size can be different, like on the checkboxes. 

I see inputs on both design systems as 40px heigh though, not 44px?

colinrotherham:
>I see inputs on both design systems as 40px heigh though, not 44px?

Sorry, typo, edited it

I meant to say radios/checkboxes are 44px but they still use 40px for text inputs etc

They have a nice form alignment example page here though:
https://govuk-frontend-review.herokuapp.com/examples/form-alignment

anandamaryon1:
Presented and discussed with design system review panel. **Outcome: OK to release pending guidance**. 

Concerns are around how the guidance can be clear about using them only where needed, as they are smaller and therefore marginally less accessible. 

I've made an initial draft of guidance for small buttons: https://github.com/nhsuk/nhsuk-service-manual/pull/2350

anandamaryon1:
Still unsure about the focus state change and keen to keep playing with it, so to that end, my current thinking is to try and decouple the new focus state and refactor with a small version of our current buttons. So that we can release small buttons and password input sooner, and continue to explore the focus state and refactor.

My worry with the current focus state update is the lack of contrast between default and focus on secondary buttons: 

image

paulrobertlloyd:
> I wonder whether we could/should make them, and text inputs, 44px so that they pass AAA?

Something for a separate PR? Otherwise this starts to unravel!

> My worry with the current focus state update is the lack of contrast between default and focus on secondary buttons

Right now the focus state for ‘primary’ buttons lacks contrast. With this change, only the focus state for secondary buttons has that issue (while increasing contrast for all buttons on dark backgrounds). That seems like a better state of affairs, no?

anandamaryon1:
Yes you're right, best to raise as a separate issue/PR. 

On the focus style: the current button focus states essentially invert the colours on primary buttons, which is a much more obvious / higher contrast change. Green to yellow is 4.14, and then the white label to black is 14.4.

(current styles)
image

colinrotherham:
Hi @paulrobertlloyd

I've rebased this so it's a very minimal change on top of https://github.com/nhsuk/nhsuk-frontend/pull/1643

paulrobertlloyd:
Had a play this evening, and made the following changes:

- Increased the border radius from 4px to 8px (6px for small buttons)
- Increased the inline padding (from 16px to 20px, to give a more even spacing on all sides)
- Reduced the height of the shadow (still have ‘pressed’ behaviour, but have it look less odd when set aside text inputs)
  - Added side-effect is that the shadow is deeper when focused 
- Reviewed button alignment (again), now that the overall height calculations are different

> [!NOTE]
> I haven’t updated the reference images, so look at the preview deployment to see the changes 

All buttons:

Screenshot of all buttons

Button alongside text input:

Screenshot of button alongside text input

colinrotherham:
I think I've managed to bring this PR back to life bar [one temporary fix](https://github.com/nhsuk/nhsuk-frontend/pull/1600/commits/afc6c16dba809ed4b1d4fb42cbf5c549b257828c)

Still seeing the button border radius artifacts that [showed up in this earlier screenshot too](https://github.com/nhsuk/nhsuk-frontend/pull/1600#issuecomment-3555318462)

Button border radius artifacts

PR #1602, opened 2025-09-23
Remove invalid option from example

edwardhorsford:
I don't believe `href` is a valid option when the back link is a button.


CHANGELOG.md
```diff
@@ -41,6 +41,7 @@ This change was introduced in [pull request #1594: Always set input `:focus` box
 - [#1593: Fix Sass deprecated `core/settings/colours` partial](https://github.com/nhsuk/nhsuk-frontend/pull/1593)
 - [#1594: Always set input `:focus` box-shadow colour](https://github.com/nhsuk/nhsuk-frontend/pull/1594)
 - [#1599: Make border colour on `nhsuk-panel` mixin optional](https://github.com/nhsuk/nhsuk-frontend/pull/1599)
+- [#1602: Remove invalid `href` option from back link component when used as a button](https://github.com/nhsuk/nhsuk-frontend/pull/1602)
```
colinrotherham: Don't worry about a changelog entry

The examples in **macro-options.mjs** are just for internal use

Good to fix it though (the `href` was silently ignored)

PR #1603, opened 2025-09-23
Update header examples

paulrobertlloyd:
## Description

- Use consistent naming scheme (also helps with ordering of examples)
- Add tests for organisation options (with split, with descriptor)
- Build up header examples by sub-component, alongside fuller examples
- Replicate all tests for blue header using white header variant
- Reduce the number of screenshots taken (from 141 to 76)

This change should mean more variations are covered, and the page demonstrating the different examples should be easier to parse.

## Checklist

- [ ] CHANGELOG entry


packages/nhsuk-frontend/src/nhsuk/components/header/macro-options.mjs
```diff
@@ -276,85 +276,108 @@ export const params = {
   }
 }
 
+const nhsukNavigationItems = [
```
colinrotherham: Lovely stuff here @paulrobertlloyd!

Well worth the hard work 🙌

Do you mind keeping these files as dumb as possible and avoid this const?

Repetitive yeah, but easy to edit for all

packages/nhsuk-frontend/src/nhsuk/components/header/macro-options.mjs
```diff
@@ -276,85 +276,108 @@ export const params = {
...
+  }
+]
+
 /**
  * Nunjucks macro option examples
  *
  @satisfies {{ [example: string]: MacroExample }}
  */
 export const examples = {
-  'default': {
```
colinrotherham, line 285: Can we keep a default example at the top? 🙏

I'm eventually going to make an "all components" page showing the one default from all

Think of it as the showcase example from the top of the design system component page

packages/nhsuk-frontend/src/nhsuk/components/header/macro-options.mjs
```diff
@@ -276,85 +276,108 @@ export const params = {
...
-        placeholder: 'Search',
-        visuallyHiddenLabel: 'Search the NHS digital service manual'
-      },
-      navigation: {
+      }
+    },
+    options: {
+      width: false
+    },
+    screenshot: {
```
colinrotherham, line 349: Do you still think we need so many screenshots for the header?

colinrotherham:
That failing regression test is flaky, it comes and goes 😔

packages/nhsuk-frontend/src/nhsuk/components/header/macro-options.mjs
paulrobertlloyd: Done.

packages/nhsuk-frontend/src/nhsuk/components/header/macro-options.mjs
paulrobertlloyd: I agree, there are a lot! Which ones do you think we should reduce the number for/not screenshot?

colinrotherham: They took nearly 6 mins in this PR versus nearly 4 mins in #3690 😮

I'd have them all if they were fast, but we'd best narrow it down

E.g. Some obvious ones:

* Fewer overflow sizes? Don't need to double up on blue and white
* Skip the larger sizes for the various organisation name variations

During development we needed it, but do you reckon we still need all the hover and focus states for linked logo, linked service name, combined logo and search input? Doubled up for both blue and white too

Similarly we have screenshots for both logged in and logged out—do we need both or none? We already have complex account and account with search examples

See what you think

colinrotherham:
Sorry @paulrobertlloyd but could you remove the lovely loop for the white variants?

Best if we keep things as dumb as possible, here too

E.g. I'm assuming `initExample('white with navigation')` won't work with Visual Studio Code autocomplete on your other PR unless TypeScript sees a non-dynamic object key named `white with navigation`

paulrobertlloyd:
> Best if we keep things as dumb as possible, here too

Hmmmm, this to me seems like the best way to ensure we get full coverage of variants. If there’s a better way of doing this or documenting the loop, we can do that, but in this instance I think this is the best way to make sure we get all the examples shown/covered.

I wonder… is there something we can do in the macro example object to enable outputting multiple variants of a component for a given configuration… thinking of other components like buttons that have variants keyed off class modifier(s).

colinrotherham:
Yeah I'm up for exploring better approaches

Could have something like this?

```patch
- initExample('white with navigation')
+ initExample('with navigation', { variant: 'white' })
```

Just want to make sure the tests can get access to the same HTML when needed

One for another PR though? 🙏

We're doing it the dumb copy and paste way for small radios and checkboxes btw (for now)

paulrobertlloyd:
Reverted to ‘dumb’ list of examples, and reduced the number of screenshots generated by:

- only adding screenshots across multiple breakpoints for:
  - default example
  - complex account layout with search and navigation
  - example of navigation with overflow
- only adding screenshots of mobile and desktop breakpoints for:
  - simple account examples
- only adding screenshots for the desktop breakpoint for the remaining examples including all white variant examples

This reduces the number of screenshots generated from 141 to 76, while still covering the most complex layouts, logo hover/focus states, search input focus state and the 2 colour variations.

paulrobertlloyd:
I did – but I think there are leftovers for other components, going by the file diff after running `visual:test:ref` (can address in another PR if needed).

PR #1609, opened 2025-09-29
Remove the margin below from last item in button group

frankieroberto:
This removes the margin from below the last item within the button group in mobile view, as this otherwise doubles up with the margin below the button group container itself.

### DIFF
--- a/packages/nhsuk-frontend/src/nhsuk/core/objects/_button-group.scss
+++ b/packages/nhsuk-frontend/src/nhsuk/core/objects/_button-group.scss
@@ -62,6 +62,13 @@
       margin-bottom: $vertical-gap + $button-shadow-size;
     }
 
+    // Remove the margin from below the last item within the
+    // button group in mobile view, as this otherwise doubles up
+    // with the margin below the button group container itself.
+    > :last-child {
+      margin-bottom: 0;
+    }
+
     // On tablet and above, we also introduce a 'column gap' between the
     // buttons and links in each row and left align links
     @include nhsuk-media-query($from: tablet) {


packages/nhsuk-frontend/src/nhsuk/core/objects/_button-group.scss
```diff
@@ -62,6 +62,13 @@
       margin-bottom: $vertical-gap + $button-shadow-size;
     }
 
+    // Remove the margin from below the last item within the
+    // button group in mobile view, as this otherwise doubles up
+    // with the margin below the button group container itself.
+    > :last-child {
+      margin-bottom: 0;
```
colinrotherham, line 69: Can I see this in action anywhere? I'm not seeing it on the [**button group** example](https://nhsuk-frontend-pr-1609.herokuapp.com/nhsuk-frontend/examples/button-group/) because all paragraphs and non-button links have a higher-specificity `margin-bottom` above

anandamaryon1: Seems to work on the buttons but not links, is this intended? And applies to desktop too, not mobile only.

colinrotherham:
Mind if we close this @frankieroberto?

Unfortunately there's a complex margin dance between links and buttons within button groups 😞

Due to button shadows and transparent borders, settings links to have the same font size and line height wasn't enough. We have to add extra margin to expand their box size. Our [calculations needed bug fixes recently](https://github.com/nhsuk/nhsuk-frontend/pull/1643/commits/e28b4717411e23246d6d830b9671139d184aaf1f) in https://github.com/nhsuk/nhsuk-frontend/pull/1643

Link margins in button groups

<br>
<br>

I'll push an alternative fix into https://github.com/nhsuk/nhsuk-frontend/pull/1196

PR #1612, opened 2025-10-01
Fix styling of action link when used in emergency cards

anandamaryon1:
## Description
Remove underline from action link when used in emergency care card. 

Currently, because of the default link styling adding `text-decoration: underline`, there is no hover state.

Before:
image

After: 
image

## Checklist

- [x] CHANGELOG entry

### DIFF
--- a/packages/nhsuk-frontend/src/nhsuk/components/card/_index.scss
+++ b/packages/nhsuk-frontend/src/nhsuk/components/card/_index.scss
@@ -252,7 +252,7 @@ $card-border-hover-colour: $nhsuk-card-border-hover-colour;
       color: $nhsuk-reverse-text-colour;
       background-color: nhsuk-colour("black");
 
-      a {
+      a:not(.nhsuk-action-link) {
         @include nhsuk-link-style-white;
       }
 
--- a/packages/nhsuk-frontend/src/nhsuk/components/card/_index.scss
+++ b/packages/nhsuk-frontend/src/nhsuk/components/card/_index.scss
@@ -252,7 +252,7 @@ $card-border-hover-colour: $nhsuk-card-border-hover-colour;
       color: $nhsuk-reverse-text-colour;
       background-color: nhsuk-colour("black");
 
-      a:not(.nhsuk-action-link) {
+      a:not(.nhsuk-action-link):not(.nhsuk-button) {
         @include nhsuk-link-style-white;
       }
 
--- a/packages/nhsuk-frontend/src/nhsuk/components/card/macro-options.mjs
+++ b/packages/nhsuk-frontend/src/nhsuk/components/card/macro-options.mjs
@@ -216,13 +216,14 @@ export const examples = {
         <li>you're coughing up more than just a few spots or streaks of blood – this could be a sign of serious bleeding in your lungs</li>
         <li>you have severe difficulty breathing – you're gasping, choking or not able to get words out</li>
       </ul>
-      components.render('action-link', {
+
+      ${components.render('action-link', {
         context: {
           classes: 'nhsuk-action-link--reverse',
           text: 'Find your nearest A&E',
           href: '#'
         }
-      })
+      })}
     `
   },
   'primary (with chevron)': {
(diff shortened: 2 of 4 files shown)


packages/nhsuk-frontend/src/nhsuk/components/card/_index.scss
```diff
@@ -252,7 +252,7 @@ $card-border-hover-colour: $nhsuk-card-border-hover-colour;
       color: $nhsuk-reverse-text-colour;
       background-color: nhsuk-colour("black");
 
-      a {
+      a:not(.nhsuk-action-link) {
```
colinrotherham: Ahh we do something similar for button groups via `a:not(.nhsuk-button)`

Could you please check if buttons-as-links aren't incorrectly given underlines too? 😬

packages/nhsuk-frontend/src/nhsuk/components/card/macro-options.mjs
```diff
@@ -205,6 +205,26 @@ export const examples = {
...
+        <li>you're coughing up more than just a few spots or streaks of blood – this could be a sign of serious bleeding in your lungs</li>
+        <li>you have severe difficulty breathing – you're gasping, choking or not able to get words out</li>
+      </ul>
+      <a class="nhsuk-action-link nhsuk-action-link--reverse" href="#">
+        <svg class="nhsuk-icon nhsuk-icon--arrow-right-circle" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" width="16" height="16" focusable="false" aria-hidden="true">
+          <path d="M12 2a10 10 0 0 0-10 9h11.7l-4-4a1 1 0 0 1 1.5-1.4l5.6 5.7a1 1 0 0 1 0 1.4l-5.6 5.7a1 1 0 0 1-1.5 0 1 1 0 0 1 0-1.4l4-4H2A10 10 0 1 0 12 2z"></path>
+        </svg>
+
+        <span class="nhsuk-action-link__text">Find your nearest A&amp;E</span>
+      </a>
```
colinrotherham: Thanks for adding the example!

Can you please use the `components.render()` helper?

It makes sure the HTML never gets stale and renders using the action link macro

```suggestion
      components.render('action-link', {
        context: {
          classes: 'nhsuk-action-link--reverse',
          text: 'Find your nearest A&E',
          href: '#'
        }
      })
```

anandamaryon1: Ah cool I didn't know we could do this, very nice

anandamaryon1: Whoops, committed your suggestion but it doesn't render correctly. I don't know the correct syntax to get this helper to work.

packages/nhsuk-frontend/src/nhsuk/components/card/_index.scss
anandamaryon1: ooh good thinking, yes they are, pushed a fix

packages/nhsuk-frontend/src/nhsuk/components/card/macro-options.mjs
colinrotherham: My fault, I missed the curly braces. Fixed in https://github.com/nhsuk/nhsuk-frontend/pull/1612/commits/393066be1f680eb475dda6e36765aca9af17ce59

colinrotherham (approved):

shared/lib/components.mjs
```diff
@@ -45,6 +45,9 @@ export async function load(component) {
         // Sort default to top
         ['default', ''],
 
+        // Sort urgent with non-urgent
+        ['non-', ''],
+
```
colinrotherham, line 50: For context, I've updated our component example sorting function to strip the text `non-`

Visually you still see "non-urgent" but [it's now sorted alongside "urgent" on the page](https://nhsuk-frontend-pr-1612.herokuapp.com/nhsuk-frontend/components/card/#non-urgent-blue)

We do similar things so sizes are sorted from S, M, L, XL etc

colinrotherham:
[GitHub are experiencing issues with their macOS runners](https://www.githubstatus.com) which explains the regression test timeouts

packages/nhsuk-frontend/src/nhsuk/components/card/macro-options.mjs
anandamaryon1: Ah I see, thank you

PR #1628, opened 2025-10-15
Fix Node.js `ReferenceError: HTMLElement is not defined` when imported server-side

colinrotherham:
## Description

This PR fixes NHS.UK React components when bundled with `nhsuk-frontend`

Even though NHS.UK React components lazy-load `import('nhsuk-frontend')` in browsers only, tools like [webpack](https://webpack.js.org) include NHS.UK frontend code in the same bundle so this error is thrown:

```console
ReferenceError: HTMLElement is not defined
    at <static_initializer> (file:///path/to/project/node_modules/nhsuk-frontend/src/nhsuk/component.mjs:23:24)
    at file:///path/to/project/node_modules/nhsuk-frontend/src/nhsuk/component.mjs:7:3
```

Browser globals `HTMLElement` and `HTMLAnchorElement` are not defined in Node.js

### Example

Here's a really awful example of how NHS.UK frontend might be imported but only run client-side:

```jsx
import { initAll } from 'nhsuk-frontend'; // 🔥 HTMLElement is not defined
import useSSR from 'use-ssr'

function Example() {
  const { isBrowser } = useSSR()
  
  if (isBrowser) {
    initAll()
  }

  return (<p>Hello {isBrowser ? 'browser' : 'server'}</p>);
}
```

## Checklist

- [x] CHANGELOG entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -199,7 +199,8 @@ This change was introduced in [pull request #1620: Make all Nunjucks components 
 - [#1594: Always set input `:focus` box-shadow colour](https://github.com/nhsuk/nhsuk-frontend/pull/1594)
 - [#1599: Make border colour on `nhsuk-panel` mixin optional](https://github.com/nhsuk/nhsuk-frontend/pull/1599)
 - [#1601: Fix accessible name for linked logo in header component](https://github.com/nhsuk/nhsuk-frontend/pull/1601)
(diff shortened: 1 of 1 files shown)


anandamaryon1 (approved):

colinrotherham:
Here's a Next.js sample app (Pages Router) before and after which uses webpack bundling

I don't see the issue using App Router

## Before

Next js Pages Router before

## After

Next js Pages Router after

PR #1633, opened 2025-10-20
Review global `box-sizing` usage

paulrobertlloyd:
## Description

Globally setting `box-sizing` using best practice from a decade ago is causing problems in components, not least checkboxes and radios inside `nhsuk-details__text` wrappers.

Best practice as of 2025 is to [not set this value](https://www.oddbird.net/2025/09/04/box-model/) globally — certainly if starting a project from scratch, and using intrinsic layouts which don't use fixed widths.

An alternative approach is set this value globally using the `*` selector, and not attempt to force inheritance. However, as this doesn’t encompass pseudo elements, this then means adding this value explicitly in such cases where it is needed.

~~I tried removing the global `box-sizing` reset, but a number of tests failed. Although not beyond the realms of fixing, I have kept the global default for `box-sizing` as `border-box`, removed it from rules where it is inherited and kept it for the pseudo elements used to draw checkboxes and radios.~~

~~The only visual differences are the space between the underlined NHS logo.~~

I’ve chosen the former approach; not setting a global `box-sizing` value and instead setting it where necessary. This is especially important as the global `box-sizing` was never imported on a per-component basis. 

I’ve also added an example of small radios and checkboxes inside details elements… arguably this example page can now be removed, but maybe we want to keep it around for a while.

## Checklist

- [x] CHANGELOG entry

### DIFF
--- a/packages/nhsuk-frontend/src/nhsuk/core/generic/_box-sizing.scss
+++ b/packages/nhsuk-frontend/src/nhsuk/core/generic/_box-sizing.scss
@@ -1,0 +1,7 @@
+@use "../settings" as *;
+
+@include nhsuk-warning(
+  "import-using-generic-box-sizing",
+  "Importing using 'core/generic/box-sizing' is deprecated. Remove your import statement for 'core/generic/box-sizing'.",
+  $silence-further-warnings: false
+);
--- a/packages/nhsuk-frontend/src/nhsuk/core/core.unit.test.mjs
+++ b/packages/nhsuk-frontend/src/nhsuk/core/core.unit.test.mjs
@@ -205,6 +205,22 @@ describe('Core', () => {
         expect.anything()
       )
     })
+
+    it('outputs a warning when importing the core "generic/box-sizing" file', async () => {
+      const sass = outdent`
+        @forward "core/generic/box-sizing";
+      `
+
+      await compileStringAsync(sass, {
+        loadPaths: ['packages/nhsuk-frontend/src/nhsuk'],
+        logger
+      })
+
+      expect(logger.warn).toHaveBeenCalledWith(
+        `Importing using 'core/generic/box-sizing' is deprecated. Remove your import statement for 'core/generic/box-sizing'. To silence this warning, update $nhsuk-suppressed-warnings with key: "import-using-generic-box-sizing"`,
+        expect.anything()
+      )
+    })
   })
 })
 


colinrotherham (changes requested):
I love this 🙌

We've got one thing to resolve for individually imported components:

```scss
@use "../../core/settings" as *;
@use "../../core/tools" as *;
```

☝️ The familiar "minimum things" lacks the box-sizing partial

Any ideas?

paulrobertlloyd:
@colinrotherham Given the fact that:

1. Best practice is not to set a global value for box-sizing (most elements don’t need this value to be changed)
2. Some components previously explicitly set this value and some didn’t
3. `core/generic` is not imported into individual component styles

I have now:
1. Removed the global `box-sizing` value entirely
2. Reviewed where `box-sizing` needs to be explicitly set (and where it does not)

This seems like a more consistent approach, and means that components won’t render incorrectly if imported individually. Furthermore, there are no longer any visual regressions.

packages/nhsuk-frontend/src/nhsuk/core/generic/_box-sizing.scss
colinrotherham, line 1: Happy with this file removal making it v11 only?

paulrobertlloyd:
Rebased `main` to fix `CHANGELOG.md` conflict.

packages/nhsuk-frontend/src/nhsuk/core/generic/_box-sizing.scss
paulrobertlloyd: I’ll let you take it from here. Personally, I think you could get away with this being in v10 because this fixes issues present in v10.

Is the issue might be people `@use`ing `core/generic` and it failing? Could we keep it as an empty file in v10, and remove in v11?

colinrotherham: Yeah that's it

Keep the (mostly) empty file but log a deprecation warning and we can have it in v10?

Like this one:

https://github.com/nhsuk/nhsuk-frontend/blob/main/packages/nhsuk-frontend/src/nhsuk/core/settings/_colours.scss

paulrobertlloyd:
> What's the method for figuring out what needs box-sizing adding or not?

Combination of
- looking to see if element has a set width and borders/padding
- visual regression testing

packages/nhsuk-frontend/src/nhsuk/core/generic/_box-sizing.scss
paulrobertlloyd: Brill, works for me. I’ll let you manage how this filters down to releases, if that’s okay?

packages/nhsuk-frontend/src/nhsuk/components/header/_index.scss
```diff
@@ -492,7 +494,6 @@ $nhsuk-header-reverse-item-active-colour: $nhsuk-reverse-text-colour;
 
   .nhsuk-header__menu-toggle {
     $_chevron-font-size: 19;
-    box-sizing: border-box;
```
colinrotherham, line 495: Do we need to keep this?

Just thinking of the React library where element types can be swapped

packages/nhsuk-frontend/src/nhsuk/components/panel/_index.scss
```diff
@@ -30,7 +30,6 @@ $nhsuk-border-width-panel: $nhsuk-panel-border-width;
 
 @include nhsuk-exports("nhsuk/components/panel") {
   .nhsuk-panel {
-    box-sizing: border-box;
```
colinrotherham, line 33: Are we definitely alright removing this?

We can't guarantee what `.nhsuk-panel` is added to

colinrotherham:
Thanks @paulrobertlloyd I've added https://github.com/nhsuk/nhsuk-frontend/pull/1633/commits/f4b7c2c8558e57e5164f811b6c49943eb934b667

> Combination of
>
> * looking to see if element has a set width and borders/padding
> * visual regression testing

I've checked against GOV.UK Frontend just now too

(Thanks for adding it to radio/checkbox labels so our `max-width` with calc still works 🙌)

packages/nhsuk-frontend/src/nhsuk/core/generic/_box-sizing.scss
colinrotherham: Resolved in https://github.com/nhsuk/nhsuk-frontend/commit/f4b7c2c8558e57e5164f811b6c49943eb934b667

packages/nhsuk-frontend/src/nhsuk/components/header/_index.scss
paulrobertlloyd: There is no width applied to this element (directly) so I don’t think it’s doing anything.

packages/nhsuk-frontend/src/nhsuk/components/panel/_index.scss
paulrobertlloyd: I’d be happy to reinstate this rule here given the somewhat similar card component  needs it too.

colinrotherham: Yeah let's do it just in case

PR #1635, opened 2025-10-21
Resolve Nunjucks template issues flagged by Jinja port

colinrotherham:
## Description

This PR resolves various Nunjucks issues found during https://github.com/NHSDigital/nhsuk-frontend-jinja/pull/64

* Strict equality, e.g. `=== false`
* Usage of `elseif` versus `elif`
* Usage of `[x, y, z].includes()` versus `in [x, y, z]`
* Usage of `+` (addition) versus `~` for strings
* Single not double quotes in places

I've also included bug fixes to `nhsukAttributes()` affecting `safe` or `escape` filtered values. Since the macro came originally from GOV.UK Frontend I've contributed these back with the team via:

* https://github.com/alphagov/govuk-frontend/pull/6351

## Checklist

- [x] CHANGELOG entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # NHS.UK frontend Changelog
 
+## Unreleased
+
+### :wrench: **Fixes**
+
+- [#1635: Resolve Nunjucks template issues flagged by Jinja port](https://github.com/nhsuk/nhsuk-frontend/pull/1635)
+
 ## 10.1.0 - 15 October 2025
 
 ### :new: **New features**


packages/nhsuk-frontend/src/nhsuk/components/button/template.njk
```diff
@@ -21,21 +21,21 @@
...
 
-{% if element == 'a' %}
-<a{{ commonAttributes | safe }} href="{{ params.href if params.href else '#' }}" draggable="false" role="button">
+{% if element == "a" %}
+<a{{ commonAttributes | safe }} href="{{ params.href if params.href else "#" }}" draggable="false" role="button">
   {{ params.html | safe if params.html else params.text }}
 </a>
 
-{% elseif element == 'button' %}
+{% elif element == "button" %}
```
frankieroberto, line 33: `elseif` seems slightly more readable to me but 🤷

colinrotherham: Agree entirely, hadn't realised until now

These were the odd ones out (and GOV.UK Frontend do the same)

colinrotherham:
>Is there a Nunjucks style linter we could set up to catch these?

@frankieroberto Yeah we could use [**djLint** (Python) with the Nunjucks profile](https://djlint.com/docs/languages/nunjucks/)

A lot of these were flagged using the Jinja profile over in:
https://github.com/NHSDigital/nhsuk-frontend-jinja

PR #1637, opened 2025-10-23
Add an example of a non-clickable secondary card with metadata

frankieroberto:
As used on https://design-history.prevention-services.nhs.uk


colinrotherham:
Are you alright adding the visual regression image and we can get this in? 🙏

Thanks @frankieroberto

anandamaryon1:
Good to capture in the examples, thanks

colinrotherham (approved):

Issue #504, opened 2019-07-22
Empty items should not display in macros


GrilloPress:
If you use logic on an item in a radio macro or a summary list macro to not display, the outer container still prints on the screen. Creating redundant `div`s, border and spacing.

When you use logic so that an item is not true, nothing should print.

This issue is the same as found in the govuk-frontend [#1494](https://github.com/alphagov/govuk-frontend/issues/1494)

## Example code

Truncated example:
```
{{ summaryList({
        rows: [{
            key: {
              text: "Contract start date"
            },
            value: {
              text: startDate
            },
            actions: {
              items: [
                {
                  href: "/submit-your-edec/sections/practice-details/return-user/contract-start-date",
                  text: "Change",
                  visuallyHiddenText: "contract start date"
                }
              ]
            }
          } if data['contractType'] == 'apms'
      ]
    }) }}
```

If true this should print, if not then this should not print.

However, when not true, the outer row still gets printed.

HTML that gets printed when false:

```
<div class="nhsuk-summary-list__row">
      <dt class="nhsuk-summary-list__key">
        
      </dt>
      <dd class="nhsuk-summary-list__value">
        
      </dd>
      
    </div>
```

Example in full summary list printing out two extra lines and additional padding:

![image](https://user-images.githubusercontent.com/5756226/61619441-c26bd800-ac66-11e9-86c2-db29f57519eb.png)

## Suggested fix

Like in the govuk-frontend [#1494](https://github.com/alphagov/govuk-frontend/issues/1494) the fix should be something like:

1. testing if a row or item is empty and not printing anything if empty
2. add a manual flag inside the row or item to toggle display. This would make it easier to prototype and select if a row should output or not [#198](https://github.com/alphagov/govuk-design-system-backlog/issues/198)

tomdoughty:
I have had a quick look at this... to implement either solution we need to add a filter to the for in loop inside all necessary files. Therefore I think we should use a custom filter which would cover both suggested fixes. Example below shows how you can hide by setting the hideEmpty attribute or by setting the item to falsey.

In nunjucks config:
```
env.addFilter('hideEmpty', arr => arr.filter((val) => {
  if (val.hideEmpty || !val) return false;
  return true;
}));
```

In component:
```
{% for row in params.rows | hideEmpty %}`
...
{% endfor %}
```

Example useage:
```
{{ summaryList({
  rows: [
    {
      key: {
        text: "Name"
      },
      value: {
[...]

PR #1653, opened 2025-10-29
Only show a task list item if not empty

paulrobertlloyd:
## Description

The task list component currently doesn’t check for the presence of an item before displaying list item HTML, meaning empty space is shown:

Screenshot 2025-10-29 at 18 03 55 

## Checklist

- [x] CHANGELOG entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ This change was introduced in pull requests [#1633: Review global `box-sizing` u
 
 - [#1633: Review global `box-sizing` usage](https://github.com/nhsuk/nhsuk-frontend/pull/1633)
 - [#1635: Resolve Nunjucks template issues flagged by Jinja port](https://github.com/nhsuk/nhsuk-frontend/pull/1635)
+- [#1653: Only show a task list item if not empty](https://github.com/nhsuk/nhsuk-frontend/pull/1653)
 
 ## 10.1.0 - 15 October 2025
 


colinrotherham:
Nice one! We don't have many loops across our components, so would be good to assess https://github.com/nhsuk/nhsuk-frontend/issues/504

Worth seeing if we could just uplift from [govuk/components/task-list/template.njk](https://github.com/alphagov/govuk-frontend/blob/main/packages/govuk-frontend/src/govuk/components/task-list/template.njk)?

They use a different approach to keep the HTML indentation nice and clean

paulrobertlloyd:
Uplifted with upstream GOV.UK component, but note that:

- we’re using `~` (for Jinga compatibility?)
- we have `params.id`

The rest is pretty consistent with what’s already there. Let’s address loops in other components in another PR.

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ This change was introduced in pull requests [#1633: Review global `box-sizing` u
 
 ### :wrench: **Fixes**
 
+- [#1653: Only show a task list item if true](https://github.com/nhsuk/nhsuk-frontend/pull/1653)
```
colinrotherham: Minor nit pick but "if true" could do with clarifying

Maybe "if not empty" or similar

colinrotherham (approved):

PR #1654, opened 2025-10-30
[v10] Add option to make tables sortable

colinrotherham:
## Description

This expands the existing [table component](https://service-manual.nhs.uk/design-system/components/table) to enable a new option to make some or all of the columns sortable by clicking the column header. It is partly inspired by [sortable table component](https://design-patterns.service.justice.gov.uk/components/sortable-table/) from MOJ Frontend.

The table sorting feature is done using JavaScript as a progressive enhancement. If JavaScript is unavailable or not supported, then the table remains in its initial state, and no sort buttons are added.

If you're using the `tables` Nunjucks macro, to configure header cells:

- Add the `sort: "ascending"` or `sort: "descending"` option to the currently sorted column only
- Add the `sort: true` option to enable sorting on any other columns

For example:

```patch
  {{ table({
    caption: "Childhood MMR coverage",
    head: [
      {
-       text: "Nation"
+       text: "Nation",
+       sort: "ascending"
      },
      {
        text: "MMR",
-       format: "numeric"
+       format: "numeric",
+       sort: true
      }
    ],
    rows: [
      [
        {
          text: "Northern Ireland"
        },
        {
          text: "86.4%",
          format: "numeric"
        }
      ],
      [
        {
          text: "Scotland"
        },
        {
          text: "89.2%",
          format: "numeric"
        }
      ],
      [
        {
          text: "Wales"
        },
        {
          text: "89.5%",
          format: "numeric"
        }
      ]
    ]
  }) }}
```

Table sorting handles numeric values when `format: "numeric"` is set or when cells contain numbers, or will otherwise sort them alphabetically (or reverse-alphabetical).

You can specify an alternative sort value using the Nunjucks `sortValue` option:

```patch
  {{ table({
    caption: "Appointments",
    head: [
      {
        text: "Title",
        sort: true
      },
      {
        text: "Date",
        sort: "ascending"
      }
    ],
    rows: [
      [
        {
          text: "Vaccination"
        },
        {
-         text: "4 January 2026"
+         text: "4 January 2026",
+         sortValue: "2026-01-04"
        }
      ],
      [
        {
          text: "Health check"
        },
        {
-         text: "15 February 2026"
+         text: "15 February 2026",
+         sortValue: "2026-02-15"
        }
      ]
    ]
  }) }}
```

## Screenshots

Screenshot 2026-01-12 at 16 19 22

Screenshot 2026-01-12 at 16 19 37

## Checklist

- [x] CHANGELOG entry


paulrobertlloyd:
Question of icons…

- Is it useful having the double pointed arrow for unsorted columns (as opposed to no icon)
- Should the arrows be more chevron shaped, to be consistent with iconography used elsewhere?

On the later point, when I was using sortable tables in my service, I used the following SVG:

```svg
<svg version="1.1" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 27 27">
  <path fill="#212b32" d="m13 18.6-7-7.2c-.6-.5-.6-1.4 0-2 .5-.5 1.3-.5 2 0l6 6.2 6-6.2c.7-.5 1.5-.5 2 0 .6.6.6 1.5 0 2l-7 7.2c-.3.3-.6.4-1 .4s-.7-.1-1-.4Z"/>
</svg>
```

anandamaryon1:
Pushed some initial styling for the sortable tables heading buttons and icons, based on designs worked on in the sortable tables working group. 

For discussion.

Some notes:

* I've kept the icon to the left of the heading for right-aligned header cells, to avoid icons being placed right beside each other.
* I've reduced the current sorted underline from 4px to 2px, to make it play nicer with the focus style, I'm open to re-exploring this. (it looked odd since the sorted underline indicator runs from bottom up, and the focus style runs from top down.)
* Playing with it I notice that the first click makes the column ascending - showing smallest first, which makes sense with A-Z. But for some reason, I expect biggest first, with numbers when I click for an initial sort.

To do:
* Give non-button headings matching padding as buttons. (currently only buttons have 2px padding on the outer edge)
* Consider icon and spacing - ascending and descending arrows have inbuilt padding (in the SVG) which affects spacing between the icon and heading text. But, SVGs ideally should all be the same size to avoid content jumping around when swapping them out. Open to ideas.

Screenshot:

image

image

image

<hr>
Preview it:

https://nhsuk-frontend-pr-1654.herokuapp.com/nhsuk-frontend/components/tables/with-numeric-data-sortable/

colinrotherham:
Can we take any inspiration from tabs where only the middle bit is focused?

Tab with focus

packages/nhsuk-frontend/src/nhsuk/components/tables/table.mjs
```diff
@@ -0,0 +1,354 @@
...
+
+  /**
+   * Sortable table config
+   *
+   * @see {@link SortableTableConfig}
+   * @constant
+   * @type {SortableTableConfig}
+   */
+  static defaults = Object.freeze({
+    statusMessage: 'Sort by %heading% (%direction%)',
```
frankieroberto: This could be [my fault](https://github.com/frankieroberto/sortable-table/blob/master/src/sortable-table.js#L7) but I wonder now if this ought to be in present tense – "Sorted by Name (ascending)"?

Could also consider "Ordered by Name (A to Z)" and "Ordered by Height (biggest to smallest)" as more understandable, perhaps?

We should also remember to add in the [I18n](https://github.com/nhsuk/nhsuk-frontend/blob/main/packages/nhsuk-frontend/src/nhsuk/i18n.mjs) function so that it can be customised or translated.

anandamaryon1: Yeah I like the idea of avoiding ascending and descending, I often forget which is which 😅 

Could that be done by linking it to the data type, eg. strings are "A-Z" and numbers are "biggest to smallest"? Or "biggest first". And potentially more of them for any other types, eg. dates.

anandamaryon1:
> Can we take any inspiration from tabs where only the middle bit is focused?
> 
> Tab with focus

That could make things a lot less hacky with styling the buttons. I'm fine with the focus being on the text only, but we do want the increased target area, and the hover should match, to show where you can click to activate it. Would that rely on JS click events for the table header?

colinrotherham:
Don't let the small target area from tabs put you off, both could be bigger

Expanders are similar where target area focus styles are removed, with (fake) focus styles on the text only:

Expander focus

<br>
<br>

>…Would that rely on JS click events for the table header?

I'm assuming yes, but we'd need to research why click events on child elements weren't used

The current single-listener event bubbling approach could be for either compatibility or performance

packages/nhsuk-frontend/src/nhsuk/components/tables/table.mjs
frankieroberto: Yeah, it could potentially be different for different types. Although we only support text and numbers at the moment. 

At the moment we don't anything cover for dates (as date parsing is notoriously tricky!) so for dates users would have to do something like `data-sort-value="2025-02-20"` (which would sort alphabetically but still be in order) or `data-sort-value="20250220"` (which would sort numerically). Either would work.

frankieroberto:
> Expanders are similar where target area focus styles are removed, with (fake) focus styles on the text only

Task list does the same. On the other hand the combined logo + service name link in the header has the focus style cover the whole clickable area, including the logo.

paulrobertlloyd:
Probably need an audit of focus styles… part of me wants to suggest making focus styles big and obvious, now that Firefox doesn’t show them on click (or can at least be avoided)

anandamaryon1:
Seems all the other components expand the click area by placing an absolutely positioned `:after` element to cover the parent container. This is tricky with the table header buttons because of `display: table-cell` (hence my earlier hacky CSS to grow the buttons). 

I've now removed that so that the focus is placed only on the natural sized buttons, and added a hover state to the header cells to illustrate how I'd like them to be, but of course this doesn't trigger the buttons. 

Can someone help me add a click event to the table heads? Or maybe there's a better way?

I also notice a possible slight bug where the `aria-sort="none"` is added to none sortable columns when one is sorted, but maybe this is on purpose to identify the state of the columns?

colinrotherham:
>Can someone help me add a click event to the table heads? Or maybe there's a better way?

There's an event listener on the `<head>` element you can use

```mjs
this.$head.addEventListener('click', this.onSortButtonClick.bind(this))
```

But later on you'll see it ignores all clicks unless it bubbled up from a `button` element

```mjs
onSortButtonClick(event) {
  const $target = /** @type {HTMLElement} */ (event.target)
  const $button = $target.closest('button')

  if (!$button?.parentElement) {
    return
  }
  
  // …
 }
```

In theory you could change this code to allow clicks from non-clickable things

But we shouldn't do really

Is it a bad thing that only the `<button>` is clickable, like MOJ do?

colinrotherham:
Rebased with `main` to fix the conflict

Table examples will now be in ~`table/macro-options.mjs`~ `table/fixtures.mjs`

frankieroberto:
> > Can someone help me add a click event to the table heads? Or maybe there's a better way?
>
> Is it a bad thing that only the `<button>` is clickable, like MOJ do?

The MOJ one doesn't have any hover state. We’ve added a hover state (background colour change to grey) for the whole `<thead>`. The clickable area should be match the hover state area, I think – like on the Task list rows. And in general, the bigger the hit area the better?

Would one option be to make button size expand to fill the entirety of the `<thead>`, and then move the background hover colour to the `<button>`?

I also just spotted that the task list rows are using `color.adjust(nhsuk-colour("grey-5"), $lightness: -6%);` whereas the sortable table headers are using `nhsuk-colour("grey-4")` – should they be consistent? The task list one has more of a noticeable blue tint to me, and isn’t quite as dark.

packages/nhsuk-frontend/src/nhsuk/components/tables/table.mjs
frankieroberto, line 1: I wonder if this should be moved to `table/table.mjs` rather than being treated as a separate component? Especially as I think we’d describe in the design system website as just an optional feature of the Table component, and there’s no separate macro for it?

anandamaryon1:
> > > Can someone help me add a click event to the table heads? Or maybe there's a better way?
> > 
> > 
> > Is it a bad thing that only the `<button>` is clickable, like MOJ do?
> Would one option be to make button size expand to fill the entirety of the `<thead>`, and then move the background hover colour to the `<button>`?

This is what I had first, which kinda worked but relied on some hacky CSS to get the button to grow to fill the `<thead>`. And meant the focus state applied to the whole area ([screens here](https://github.com/nhsuk/nhsuk-frontend/pull/1654#issuecomment-3576651260)). 

I'm not sure on the best approach from a frontend dev perspective but what I want to achieve is the whole `<thead>` clickable, with a hover state that matches the clickable area. And a focus state that applies only to the text if possible, to better match other components, eg. tabs. 

> I also just spotted that the task list rows are using `color.adjust(nhsuk-colour("grey-5"), $lightness: -6%);` whereas the sortable table headers are using `nhsuk-colour("grey-4")` – should they be consistent? The task list one has more of a noticeable blue tint to me, and isn’t quite as dark.

Ohh, well spotted, will need to take a look at this. I think I was using pagination as my reference.

frankieroberto:
@anandamaryon1 the task list trick is to use this to generate a bigger click area, but then the focus background colour only applies to the smaller `<a>` element:

```css
.nhsuk-task-list__link::after {
  content: "";
  display: block;
  position: absolute;
  top: 0;
  right: 0;
  bottom: 0;
  left: 0;
}
```

Bit of a hack but it seems to work?! There might be a better way though.

colinrotherham:
Sorry for the rebase, but I've updated this branch to pick up the new status check names

packages/nhsuk-frontend/src/nhsuk/components/tables/table.mjs
frankieroberto: This is done now in 6eb8f5f73beafcfd390b6dd9079ce8c14e78e5ae

packages/nhsuk-frontend/src/nhsuk/components/tables/table.mjs
frankieroberto: Hmm, given that the code doesn't differentiate between sorting text and sorting dates (both are treated as text), I'm not sure "Ordered by %{heading} (A-Z)" would work - would cause more confusion for dates!

I'll leave as-is and close this for now, but it might need more testing with screen reader users to determine whether "ascending" and "descending" are confusing or not. If they are, then we might need the user to specify what type of thing is being sorted in order to able to do something like "A-Z" or "newest first" or "largest first" or "highest first" depending on what it is... 🤔

frankieroberto:
@colinrotherham @anandamaryon1 I've updated this PR (and its description) to:

* merge the CSS and javascript into the Table component, rather than having it be separate (although we could still choose to describe it as a separate component in the Service Manual website if we wanted to)
* fix the bug with the non-sortable column having a hover state
* adding a 2px padding to the non-sortable columns (only if JavaScript is available)
* update the macro to always add the `data-module="nhsuk-table"` attribute, and have the JavaScript exit early if there are no sortable columns
* add some JSDOM tests

I think this is now ready for a wider review.

colinrotherham:
@frankieroberto @anandamaryon1 I've updated this PR to target v10.x

Can you please check the "link" colours and update if needed?

Good to get this out sooner rather than later

Sortable table link styles

anandamaryon1:
Yep the link colours are OK for now (pre-v11). I don't think we need to change the "active" border-bottom colour, it can stay blue. 

We're doing pilot testing and UR on this over the next few weeks, so can't say yet whether it'll be v10 or v11 when it's ready to release.

colinrotherham:
Thanks for updating this @frankieroberto

Could I add some feedback in light of https://github.com/nhsuk/nhsuk-frontend/pull/1712 (icon buttons)?

We now have all `"arrow-up"`, `"arrow-down"`, `"arrow-left"`, `"arrow-right"` orientations included

Are you alright replacing `"sort-up"` and `"sort-down"` with them?

For comparison, the current ones are offset slightly left and maybe the up/down icon is too large or thick?

Example showing sort icons

<br>
<br>

**Update:** Mostly unrelated to this PR, but the size/position difference is a bit a bit more obvious with text

Sort icon button with text

frankieroberto:
> Thanks for updating this @frankieroberto
> 
> Could I add some feedback in light of #1712 (icon buttons)?
> 
> We now have all `"arrow-up"`, `"arrow-down"`, `"arrow-left"`, `"arrow-right"` orientations included
> 
> Are you alright replacing `"sort-up"` and `"sort-down"` with them?

Nice! This is done now in 501897b. For the time being the up-down icon is the up and down arrows but with each of them nudged apart using an inline transformation. We can tweak this and flatten it if needed.

The minor differences in icons was because one of them was 22x22 and the other was 24x24. These are consistent now.

In 37b6511f46c7ab93d4980e1e040d60ab9e81f917 I’ve switched to using the `nhsIcon` macro in a `<template>` element, and using that within the javascript rather than hardcoding the SVG into the javascript itself.

frankieroberto:
@colinrotherham @edwardhorsford I’ve updated this PR with some changes:

1. rename the `data-initial-sort` attribute to `data-sort-first-direction` - hopefully this makes the intent a bit clearer and less confusing? It sets the direction that the column will be first sorted on when clicked, and then after that it toggles. Default when unspecified is ascending first.

2. added 3 Nunjucks params to act as easier-to-remember helpers, as an alternative to setting an attributes objects: 

| Nunjucks param | HTML attribute |
| ---------- | -------- |
| `sortable: true`<br><br>Makes a column sortable | `aria-sort="none"` |
| `sorted: "ascending"`<br><br>Used with the above - sets the column as sorted on page load | `aria-sort="ascending"` |
| `sortFirstDirection: "descending"`<br><br>Set the direction to first sort on when clicked, and then toggles | `data-sort-first-direction="descending"` |

Does this seem clear?

colinrotherham:
@frankieroberto Yeah seems clear to me

Some questions:

* Do we definitely need a configurable first sort direction?
* Should the first sort direction apply at table- not column-level?

There's something nice about the MOJ sortable table being predictable, picking `"ascending"` by default:

```js
const newSortDirection =
  sortDirection === 'none' || sortDirection === 'descending'
    ? 'ascending'
    : 'descending'
```

It might be strange if I go from column-to-column and each one sorts itself differently on click?

So wondered if we definitely needed it or whether it should be a whole-table default for consistency

packages/nhsuk-frontend/src/nhsuk/components/tables/table.mjs
```diff
@@ -0,0 +1,401 @@
...
+    } else if (sortDirection === 'ascending') {
+      newSortDirection = 'descending'
+    } else if (
+      sortFirstDirection === 'ascending' ||
+      sortFirstDirection === 'descending'
+    ) {
+      newSortDirection = sortFirstDirection
+    } else {
+      newSortDirection = 'ascending'
+    }
```
colinrotherham: Is it weird that we throw away the user's `sortFirstDirection` here?

i.e. We only use it when not already sorted

colinrotherham: Similarly it looks like `sortFirstDirection: "none"` isn't handled

frankieroberto:
@colinrotherham:

> Do we definitely need a configurable first sort direction?
> Should the first sort direction apply at table- not column-level?

I think it’s a nice thing to have, and should apply at a column level, as you might have a table with, say, organisation or area names in column A, which are more usefully sorted in alphabetical (A-Z) order, but numerical data in column B, which depending on context most users may be more likely to be interested in sorting by biggest-to-smallest.

It was a suggestion from @edwardhorsford so he may have more context.

This was the example I used in the fixtures file:

Screenshot 2026-04-29 at 10 36 34

colinrotherham:
As in, we think the new option is necessary above and beyond just clicking twice `ascending` → `descending`?

packages/nhsuk-frontend/src/nhsuk/components/tables/table.mjs
frankieroberto: I think the logic makes sense (although perhaps it's hard to follow?). If the column is already sorted in ascending or descending order, then clicking the heading should always toggle it to the opposite order.

The `sortFirstDirection` is only used to decide which direction to sort in first when going from unsorted to sorted.

`sortFirstDirection: "none"` is handled in the sense that the default first direction will apply (`ascending`).

In reality you only need to set `sortFirstDirection='descending'` as `sortFirstDirection='ascending'` is the default. But you can still set it to show that it’s intentional if you like.

packages/nhsuk-frontend/src/nhsuk/components/tables/macro-options.mjs
```diff
@@ -97,6 +103,24 @@ const options = {
...
+      sortable: {
+        type: 'boolean',
+        required: false,
+        description: 'Set to true to allow this column to be sorted',
+        released: '10.x.0'
+      },
+      sorted: {
+        type: 'string',
+        required: false,
+        description: 'Set to "ascending" or "descending" if the column is sorted on initial page load',
```
colinrotherham: Isn't this written to `aria-sort` so represents the current sort order?

(Rather than an automatic sort that runs on page load)

```suggestion
        description: 'The current sort direction of the column – `"ascending"`, `"descending"` or `"none"`.' Defaults to `"none"`,
```

packages/nhsuk-frontend/src/nhsuk/components/tables/macro-options.mjs
```diff
@@ -97,6 +103,24 @@ const options = {
...
+        released: '10.x.0'
+      },
+      sortFirstDirection: {
+        type: 'string',
+        required: false,
+        description: 'Set to "ascending" or "descending" to set the direction that the column should be first sorted on when clicked. It can then be clicked again to toggle. Default is "ascending".',
```
colinrotherham:
```suggestion
        description: 'The first sort direction applied to the column when clicked – `"ascending"` or `"descending"`.' Defaults to `"ascending"`,
```

packages/nhsuk-frontend/src/nhsuk/components/tables/table.mjs
colinrotherham: Is it a problem if we always honour `sortFirstDirection` when set?

Or are we trying to prevent the first click from doing nothing, because current and first direction match?

(Appreciate that `sortFirstDirection: "none"` to _remove_ the already-sorted direction is impossible)

edwardhorsford:
>As in, we think the new option is necessary above and beyond just clicking twice ascending → descending?

Yes, for some data sets ascending > descending is the less natural order. We can default the direction, but should let authors tell us the correct order if necessary.

A good example might be a table that listed departments and arms length bodies and their budgets and head counts. If you sorted ascending first by default and sorted by budget or head count, you'd just show all the tiny departments. Showing the largest departments first is likely the better default / thing users are intersted in.

@frankieroberto I'm not sure of the name `sorted`, and wonder if it should be `isSorted: true` or `initialSortColumn: true`, and only allowed to exist on one item. It can then either use the initial sort direction or else sort using the default. I think this would also probably be more readable/ scannable when there are already ascending / descending attributes on the macro.

packages/nhsuk-frontend/src/nhsuk/components/tables/macro-options.mjs
frankieroberto: It's kinda both. It should represent the current HTML sort order on page load (for if javascript is not available), but at the moment the JavaScript does also do an initial sort by this column when it initialises.

Probably the guidance would encourage people to not rely on that though, and instead have a good default order in the HTML?

(Could even consider removing the behaviour of the javascript doing an initial sort, to force the issue? 🤔)

colinrotherham: Yeah let's remove it unless we want to test it (and check for unnecessary sorting announcements)

packages/nhsuk-frontend/src/nhsuk/components/tables/macro-options.mjs
frankieroberto: Updated, ta.

packages/nhsuk-frontend/src/nhsuk/components/tables/macro-options.mjs
frankieroberto: Hmm on the other hand, maybe it’s good for the javascript to do a sort on load, even if nothing should change - just in case the HTML is rendered with the rows in the wrong order? Otherwise the user would see the table as being sorted by that column, but it wouldn't actually be...

frankieroberto:
> @frankieroberto I'm not sure of the name `sorted`, and wonder if it should be `isSorted: true` or `initialSortColumn: true`, and only allowed to exist on one item. It can then either use the initial sort direction or else sort using the default. I think this would also probably be more readable/ scannable when there are already ascending / descending attributes on the macro.

It should only exist on one column for sure - not sure what we should do if zero or 2+ columns have it though? Maybe the whole javascript should then not load?

`initialSortColumn: true` is maybe clearer than `sorted: "ascending"` - but does mean that you then need 2 params to set which column is initially sorted and its direction.

Could use a combination of `initialSortColumn: true` and `sortFirstDirection: "ascending"` - is that what you were suggesting?

The only thing we'd lose is the ability to set a column as being initially sorted in one order (say descending), but to have the `sortFirstDirection` as the opposite - so that if you sort a different column and then sort the first column again, it'd be in a different order.

I can’t think of a scenario where you'd want to do that though?

anandamaryon1:
This is coming along great. I like the new `sortFirstDirection`. When used well it helps the sort order to be more intuitive. And having it as optional with a fallback means it'll hopefully only be used when the designer has good reason. 

I'm not sure what your issue with `sorted: "ascending"` @edwardhorsford , just that it's maybe ambiguous or harder to scan when used with `sortable: true`? Could it be `initialSort` or `sortOnLoad` instead?

edwardhorsford:
@anandamaryon1 yeah, it feels harder to scan, but also possibly confusing when there's another parameter that might be present and also saying "ascending" or "descending". It's not obvious this is a boolean thing that only one column can have.

packages/nhsuk-frontend/src/nhsuk/components/tables/table.mjs
```diff
@@ -0,0 +1,401 @@
...
+
+  /**
+   * Sortable table config
+   *
+   * @see {@link TableConfig}
+   * @constant
+   * @type {TableConfig}
+   */
+  static defaults = Object.freeze({
+    statusMessage: 'Sort by %heading% (%direction%)',
```
frankieroberto: @anandamaryon1 @colinrotherham @edwardhorsford @sarawilcox one thing we discussed in the call last week was updating this default status text (which is visually hidden) to be past tense, to reflect that it’s how the table is _currently_ sorted.

However this would be deviation from the MOJ sortable table component, so I think we said we’d check if they’d written it this way for a reason or not?

```suggestion
    statusMessage: 'Sorted by %heading% (%direction%)',
```

edwardhorsford: Reads nicer to me

anandamaryon1: Agree past tense makes more sense. I've asked them.

anandamaryon1:
Having just had a play updating the examples on the guidance PR, I agree with @edwardhorsford approach for `initialSortColumn: true` to make it more explicit and reduce the number of  times `ascending` or `descending` is listed out in the options – it becomes a bit hard to parse.

anandamaryon1:
I wonder whether we need to consider adding scrollable containers alongside this? We agreed a responsive design is out of scope for now and of limited value. However, it doesn't feel right to release something which breaks on small screens. I know tables are an exception since they can break layouts by default but, it feels we should do something better…

frankieroberto:
@anandamaryon1 could we look at that as a separate issue, rather than tying it to sortable tables?

Adding the sortable headings doesn’t affect the width much (although it does a bit, to make space for the icons), and some of the examples still fit on mobile ok:

image

But I’d agree that we could look at tables on mobile generally. This could include both horizontal scrolling and reducing whitespace or font size - something the `govuk-table--small-text-until-tablet` class in the GOV Design System already does! https://design-system.service.gov.uk/components/table/

anandamaryon1:
Yeah not saying it should be a sortable tables specific thing, just maybe we should make sure they're both available at the same time. 

Good point that there are other features we could add to squeeze more in on smaller screens.

anandamaryon1:
Pushed a little update to give table examples row headers, based on accessibility advice. Implementing over on service manual table examples but thought best to add here too.

anandamaryon1:
Noticed a small bug: the hover style initiates when you hover over the header's border-bottom, but it's not clickable. Added a fix in [116eeb5](https://github.com/nhsuk/nhsuk-frontend/pull/1654/commits/116eeb52879603fcc26806a80b32fa076fd28d8e)

colinrotherham:
I'm going to give this branch a rebase, resolve the conflict, and keep a `v11.0.0` version number for now

It makes sure the preview branch becomes `v11.0.0-preview-*` which is compatible with prototypes

packages/nhsuk-frontend/src/nhsuk/macros/icon.njk
```diff
@@ -82,6 +82,13 @@
   <path d="M6.3 9.3l5-5c.2-.2.4-.3.7-.3.3 0 .5.1.7.3l5 5a1 1 0 0 1-1.4 1.4l-3.3-3.3L13 19a1 1 0 0 1-2 0L11 7.4l-3.3 3.3a1 1 0 1 1-1.4-1.4Z"/>
 {% elif name == "arrow-down" %}
   <path d="M17.7 14.7l-5 5c-.2.2-.4.3-.7.3-.3 0-.5-.1-.7-.3l-5-5a1 1 0 0 1 1.4-1.4l3.3 3.3L11 5a1 1 0 0 1 2 0L13 16.6l3.3-3.3a1 1 0 1 1 1.4 1.4Z"/>
+{% elif name == "arrow-up-down" %}
+  <g transform="translate(-4.8, 0)">
+    <path d="M6.3 9.3l5-5c.2-.2.4-.3.7-.3.3 0 .5.1.7.3l5 5a1 1 0 0 1-1.4 1.4l-3.3-3.3L13 19a1 1 0 0 1-2 0L11 7.4l-3.3 3.3a1 1 0 1 1-1.4-1.4Z"/>
+  </g>
+  <g transform="translate(4.8, 0)">
+    <path d="M17.7 14.7l-5 5c-.2.2-.4.3-.7.3-.3 0-.5-.1-.7-.3l-5-5a1 1 0 0 1 1.4-1.4l3.3 3.3L11 5a1 1 0 0 1 2 0L13 16.6l3.3-3.3a1 1 0 1 1 1.4 1.4Z"/>
+  </g>
```
colinrotherham: Are we're happy with these icons now?

We'll need to remove `transform` and swap in an already-transformed path for release

packages/nhsuk-frontend/src/nhsuk/components/tables/template.njk
```diff
@@ -130,3 +153,15 @@
...
+  <template class="nhsuk-js-arrow-up-icon">
+    {{ nhsukIcon("arrow-up") }}
+  </template>
+  <template class="nhsuk-js-arrow-down-icon">
+    {{ nhsukIcon("arrow-down") }}
+  </template>
+  <template class="nhsuk-js-arrow-up-down-icon">
+    {{ nhsukIcon("arrow-up-down") }}
+  </template>
+{% endif %}
```
colinrotherham: Shall we just use CSS background images now that SVG icons are generated?

* ~Hard coded SVGs in the JavaScript~
* ~Hard coded SVGs in the HTML~

Our two compromises are no longer necessary

frankieroberto: @colinrotherham I thought there was a limitation with CSS SVG background images not being able to use `fill: currentcolor` for forced colour mode? But if that's not an issue then yeah, CSS would be easier!

packages/nhsuk-frontend/src/nhsuk/macros/icon.njk
frankieroberto: I think so? @anandamaryon1?

Is there any easy way to do the flattening which doesn't involve manually subtracting 4.8 from every other number? 😄

packages/nhsuk-frontend/src/nhsuk/components/tables/template.njk
colinrotherham: Ahh yes, issue no more!

[For search input](https://github.com/nhsuk/nhsuk-frontend/pull/1660/changes/01b35b88a0e410c889114e0bd8d95f5203da3432) we [picked up a GOV.UK Frontend approach](https://github.com/alphagov/govuk-frontend/blob/366fb10c528a24ebb568a8344b14dd5970e66c72/packages/govuk-frontend/src/govuk/components/footer/_mixin.scss#L100-L118) to use:

* `background` and `background-size` for older browsers
* `mask` and `mask-size` for newer browsers

There's a screenshot in https://github.com/nhsuk/nhsuk-frontend/pull/1660#issuecomment-4360351447

frankieroberto:
@anandamaryon1 @colinrotherham @edwardhorsford ok I’ve updated the param name in 7279943d292e9d448d0c7ff90b63478a8f3d4bb3 from `sorted` to `initialSortColumn ` based on our discussion in the comments above.

Now instead of setting `sorted: 'ascending'` to indicate the initial sort column on page load, you'd set `initialSortColumn: true` and `sortFirstDirection: 'ascending'`.

Hopefully this will be clearer! (Avoids having 2 params which both accept ascending/descending as options).

packages/nhsuk-frontend/src/nhsuk/components/tables/table.mjs
frankieroberto: I've updated it in 9ebd6f5385394c2f5a8b9a1679b863e098e687eb to use 'sorted'.

packages/nhsuk-frontend/src/nhsuk/macros/icon.njk
colinrotherham: I usually ask Copilot 😬

Would be good to determine if 4.8 gives us the centre (can also ask for this to be checked too)

packages/nhsuk-frontend/src/nhsuk/components/tables/table.mjs
```diff
@@ -0,0 +1,401 @@
...
+
+    const columnNumber = Number.parseInt($sortButton?.dataset.index ?? '0', 10)
+
+    if (
+      !$heading ||
+      !$sortButton ||
+      !(sortDirection === 'ascending' || sortDirection === 'descending')
+    ) {
+      return
+    }
```
colinrotherham: Have you had a look at the Sonar check?

There isn't any code coverage for this bit

packages/nhsuk-frontend/src/nhsuk/components/tables/table.mjs
```diff
@@ -0,0 +1,401 @@
...
+
+    if (!$button?.parentElement) {
+      return
+    }
+
+    const $heading = $button.parentElement
+    const sortDirection = $heading.getAttribute('aria-sort')
+    const sortFirstDirection = $heading.dataset.sortFirstDirection
+
+    const columnNumber = Number.parseInt($button.dataset.index ?? '0', 10)
```
colinrotherham: Similarly no coverage for this bit

Is it correct for us to assume `'0'` when no index is in the HTML?

E.g. Let's say if one of the NHS.UK frontend ports omit it (or allow it to be overridden)

packages/nhsuk-frontend/src/nhsuk/components/tables/table.mjs
```diff
@@ -0,0 +1,401 @@
...
+    const { config, $status } = this
+
+    if (
+      !(direction === 'ascending' || direction === 'descending') ||
+      !$button.parentElement ||
+      !config.statusMessage ||
+      !$status
+    ) {
+      return
+    }
```
colinrotherham: We haven't got code coverage here for when the direction isn't ascending/descending

packages/nhsuk-frontend/src/nhsuk/macros/icon.njk
anandamaryon1: Yes happy with the icons

packages/nhsuk-frontend/src/nhsuk/components/tables/template.njk
frankieroberto: Aha nice! Let's do that then?

packages/nhsuk-frontend/src/nhsuk/macros/icon.njk
colinrotherham: On the topic of things to remember, I pushed https://github.com/nhsuk/nhsuk-frontend/pull/1654/changes/dfd333a1ed7b302ae5a6b45d46285b7781126101 to make sure the new icon (SVG) is generated

frankieroberto: @colinrotherham ok this is done in 0b32c4f949695081ea94febad0f9b653756d5dbe - I've added a comment to remind our future selves that the up-down arrow is derived from both the up and down arrows, but each pushed 4.8 in either direction.

If the arrows are ever changed this’ll need to be re-applied.

packages/nhsuk-frontend/src/nhsuk/components/tables/template.njk
frankieroberto: Switched to CSS in 42360dce4eddd02b18d0d19a4a28eb3c395c0e36 using the `mask` for newer browsers! 

It makes the Javascript and Nunjucks simpler too! 🙌

packages/nhsuk-frontend/src/nhsuk/components/tables/table.mjs
frankieroberto: I've refactored this a bit to use Typescript annotations for the directions, so that bit of the guard clause is no longer necessary.

Also renamed the function a bit to make it clearer.

Other ideas for refactoring welcome!

anandamaryon1:
Couple of bugs I found just now:

1. Icons being clipped when table squeezed (see 6-in-1):
image

Fixed by: [1bc4fab](https://github.com/nhsuk/nhsuk-frontend/pull/1654/commits/1bc4fab821c8ff5550fe1cc1900f585c6ce947f2) (adding `flex-shrink:0`)

----
2. No icon when `sortFirstDirection` is omitted:
image

Not sure on the fix for this? Was this expected with the rejig of Nunjucks options? @frankieroberto 

----
And one new thought:
3. Should we make the icons (slightly) smaller when the font-size goes smaller, on mobile? 
They currently look a bit large to my eye:
image

Eg. 1.5em to 1.25em:
image

frankieroberto:
@anandamaryon1 fixed the second bug in b4cfc288f72ce38d47966786d4bbf8d97dfb0f24.

There are 2 options if `initialSortColumn: true` is set, but `sortFirstDirection` is missing:

1. default to a particular initial sort direction (ascending or descending)
2. don't sort the column at all initially (ie ignore `initialSortColumn: true`)

I think the second option is safer, so I've gone with that.

anandamaryon1:
> @anandamaryon1 fixed the second bug in [b4cfc28](https://github.com/nhsuk/nhsuk-frontend/commit/b4cfc288f72ce38d47966786d4bbf8d97dfb0f24).
> 
> There are 2 options if `initialSortColumn: true` is set, but `sortFirstDirection` is missing:
> 
> 1. default to a particular initial sort direction (ascending or descending)
> 2. don't sort the column at all initially (ie ignore `initialSortColumn: true`)
> 
> I think the second option is safer, so I've gone with that.

Thanks!

I'd have maybe gone with option 1. Since columns already kind of have a default of sorting ascending. But not too precious. 

Any thoughts on Q3, icon sizing?

packages/nhsuk-frontend/src/nhsuk/components/tables/_index.scss
```diff
@@ -8,6 +9,11 @@
 /// @group components/table
 ////
 
+// NHS pages have a grey background, so we need a slightly darker colour for the hover
+// This produces 1.1:1 contrast, the same as GOV.UK’s
+// This is also used by the Task list component
+$sortable-table-hover-colour: color.adjust(nhsuk-colour("grey-5"), $lightness: -6%);
```
colinrotherham: Just to flag that this is identical to the task list hover colour

We should wrap it in `nhsuk-colour-compatible()` and maybe share it as an applied colour?

```suggestion
$nhsuk-task-list-hover-colour: nhsuk-colour-compatible(color.adjust(nhsuk-colour("grey-5"), $lightness: -6%));
```

packages/nhsuk-frontend/src/nhsuk/components/tables/template.njk
```diff
@@ -60,8 +61,14 @@
...
         {%- if item.rowspan %} rowspan="{{ item.rowspan }}"{% endif %}
         {%- if params.responsive %} role="columnheader"{% endif %}
+        {%- if item.sortable %} aria-sort="{{ item.sortFirstDirection if item.initialSortColumn and item.sortFirstDirection else 'none' }}"{% endif %}
+        {%- if item.sortable and item.sortFirstDirection %} data-sort-first-direction="{{ item.sortFirstDirection }}"{% endif %}
         {{- nhsukAttributes(item.attributes) }}>
-        {{ item.html | safe | trim | indent(8) if item.html else item.text }}
+        {% if item.href %}
+          <a href="{{ item.href }}">{{ item.html | safe | trim | indent(8) if item.html else item.text }}</a>
+        {% else %}
+          {{ item.html | safe | trim | indent(8) if item.html else item.text }}
```
colinrotherham: Whilst `indent(8)` is correct, this first line will be indented with 10 spaces

If you can check the HTML output in the code details (on the review app)

packages/nhsuk-frontend/src/nhsuk/components/tables/template.njk
```diff
@@ -60,8 +61,14 @@
         {%- if item.colspan %} colspan="{{ item.colspan }}"{% endif %}
         {%- if item.rowspan %} rowspan="{{ item.rowspan }}"{% endif %}
         {%- if params.responsive %} role="columnheader"{% endif %}
+        {%- if item.sortable %} aria-sort="{{ item.sortFirstDirection if item.initialSortColumn and item.sortFirstDirection else 'none' }}"{% endif %}
```
colinrotherham:
```suggestion
        {%- if item.sortable %} aria-sort="{{ item.sortFirstDirection if item.initialSortColumn and item.sortFirstDirection else "none" }}"{% endif %}
```

packages/nhsuk-frontend/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "nhsuk-frontend",
-  "version": "10.4.2",
+  "version": "11.0.0",
```
colinrotherham: Just a reminder to remove this before we merge

We probably did it for the branch preview

frankieroberto: @colinrotherham shall I change it to 10.5.0?

colinrotherham: We've got 60 commits in this PR so ideally we rebase and drop the one that changed it

colinrotherham: Conflicts need resolving anyway

frankieroberto:
@anandamaryon1 @colinrotherham dropped the icon size down to 20px on mobile: d687d9072817f8a7c2b4f0fdd0e1a8ee2c418809. About right?

anandamaryon1:
> @anandamaryon1 @colinrotherham dropped the icon size down to 20px on mobile: [d687d90](https://github.com/nhsuk/nhsuk-frontend/commit/d687d9072817f8a7c2b4f0fdd0e1a8ee2c418809). About right?

Yep looks good to me, thank you

colinrotherham:
@frankieroberto Just a heads up, I'll be rebasing this PR branch (and others) to fix the conflicts

packages/nhsuk-frontend/src/nhsuk/components/task-list/_index.scss
```diff
@@ -12,7 +12,7 @@
 
 // NHS pages have a grey background, so we need a slightly darker colour for the hover
 // This produces 1.1:1 contrast, the same as GOV.UK’s
-$nhsuk-task-list-hover-colour: nhsuk-colour-compatible(color.adjust(nhsuk-colour("grey-5"), $lightness: -6%));
+$nhsuk-task-list-hover-colour: $nhsuk-hover-colour;
 
 // @deprecated To be removed in v11.0
 $nhsuk-task-list-hover-color: $nhsuk-task-list-hover-colour;
```
anandamaryon1, line 17: With the being deprecated I wasn't sure how to handle this here

colinrotherham: This is fine, it still maps back to the variable we've kept above

packages/nhsuk-frontend/src/nhsuk/core/settings/_colours-applied.scss
```diff
@@ -135,13 +135,22 @@ $nhsuk-reverse-border-colour: nhsuk-tint($nhsuk-brand-colour, 20%) !default;
...
+$nhsuk-hover-colour: rgba(nhsuk-colour("grey-3"), 0.25) !default;
+
+/// Reverse target area hover colour
+///
+/// Used for hover states on transparent areas used to expand target areas, 
+/// on reverse backgrounds
+///
+/// @type Colour
+
+$nhsuk-reverse-hover-colour: nhsuk-shade($nhsuk-brand-colour, 20%) !default;
```
colinrotherham: We should match the 0.25 opacity in both regular and reverse colours

Ideally the contrast should match too

packages/nhsuk-frontend/src/nhsuk/core/settings/_colours-applied.scss
```diff
@@ -135,13 +135,22 @@ $nhsuk-reverse-border-colour: nhsuk-tint($nhsuk-brand-colour, 20%) !default;
...
-/// Input hover colour
+/// Target area hover colour
 ///
-/// Used for hover states on form controls
+/// Used for hover states on transparent areas used to expand target areas
 ///
 /// @type Colour
 
-$nhsuk-hover-colour: nhsuk-colour("grey-3") !default;
+$nhsuk-hover-colour: rgba(nhsuk-colour("grey-3"), 0.25) !default;
```
colinrotherham: We've darkened the brand blue for reverse hover

Shouldn't we lighten the brand blue for this one?

anandamaryon1: No, light blue looks bad on our page background colour

colinrotherham: What about the transparency?

colinrotherham:
Just to flag some findings from this today

* **Setting the Nunjucks `sortFirstDirection` option is ignored for initial sort columns**
  — Users must omit `initialSortColumn: true` for `sortFirstDirection` to be applied on first click

* **Unlike MOJ sortable tables, we moved `aria-sort` into Nunjucks config**
  — This can lead to `aria-sort` being set twice (via Nunjucks config and manually via `attributes`)

* **We allow multiple `aria-sort="none"` at once without throwing an error**
  — [Testing shows issues](https://a11ysupport.io/tech/aria/aria-sort_attribute) with [JAWS announcing "sorted"](https://github.com/FreedomScientific/standards-support/issues/506) [rather than "unsorted"](https://webaim.org/discussion/mail_thread?thread=10651)

* **We should use table cell `format: "numeric"` to sort numbers automatically**
  — No Nunjucks option currently exists for `data-sort-value` and I missed this during testing

colinrotherham:
I've also done some testing with the NHS prototype kit in mind

## The problem with server-side sorting

Our Nunjucks template currently omits `aria-sort` unless the next sort direction is provided, yet:

* Server-side sorting only needs the current sort direction
* Client-side sorting only needs the next sort direction

Client-side sorting can optionally set the current sort direction, but it defaults to `"none"`

As a workaround, we're manually setting `aria-sort` in our server-side examples at present

## The problem with "next sort" direction

As [mentioned above](https://github.com/nhsuk/nhsuk-frontend/pull/1654#issuecomment-4936344188) we're giving our "next sort" direction dual meaning:

* When a column is already sorted, we ignore the "next sort" direction to mean "current sort" instead
* When a column is NOT already sorted, we correctly apply the "next sort" on click

We can fix this by always providing the current sort direction to the column already sorted

We then have enough information to set `aria-sort` server-side

## Reviewing our Nunjucks options

I'd like to propose:

* Moving `sortable: true` to the top-level with a `--sortable` modifier
* Renaming `initialSortColumn: true` (boolean) → `sortDirection: "ascending"` (string)
* Renaming `sortFirstDirection: "descending"` → `sortDirectionNext: "descending"` (string)

I've used the word "next" not "first" to try and indicate that it's what happens next (on click)

What's important is that we have current sort direction, these names are just suggestions

### Before

We lack the necessary information to set `aria-sort` server-side

```njk
{% set head = [
  {
    text: "First name",
    sortable: true,
    sortFirstDirection: "descending"
  },
  {
    text: "Last name",
    sortable: true,
    initialSortColumn: true,
    sortFirstDirection: "descending"
  },
  {
    text: "Company",
    sortable: true,
    sortFirstDirection: "descending"
  },
  {
    text: "Email"
  }
] %}
```

### After

We can now set `aria-sort="ascending"` server-side on the **Last name** column ready for JavaScript

```njk
{% set head = [
  {
    text: "First name",
    sortDirectionNext: "descending"
  },
  {
    text: "Last name",
    sortDirection: "ascending",
    sortDirectionNext: "descending"
  },
  {
    text: "Company",
    sortDirectionNext: "descending"
  },
  {
    text: "Email"
  }
] %}
```

Or for truly server-side only sorting (no handover to JavaScript) `sortDirectionNext` can be omitted:

```patch
  {% set head = [
    {
      text: "First name",
-     sortDirectionNext: "descending"
+     href: "/page?firstName[direction]=descending"
    },
    {
      text: "Last name",
      sortDirection: "ascending",
-     sortDirectionNext: "descending"
+     href: "/page?lastName[direction]=descending"
    },
    {
      text: "Company",
-     sortDirectionNext: "descending"
+     href: "/page?company[direction]=descending"
    },
    {
      text: "Email"
    }
  ] %}
```

colinrotherham:
Naming conventions wise, we pair Nunjucks `describedBy` with `aria-describedby`

Maybe it would be a good idea to pair `sort` with `aria-sort` instead:

```patch
  {% set head = [
    {
      text: "First name",
-     sortDirectionNext: "descending"
+     sortNext: "descending"
    },
    {
      text: "Last name",
-     sortDirection: "ascending",
-     sortDirectionNext: "descending"
+     sort: "ascending",
+     sortNext: "descending"
    },
    {
      text: "Company",
-     sortDirectionNext: "descending"
+     sortNext: "descending"
    },
    {
      text: "Email"
    }
  ] %}
```

frankieroberto:
@colinrotherham  Is your suggestion that sortDirectionNext would update whenever you toggle the sort direction?

If so, we'd then lose the ability to 'reset' to the preferred initial sort direction when you switch to sorting another column and then sort the column you were on before?

edwardhorsford:
I find `sortDirectionNext` a bit confusing. Could it be `defaultSortDirection`?

colinrotherham:
> If so, we'd then lose the ability to 'reset' to the preferred initial sort direction when you switch to sorting another column and then sort the column you were on before?

Don't worry, it's just my proposed rename of `sortFirstDirection` to `sortDirectionNext`

We'll still need to keep track of the initially proposed "next sort" order and the current state

> I find `sortDirectionNext` a bit confusing. Could it be `defaultSortDirection`?

Don't mind at all

We can keep `sortFirstDirection` if you like, I just found it very confusing that it did two things based on whether the column was already sorted or not—see [**The problem with "next sort" direction** above](https://github.com/nhsuk/nhsuk-frontend/pull/1654#issuecomment-4936677695)

>I've used the word "next" not "first" to try and indicate that it's what happens next (on click)
>
>What's important is that we have current sort direction, these names are just suggestions

Names aside, I hope the explanations above are clear

The important bit is knowing both the current sort (for already sorted columns) and next sort order

colinrotherham:
I've migrated the component text to i18n

```mjs
static defaults = Object.freeze({
  i18n: {
    sortAnnouncement: 'Sorted by %{heading} (%{direction})',
    ascending: 'ascending',
    descending: 'descending'
  }
})
```

Notice the `%heading%` placeholders are swapped with `%{heading}` like other components

This means we can now use the built in placeholder replacement:

```mjs
this.$status.textContent = this.i18n.t('sortAnnouncement', {
  heading: $heading.textContent.trim(),
  direction: this.i18n.t(direction)
})
```

colinrotherham:
I've also automatically applied numeric sorting when `format: "numeric"` is set

There's now a new Nunjucks `item.sortValue` macro option too

colinrotherham:
I've renamed `nhsuk-table--with-sortable-columns` to `nhsuk-table--sortable`

It's now set in the HTML via Nunjucks `sortable: true` at the component level

colinrotherham:
Using the [convention above](https://github.com/nhsuk/nhsuk-frontend/pull/1654#issuecomment-4936700210) to pair `aria-sort` with `sort` we can:

* Set either `sort: "ascending"` or `sort: "descending"` on a single sortable column
* Set only  `sort: "none"` on all other columns

```js
sortable: true,
head: [
  {
    text: 'Nation',
    sort: 'ascending'
  },
  {
    text: 'MMR',
    format: 'numeric',
    sort: 'none'
  },
  {
    text: '6-in-1',
    format: 'numeric',
    sort: 'none'
  },
  {
    text: 'Rotavirus',
    format: 'numeric',
    sort: 'none'
  }
],
```

Clicking a `sort: "none"` column will next apply `sort: "ascending"`

But we can alternatively use `sortNext: "descending"` to override that as necessary:

```js
sortable: true,
head: [
  {
    text: 'Nation',
    sort: 'ascending'
  },
  {
    text: 'MMR',
    format: 'numeric',
    sort: 'none',
    sortNext: 'descending'
  },
  {
    text: '6-in-1',
    format: 'numeric',
    sort: 'none',
    sortNext: 'descending'
  },
  {
    text: 'Rotavirus',
    format: 'numeric',
    sort: 'none',
    sortNext: 'descending'
  }
],
```

anandamaryon1:
Thanks @colinrotherham this refactor looks good to me. And lots of extra work on i18n and tests.

The only thing I wonder about is having to add `sort: none` to enable a column to be sorted. Feels a bit counterintuitive, but having `sort: 'ascending'` and `sortNext: 'descending'` make good sense to me. 

It's a tricky one to name. `sort: ready`, `sort: enabled`?

colinrotherham:
>The only thing I wonder about is having to add `sort: none` to enable a column to be sorted. Feels a bit counterintuitive, but having `sort: 'ascending'` and `sortNext: 'descending'` make good sense to me.

Ah I missed from Slack we talked about `sort: true` also working

Although a single column must have a ~`sort: true`~ `sort: "ascending"` named direction

```js
sortable: true,
head: [
  {
    text: 'Nation',
    sort: 'ascending'
  },
  {
    text: 'MMR',
    format: 'numeric',
    sort: true
  },
  {
    text: '6-in-1',
    format: 'numeric',
    sort: true
  },
  {
    text: 'Rotavirus',
    format: 'numeric',
    sort: true
  }
],
```

frankieroberto:
I was about to comment the same thing, `sort: ‘none’` would intuitively disable sorting.

How about `sort: true`?

Could we also remove the need to set `sortable: true` on the table as a whole? The Nunjucks macro could instead auto-detect if any columns are sortable and set the right classes and attributes on the table element?

anandamaryon1:
Yeah `sort: true` is better I think. Makes me think it could be the currently sorted column, but I think it makes sense when seen together with the other values/options. 👍

colinrotherham:
>Could we also remove the need to set `sortable: true` on the table as a whole? The Nunjucks macro could instead auto-detect if any columns are sortable and set the right classes and attributes on the table element?

I'll look into this 👀

I've pushed an update that moves `captionDescriptionText` server-side

The server-side version didn't receive the description that columns could be sorted

(Despite having all the `aria-sort` attributes in place)

colinrotherham:
Still todo:

* Move `aria-sort="none"` into `data-sort="none"`
* Throw JavaScript errors if multiple `aria-sort` are set
* Update the Nunjucks API as above (avoiding manual `aria-sort` for server-side)

colinrotherham:
@anandamaryon1 @frankieroberto Are we definitely happy with the **flash of unsorted content**™ on first load?

* Do we log a JavaScript warning if the HTML sort order doesn't match what JavaScript applies?
* Do we follow checkboxes and radios and use ~`aria-sort`~ `data-aria-sort` until initially sorted?

I know it's a nice feature but it makes `aria-sort` incorrect-until-initially-sorted

anandamaryon1:
@colinrotherham 

On the **flash of unsorted content™**, I hadn't noticed this before, had to throttle 20x to see it. Is this avoidable, when we're adding this as an enhancement? Is the alternative to prevent the table from displaying until the JS is run?

colinrotherham:
Have a look at @frankieroberto's reply here: https://github.com/nhsuk/nhsuk-frontend/pull/1654#issuecomment-4342479227

It's not something the MOJ sortable tables do and we don't have to offer an initial sort

But by design we take potentially unsorted HTML and run an initial sort on it when the JavaScript loads

If we tried to avoid it yeah, we'd have to hide content temporarily

It creates another complication where server-side sorted HTML would be allowed to set `aria-sort`, but content sorted by the browser would need `data-aria-sort` until the initial sort had been applied

frankieroberto:
When does the flash of unsorted content happen? Is that if your HTML sort order doesn't match what you've set the initial sort order to be, and so the JavaScript re-sorts it after it’s loaded?

If so, that feels acceptable - but we could try and make it clear in the guidance that the HTML sort ought should match, in order to avoid this?

colinrotherham:
>When does the flash of unsorted content happen? Is that if your HTML sort order doesn't match what you've set the initial sort order to be, and so the JavaScript re-sorts it after it’s loaded?

It happens whenever an initial sort column is configured

But you only notice it if the HTML doesn't match, yet no screen reader status message is applied

anandamaryon1:
Don't love the idea of hiding the table until it's sorted. And adding an announcement for screen readers sounds like it could be annoying noise 99.9% of the time? 

Adding guidance seems better to me, something like: "To avoid users seeing a sudden swapping of rows as the page loads, ensure the initial sort order matches your HTML." ?

colinrotherham:
New Nunjucks API is now pushed up

To enable a sortable column, either:

* Set Nunjucks `sort: true`
* Set HTML attribute `data-sort`

We no longer use `aria-sort="none"` by default

colinrotherham:
Other than some more tests, this PR is ready for review and testing

colinrotherham:
I've [updated the changelog for release](https://github.com/nhsuk/nhsuk-frontend/pull/1654/commits/a19ee7c0a597d0c0483435b8be22f1d76b929b05) with some notable changes pushed:

* [Add button aria-pressed when sorted after click](https://github.com/nhsuk/nhsuk-frontend/pull/1654/commits/919334204b1f9891a4941ddb15457e12f721c2a2)
* [Move classes into config](https://github.com/nhsuk/nhsuk-frontend/pull/1654/commits/bbdeeaf2dd0b2044ff704450f0d7ce595a2627d1)
* [Set up sortable test fixtures with compact as a variant](https://github.com/nhsuk/nhsuk-frontend/pull/1654/commits/929af8977144d5c17ddcd939cd670d8e4bdcd7e4)

Lots of extra tests added in that last one, e.g. remembering sort order when switching columns

The addition of `aria-pressed="true"` on the button came from [**ARIA: aria-sort attribute** on mdn](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-sort)

frankieroberto (approved):
Tested and all looking good to me!

colinrotherham:
I've adjusted the icon's margin like we do for icon buttons

Frees up 6px per column to help reduce this sort of thing:

Screenshot 2026-07-17 at 17 11 08

paulrobertlloyd:
@colinrotherham Shouldn’t icons be aligned to the ~~top~~ first line of the table header, and not the vertical middle of the header cell?

colinrotherham:
@paulrobertlloyd I'd prefer that

anandamaryon1:
Icon alignment in wrapped header cells:

Centred (current):
Screenshot 2026-07-20 at 09 40 51


First line: 
Screenshot 2026-07-20 at 09 40 46

Oddly, to me, aligning on the first line looks odd here. It works with checkboxes etc but here it feels unbalanced. Maybe it's a left-versus-right positioned thing. When aligned with first line, it doesn't feel as connected with the entire wrapped heading text.

colinrotherham:
I know it's not exactly the same, but Excel always keeps the arrows aligned, even when text is wrapped

Excel sortable columns

frankieroberto:
Centred does look clearer to me (if less neat).  Hopefully it'll be fairly rare for the headers to wrap (if teams choose shorter headings and not too many columns).

anandamaryon1:
Interesting, aligning headings bottom. I can see the advantage of them being closer to their column's content. A con being that it makes the headings less scannable. 

Over on gov one comment has suggested it before: (but no discussion)
https://github.com/alphagov/govuk-design-system-backlog/issues/61#issuecomment-1944505240

Had a go here:
Screenshot 2026-07-20 at 11 46 47

I'm not sure it's needed, but interested if anyone has done it before. I think it's one to pick up another time.

colinrotherham:
Looks loads better. What do you think @paulrobertlloyd?

frankieroberto:
ooh that does look better - particular on the first column where it's not wrapping at all.

colinrotherham:
An overview of the changes and bug fixes from today

I've changed the default `thead th` vertical alignment to bottom:

Table header vertical alignment

<br>
<br>

Plus included the following:

* Add support for table cell `item.align` macro option (e.g. `align: "right"`)
* Add support for table cell `item.width` macro option (e.g. `width: "one-third"`)
* Add support for table cell `format: "string"` for “No data” cells in numeric columns
* Automatically apply `item.align` and `item.format` to entire column from header cell
* Fix missed deprecation `nhsuk-table-responsive__heading`

The 2px negative margin adjustments from https://github.com/nhsuk/nhsuk-frontend/pull/2003 are now in this PR

### Some bugs

I've also included some more bug fixes:

* Prevent `item.colspan` from being applied to responsive tables
* Fix header cell `margin-right` button offset when in last column
* Fix header cell `item.html` content flattened down to plain text

For example, settings `html: "2025 to<br>2026"` becomes `"2025 to2026"` when the JavaScript runs

I've swapped `textContent` for `innerText` to resolve this

As far as I know that's all issues now resolved

frankieroberto (approved):
🚀

PR #1660, opened 2025-10-31
[v10] Add search input component

colinrotherham:
## Description

This PR contributes the search input from @paulrobertlloyd

* [**Search input** examples preview](https://nhsuk-frontend-pr-1660.herokuapp.com/nhsuk-frontend/components/search-input/)

---

Search input

<br>
<br>

Search input with text content

<br>
<br>

Search input with active button

<br>
<br>

Perhaps optionally with button text:

Search input with button text

## Checklist

- [x] CHANGELOG entry


anandamaryon1:
Should we add a large version, like: https://www.nhs.uk/search/results for use on search pages. 

As a custom instance, not suggesting we make large inputs.

colinrotherham:
>Should we add a large version, like: https://www.nhs.uk/search/results for use on search pages.
>
>As a custom instance, not suggesting we make large inputs.

Yeah, happy to

Would we use `--large` as a modifier or pair `nhsuk-search-input` with `nhsuk-search-input-xl` etc?

anandamaryon1:
Hmm, I think `-xl` maybe suggests more that the full rnage of sizes exists, so maybe go with `--large` which is then in keeping with the `--small` buttons and checkboxes.

colinrotherham:
I suppose the next problem is the button

Since the template has a hard-coded `nhsuk-button--small`, how do we remove it using an Nunjucks option?

Our Nunjucks options API would consider `button.classes` as "classes to append" only

anandamaryon1:
So does the `--large` class need to be set by a flag on the component which then adds and removed the relevant classes? 

eg. `large: true`

colinrotherham:
>So does the `--large` class need to be set by a flag on the component which then adds and removed the relevant classes?
>
>eg. `large: true`

@anandamaryon1 Yeah that's my proposal but I've not done it yet

Blocked until we decide in https://github.com/nhsuk/nhsuk-frontend/pull/1801 ([see comment and thread](https://github.com/nhsuk/nhsuk-frontend/pull/1801#discussion_r2789118069))

anandamaryon1:
Should we include a reverse version, for use on a dark backgrounds, and in the header? Eg. for use in a hero header type layout. 

I think probably yes, but it does leave a question around the border-radius. The current header search has rounded edges all around, but our inputs do not. We could make that a customisation for the header specifically.

colinrotherham:
Large and reverse search input examples added btw
https://nhsuk-frontend-pr-1660.herokuapp.com/nhsuk-frontend/components/search-input/

frankieroberto:
I like the idea of having a reversed option.

This looks a bit odd with the white input directly against the light grey button though?

Screenshot 2026-02-24 at 21 27 08

On a slightly wider point, should we make the button optional? 

This could be by allowing `button: false`, or making the button not appear if there is no `button` param, and requiring `button: {icon: true}` to be set at minimum?

I can see use cases where you might want the search field (ie the regular input but with `type: search` and the `nhsuk-search-input__input` class) but either along with some other inputs in a form, or still stand alone but with a regular button?

eg I've used a search bar like this (which could use the conjoined button, but there might be reason not to?):

Screenshot 2026-02-24 at 21 34 38

Or [here’s an example](https://dmd-browser.nhsbsa.nhs.uk/ingredient/search) which has a single search input along with a checkbox, and then the search button:

Screenshot 2026-02-24 at 21 38 02

frankieroberto:
Minor bug - the close button on the large version is not vertically centred:

Screenshot 2026-02-24 at 21 40 56

colinrotherham:
Thanks for feedback @frankieroberto

1. Setting `button: false` will omit the button
2. Fixed the large search input clear icon position
3. Added a grey tint to the reverse search input

We've still got all the [form alignment examples](https://nhsuk-frontend-pr-1660.herokuapp.com/nhsuk-frontend/examples/form-alignment/) for inputs with and without conjoined buttons

The aim of this PR was to bring @paulrobertlloyd's fab search input work into NHS.UK frontend

So would be good to descope anything that's become a bit of a stretch since October

Reverse input button tint

anandamaryon1:
Wonder whether we can align it more closely with the current header styling? 

ie. Making the button background the light blue, icon blue, shadow a darker shade of the light blue.
image

Then, for use in the header, we just add the left-side border-radius.

anandamaryon1:
We discussed this at the Design System Review Panel a few weeks back. I think we've addressed everything that was mentioned there already except for:

* Should we have a version without a button, for instances where the search is filtering realtime, making the button redundant?

Example being the search on the [prototype docs website](https://prototype-kit.service-manual.nhs.uk/).

colinrotherham:
>Wonder whether we can align it more closely with the current header styling?

Let me know if you do, I've parked this PR for now

>* Should we have a version without a button, for instances where the search is filtering realtime, making the button redundant?

Yeah you can use `button: false` to turn off the button

colinrotherham:
Here's the reverse variant [using the colours above](https://github.com/nhsuk/nhsuk-frontend/pull/1660#issuecomment-4091262453)

* [**Search input reverse** example](https://nhsuk-frontend-pr-1660.herokuapp.com/nhsuk-frontend/components/search-input/#reverse)

Hover and active state colours are taken from the header search button, but with default border widths:

Search input with button

colinrotherham:
@paulrobertlloyd @frankieroberto @edwardhorsford Can I ask for your feedback on this for @sarawilcox?

So we can help unblock NHS.UK frontend v10.5.0, but particularly looking at:

* Clear "×" button when input is populated
* Input, label and legend reverse variants
* Search variant ([used with reverse button, see comment](https://github.com/nhsuk/nhsuk-frontend/pull/1660#issuecomment-4178403560))
* Large variant

You'll want to see how this can then be used in the header:

* https://github.com/nhsuk/nhsuk-frontend/pull/1696

But that will be a later breaking change in v11

paulrobertlloyd:
The hover state for the reversed button style seems incongruous with hover styles for other button styles; whereas they are quite subtle, here the change in colour is stark:

Screenshot of reversed search button hover state.

packages/nhsuk-frontend/src/nhsuk/components/button/_index.scss
```diff
@@ -188,7 +188,8 @@ $button-shadow-size: $nhsuk-button-shadow-size;
     );
   }
 
-  .nhsuk-button--login {
+  .nhsuk-button--login,
+  .nhsuk-button--search {
```
colinrotherham, line 192: @paulrobertlloyd Could you shed some light on the choice of `:active` colour for the header search button, particularly the blue button on the white header?

You'll see I've added a `"search"` variant as an alias of `"login"`

If you wondered why, they only seemed to differ on `:active` state:

Image

paulrobertlloyd: What might it look like if:

- Blue header gets reversed button style (which is updated to have a blue tinted shadow/hover state instead of grey)
- White header uses login button style (which I’m guessing is unchanged from what’s implemented)

colinrotherham: Like in here? It's stacked on top:

* https://github.com/nhsuk/nhsuk-frontend/pull/1696

colinrotherham: (The blue tinted button would be a lovely replacement for the grey)

colinrotherham:
> The hover state for the reversed button style seems incongruous with hover styles for other button styles; whereas they are quite subtle, here the change in colour is stark:
> 
> Screenshot of reversed search button hover state.

@paulrobertlloyd Yeah I did too, but I've taken it from the header

Although minus the border width override we use here:

Header button hover

<br>
<br>

Any other ideas?

Ananda didn't suggest a hover state

colinrotherham:
After merging https://github.com/nhsuk/nhsuk-frontend/pull/1898 this is what the reverse search input looks like:

Search input with button

packages/nhsuk-frontend/src/nhsuk/components/search-input/_index.scss
```diff
@@ -0,0 +1,114 @@
...
+
+  .nhsuk-search-input__input {
+    flex: 0 1 auto;
+    margin-right: -$nhsuk-border-width-form-element;
+
+    .nhsuk-search-input--large & {
+      height: nhsuk-px-to-rem(52px);
+      padding-right: nhsuk-px-to-rem(nhsuk-spacing(2));
+      padding-left: nhsuk-px-to-rem(nhsuk-spacing(2));
+    }
```
frankieroberto: I wonder if the `--large` variant should be uprooted from the search input onto the general input and button components? 🤔

colinrotherham: Possibly a bit of a stretch, we only really did it for:
https://www.nhs.uk/search/results

But for buttons, they're already large by default? 😮

frankieroberto: Oh of course, the button is large already 🤦.

Still might be worth considering moving the large modifier to the generic text input? 🤔

edwardhorsford: Aside - my team is going to swap all our buttons to a new 'medium' variant as the large one is just too large.

packages/nhsuk-frontend/src/nhsuk/components/button/_index.scss
frankieroberto: Not a blocker, but I wonder if having the blue button aliased as both `--login` and `--search` is a indicator that we should rename it `--brand` or something (as it uses the brand colour)?

Then any sites which change the brand colour variable (eg [this one](https://www.islingtoncentralmedicalcentre.org) which uses purple!) would get the colour?

Or is that too weird? One for the future anyhow.

packages/nhsuk-frontend/src/nhsuk/components/input/_index.scss
```diff
@@ -63,6 +63,10 @@
     border-color: $nhsuk-error-colour;
   }
 
+  .nhsuk-input--reverse {
```
frankieroberto: Worth adding a comment on this to remind our future-selves? I can never remember if transparent borders take the background colour of the box or the surrounding page.

Something like

```suggestion
  // Reversed inputs on a blue background display with no visible border - the transparent border makes the border white. However we still need the border as the border turns black on focus.
  .nhsuk-input--reverse {
```

packages/nhsuk-frontend/src/nhsuk/components/button/_index.scss
colinrotherham: Yeah good shout

See how we're going to need people to pick a button colour once search input is in the header via https://github.com/nhsuk/nhsuk-frontend/commit/8848155b0ea57042179989c6cbbadcdfb3a14806

packages/nhsuk-frontend/src/nhsuk/components/search-input/_index.scss
frankieroberto: Ignore what said about the button - should we move the `--large` variant from `.nhsuk-search-input` to `.nhsuk-input`?

Feels like that’s be more flexible? Maybe there are some kiosk mode type interfaces where you just want a single larger input but it's not for searching? 🤷 

```
    .nhsuk-input--large & {
      height: nhsuk-px-to-rem(52px);
      padding-right: nhsuk-px-to-rem(nhsuk-spacing(2));
      padding-left: nhsuk-px-to-rem(nhsuk-spacing(2));
    }
```

packages/nhsuk-frontend/src/nhsuk/components/search-input/_index.scss
```diff
@@ -0,0 +1,114 @@
...
+
+    &::-webkit-search-cancel-button {
+      align-self: center;
+
+      width: nhsuk-px-to-rem($nhsuk-button-search-icon-size);
+      height: nhsuk-px-to-rem($nhsuk-button-search-icon-size);
+      margin: 0;
+
+      background-image: url("data:image/svg+xml,<svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 24 24' fill='%23212b32'><path d='M13.41 12l5.3-5.29a1 1 0 1 0-1.42-1.42L12 10.59l-5.29-5.3a1 1 0 0 0-1.42 1.42l5.3 5.29-5.3 5.29a1 1 0 0 0 0 1.42 1 1 0 0 0 1.42 0l5.29-5.3 5.29 5.3a1 1 0 0 0 1.42 0 1 1 0 0 0 0-1.42z'/></svg>");
```
frankieroberto: This looks similar to the `cross` icon in `macros/icon.njk` but the path is different. Is that deliberate?

Do we have a way of defining it in one place and using it in both somehow, to avoid them drifting out of sync?

packages/nhsuk-frontend/src/nhsuk/components/search-input/_index.scss
```diff
@@ -0,0 +1,114 @@
...
+
+    // Hide search input icons and buttons (WebKit, Blink)
+    &::-webkit-search-decoration,
+    &::-webkit-search-results-button,
+    &::-webkit-search-results-decoration,
+    &:placeholder-shown::-webkit-search-cancel-button {
+      display: none;
+    }
+
+    &::-webkit-search-cancel-button {
```
frankieroberto: Could add a comment here like in GOVUK publishing components? https://github.com/alphagov/govuk_publishing_components/pull/3673/changes#diff-834e17c3187fabdbe357ed1a1b58b7a11b8711fb1d52db1f930f2f85a806f135R95

```suggestion
      // Note: this is a non-standard CSS feature and will not do anything in Firefox. https://bugzilla.mozilla.org/show_bug.cgi?id=1417753
    &::-webkit-search-cancel-button {
```

packages/nhsuk-frontend/src/nhsuk/components/search-input/_index.scss
```diff
@@ -0,0 +1,114 @@
...
+      align-self: center;
+
+      width: nhsuk-px-to-rem($nhsuk-button-search-icon-size);
+      height: nhsuk-px-to-rem($nhsuk-button-search-icon-size);
```
frankieroberto: should the cancel icon (cross) also scale up when the input is large, like the magnifying glass icon does? 🤔 

Image

vs

Image

packages/nhsuk-frontend/src/nhsuk/components/search-input/macro-options.mjs
```diff
@@ -0,0 +1,220 @@
+export const name = 'Search input'
+
+/**
+ * Nunjucks macro option params
+ *
+ * @satisfies {{ [param: string]: MacroParam }}
+ */
+const options = {
```
frankieroberto, line 8: Is there any way to inherit all of these options from the `input` macro? Otherwise we’ll have to remember to keep em in sync.

(We have the same issue with `characterCount` though so this isn’t new)

packages/nhsuk-frontend/src/nhsuk/components/search-input/template.njk
```diff
@@ -0,0 +1,100 @@
...
+    classes: classNames,
+    attributes: params.formGroup.attributes,
+    beforeInput: params.formGroup.beforeInput,
+    afterInput: {
+      html: afterInputHtml
+    }
+  },
+  inputWrapper: {
+    classes: "nhsuk-search-input__wrapper"
+  },
```
frankieroberto, line 57: We should let users add additional classes and attributes here using `inputWrapper.classes` and `inputWrapper.attributes` (both already supported by the input component).

frankieroberto (approved):
Looking good! Minor comments and suggestions only. No blockers.

packages/nhsuk-frontend/src/nhsuk/components/search-input/macro-options.mjs
colinrotherham: Not at the moment. They'll all need a different `released: '10.5.0'` property too 😔

packages/nhsuk-frontend/src/nhsuk/components/button/_index.scss
colinrotherham: Updated in 26a1f996e0225549472630d1ff742ca7e8e5c202

packages/nhsuk-frontend/src/nhsuk/components/search-input/_index.scss
colinrotherham: Resolved in c3008f8c2e8dfe7b55a289ae518784c5131edd54

packages/nhsuk-frontend/src/nhsuk/components/input/_index.scss
colinrotherham: I've removed all `--reverse` styling for form controls (from this PR) and split them out into:

* https://github.com/nhsuk/nhsuk-frontend/pull/1914

packages/nhsuk-frontend/src/nhsuk/components/search-input/_index.scss
colinrotherham: Well spotted, no this wasn't deliberate

The code for search input, including the icon, was lifted from @paulrobertlloyd (thank you 🙏)

I've pushed up a new version that uses the `"cross"` icon via SVG file, made possible using:

* https://github.com/nhsuk/nhsuk-frontend/pull/1915

packages/nhsuk-frontend/src/nhsuk/components/search-input/_index.scss
colinrotherham: Thanks, good suggestion

I've [moved vendor pseudo-elements like `::webkit-*` to `.nhsuk-input`](https://github.com/nhsuk/nhsuk-frontend/pull/1660/commits/24291822110ecdb0f1d55a1c699cf1929a25ac2a) so they can be inherited

packages/nhsuk-frontend/src/nhsuk/components/search-input/_index.scss
colinrotherham: Yeah probably should do, I've scaled the large input icon and font size:

* [**Search input** component examples](https://nhsuk-frontend-pr-1660.herokuapp.com/nhsuk-frontend/components/search-input/)

packages/nhsuk-frontend/src/nhsuk/components/search-input/template.njk
colinrotherham: Yeah good point. Similarly with password input we don't expose this either

I'll take a look

frankieroberto: Could be a follow-up PR, or one for the backlog?

packages/nhsuk-frontend/src/nhsuk/components/search-input/_index.scss
frankieroberto: Nice.

packages/nhsuk-frontend/src/nhsuk/components/search-input/template.njk
colinrotherham: I've fixed `disabled`, `inputWrapper` and others in:

* https://github.com/nhsuk/nhsuk-frontend/pull/1916

colinrotherham:
Not specific to search input, but I've [added high contrast (forced colour) fixes for our icons](https://github.com/nhsuk/nhsuk-frontend/pull/1660/changes/1a383fe60e7390612624ec374b758b0552227239):

Search input with high contrast colours

PR #1669, opened 2025-11-04
Remove pixel font sizes where unnecessary

colinrotherham:
## Description

This PR ensures `nhsuk-font()` and `nhsuk-font-size()` output rem units only to align with GOV.UK Frontend

It was missed from NHS.UK frontend v10.0.0 yet [only affects IE 8](https://caniuse.com/rem) etc so can be considered non-breaking

```patch
  .foo {
-   font-size: 12px;
    font-size: 0.75rem;
    line-height: 1.25;
  }
  @media (min-width: 30em) {
    .foo {
-     font-size: 14px;
      font-size: 0.875rem;
      line-height: 1.42857;
    }
  }
```

## Checklist

- [x] CHANGELOG entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,14 @@ The previous class name is deprecated and will be removed in a future release.
 
 This change was introduced in [pull request #1643: Add small buttons and inline form groups](https://github.com/nhsuk/nhsuk-frontend/pull/1643).
 
+#### Rename Sass variable for base font size
+
+If you use the Sass `$nhsuk-base-font-size` variable, you should rename it to `$nhsuk-root-font-size`.
+
+The previous name is deprecated and will be removed in a future release.
+
+This change was introduced in [pull request #1669: Remove pixel font sizes where unnecessary](https://github.com/nhsuk/nhsuk-frontend/pull/1669).
+
 ### :recycle: **Changes**
 
 #### Remove global box sizing reset


packages/nhsuk-frontend/src/nhsuk/core/elements/_page.scss
```diff
@@ -39,9 +39,6 @@ body {
   color: $nhsuk-text-colour;
   background-color: $nhsuk-body-background-colour;
 
-  font-size: $nhsuk-base-font-size;
-  line-height: nhsuk-line-height($nhsuk-base-line-height, $nhsuk-base-font-size);
```
anandamaryon1, line 43: This causes a slight layout shift, which isn't a big issue, but is there good reason to remove this default `line-height`?

packages/nhsuk-frontend/src/nhsuk/core/settings/_globals.scss
```diff
@@ -20,7 +20,8 @@ $nhsuk-include-font-face: true !default;
 
 /// Font sizing and spacing
 
-$nhsuk-base-font-size: 16px !default;
+$nhsuk-root-font-size: 16px !default;
+$nhsuk-base-font-size: $nhsuk-root-font-size;
```
anandamaryon1, line 24: Assume keeping `base` is for backwards compatability?

packages/nhsuk-frontend/src/nhsuk/core/elements/_page.scss
colinrotherham: I noted the slight shift too but if you think we've missed any `font-size` and `line-height` pairs anywhere else let's add them in. They'll all be relative to the root

But yeah it was wrong of us to hard code it here

Just like with GOV.UK Frontend, we want device settings, assistive software and custom stylesheets to configure their own "root" sizes. That way, we're only ever configuring things "relative to root"

Otherwise things like this can happen:

Line height too small

colinrotherham: Worth noting that we know we're not perfect yet, this is just part of it

More of our hard-coded pixel values should follow the [approach we took for icons](https://github.com/nhsuk/nhsuk-frontend/commit/701146eaa659be1ab9a0cffe4497853869783842)

```patch
-   width: 28px;
-   height: 28px;
+   width: nhsuk-px-to-rem($nhsuk-icon-size);
+   height: nhsuk-px-to-rem($nhsuk-icon-size);
```

packages/nhsuk-frontend/src/nhsuk/core/settings/_globals.scss
colinrotherham: It's noted as a deprecated change in the CHANGELOG (and means we match GOV)

packages/nhsuk-frontend/src/nhsuk/core/elements/_page.scss
anandamaryon1: Thanks @colinrotherham that makes sense 👍 

It's a tiny shift, I don't think it's worth trying to offset it.

PR #1682, opened 2025-11-12
Add form controls states example page

colinrotherham:
## Description

This PR adds a test page for all form controls and their states (hover, focus etc)

It was suggested in https://github.com/nhsuk/nhsuk-frontend/issues/1066#issuecomment-3510352510 when the issue was closed

* [**Form controls states** example page](https://nhsuk-frontend-pr-1682.herokuapp.com/nhsuk-frontend/examples/form-controls-states/)

Screenshot showing small form controls and their states

## Checklist

- [ ] Tested against our [testing policy](https://github.com/nhsuk/nhsuk-frontend/blob/master/docs/contributing/testing.md) (Resolution, Browser & Accessibility)
- [x] Follows our [coding standards and style guide](https://github.com/nhsuk/nhsuk-frontend/blob/master/docs/contributing/coding-standards.md)
- [ ] CHANGELOG entry

PR #1683, opened 2025-11-12
Update Nunjucks macro options for nested items

colinrotherham:
## Description

This PR makes some Nunjucks macro option consistency changes for nested items:

- Action link and skip link `html` option
- Checkboxes and radios item `classes` option
- Contents list item `html`, `classes` and `attributes` options
- Summary list action item `classes` and `attributes` options

I've also had another review of Markdown formatting for attribute names

## Checklist

- [x] CHANGELOG entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,17 @@ You can now add inline buttons to text inputs and select menus using the `formGr
 
 This was added in [pull request #1643: Add small buttons and inline form groups](https://github.com/nhsuk/nhsuk-frontend/pull/1643).
 
+#### Updated Nunjucks macro options for nested items
+
+For consistency with other components with nested items, we’ve added new Nunjucks macro options:
+
+- Action link and skip link `html` option
+- Checkboxes and radios item `classes` option
+- Contents list item `html`, `classes` and `attributes` options
+- Summary list action item `classes` and `attributes` options
+
+This was added in [pull request #1683: Update Nunjucks macro options for nested items](https://github.com/nhsuk/nhsuk-frontend/pull/1683).
+
 ### :wastebasket: **Deprecated features**
 
 #### Rename input wrapper HTML class


packages/nhsuk-frontend/src/nhsuk/components/checkboxes/macro-options.mjs
```diff
@@ -201,6 +201,11 @@ export const params = {
         required: false,
         description: 'If `true`, checkbox will be disabled.'
       },
+      classes: {
+        type: 'string',
+        required: false,
+        description: 'Classes to add to the checkbox input tag.'
+      },
```
frankieroberto, line 208: Huh, so we already supported this but forgot to document it... 🫠

packages/nhsuk-frontend/src/nhsuk/components/action-link/template.njk
```diff
@@ -1,11 +1,18 @@
...
 {% from "nhsuk/macros/icon.njk" import nhsukIcon %}
 
-<a {%- if params.id %} id="{{ params.id }}"{% endif %} class="nhsuk-action-link
-  {%- if params.classes %} {{ params.classes }}{% endif %}" href="
-  {%- if params.href %}{{ params.href }}{% else %}#{% endif %}"
+{#- Set classes for this component #}
+{%- set classNames = "nhsuk-action-link" -%}
+
+{%- if params.classes %}
+  {% set classNames = classNames ~ " " ~ params.classes %}
```
frankieroberto, line 8: Any reason not to use `+` here? I've never seen the `~` operator before for Nunjucks, and it's not listed on https://mozilla.github.io/nunjucks/templating.html

packages/nhsuk-frontend/src/nhsuk/components/contents-list/template.njk
```diff
@@ -1,20 +1,30 @@
...
+    {% if item.current %}
+      <span class="nhsuk-contents-list__current {%- if item.classes %} {{ item.classes }}{% endif %}" {{- nhsukAttributes(item.attributes) }}>
+        {{- item.html | safe | trim | indent(6) if item.html else item.text -}}
+      </span>
+    {% else %}
+      <a class="nhsuk-contents-list__link {%- if item.classes %} {{ item.classes }}{% endif %}" href="{{ item.href }}" {{- nhsukAttributes(item.attributes) }}>
+        {{- item.html | safe | trim | indent(6) if item.html else item.text -}}
+      </a>
+    {% endif %}
+    </li>
```
frankieroberto, line 27: Feels to me that `item.classes` should be added to the `li` here, especially as "item" matches "list item"?

It’s also the most flexible, as you can do descendent selectors in CSS much easier than the other way around (although we now have `:has()`).

packages/nhsuk-frontend/src/nhsuk/components/checkboxes/template.njk
```diff
@@ -32,8 +32,9 @@
     {% set itemHintId = itemId ~ "-item-hint" if hasHint else "" %}
     {% set itemDescribedBy = describedBy if not hasFieldset else "" %}
     {% set itemDescribedBy = (itemDescribedBy ~ " " ~ itemHintId) | trim %}
+    {% set itemClasses = "nhsuk-checkboxes__input" ~ (" " ~ item.classes if item.classes) %}
     <div class="nhsuk-checkboxes__item">
-      <input class="nhsuk-checkboxes__input" id="{{ itemId }}" name="{{ itemName }}" type="checkbox" value="{{ item.value }}"
+      <input class="{{ itemClasses }}" id="{{ itemId }}" name="{{ itemName }}" type="checkbox" value="{{ item.value }}"
```
frankieroberto, line 37: Part of me thinks that `item.classes` should be added to the `nhsuk-checkboxes__item` node. However, the other `item.` options are applied to the `input`, so I think it's ok (if unintuitive) here.

packages/nhsuk-frontend/src/nhsuk/components/checkboxes/macro-options.mjs
colinrotherham: Ahh no this is the documentation for the new option ([see commit changes](https://github.com/nhsuk/nhsuk-frontend/pull/1683/commits/ddadd7d9a1c8d4fc788202b5dcd33cf64807e475))

packages/nhsuk-frontend/src/nhsuk/components/action-link/template.njk
colinrotherham: We took a bit of a call to avoid weird and wonderful Jinja bugs where `undefined` or `null` could blow up. The process to upgrade the Jinja templates is quite manual so this was a huge help

Jinja needs `~` to make sure each joined thing is stringified

Whereas Nunjucks `+` follows JavaScript already, e.g. `undefined` → `'undefined'`

We've used `~` instead of `+` since https://github.com/nhsuk/nhsuk-frontend/pull/1635/commits/299bd4476864aef935a933b0b110f732ac830804 in

* https://github.com/nhsuk/nhsuk-frontend/pull/1635

packages/nhsuk-frontend/src/nhsuk/components/checkboxes/template.njk
colinrotherham: Sorry, yeah this is copied from https://github.com/nhsuk/nhsuk-frontend/pull/1682#discussion_r2518912843

I've applied it to the input so that `item.id` and `item.classes` stay on the same element

Same as  `item.name`, `item.value`, `item.checked` and `item.disabled`

packages/nhsuk-frontend/src/nhsuk/components/contents-list/template.njk
colinrotherham: Did you see I brought over `item.classes` and `item.attributes` for summary list though?

Appreciate it's a different component but GOV.UK Frontend puts both on the `<a>` not the `<li>`

packages/nhsuk-frontend/src/nhsuk/components/checkboxes/macro-options.mjs
frankieroberto: Sorry, I could’ve sworn I hadn't seen this in the macro-options changes.

packages/nhsuk-frontend/src/nhsuk/components/contents-list/template.njk
frankieroberto: True, and we already have `item.href`!

PR #1684, opened 2025-11-14
Fix inconsistent card BEM modifiers etc

colinrotherham:
## Description

This PR simplifies HTML markup for the card component so each card variant is clearly identified.

The changes are 100% backwards compatible as follows:

- Add the missing `nhsuk-card--primary` modifier class to primary cards.
- Remove the unnecessary `nhsuk-card__heading--feature` modifier class from feature card headings.
- Remove the unnecessary `nhsuk-card__content--feature` modifier class from feature card content.
- Remove the unnecessary `nhsuk-card__content--primary` modifier class from primary card content.
- Remove the unnecessary `nhsuk-card__content--secondary` modifier class from secondary card content.
- Rename the `<div class="nhsuk-card--care__heading-container"` class attribute to match `<div class="nhsuk-card__heading-container"`.
- Rename the `<div class="nhsuk-card--care__heading"` class attribute to match `<div class="nhsuk-card__heading"`.

The previous class names are deprecated and will be removed in a future release.

## Checklist

- [x] CHANGELOG entry

### DIFF
--- a/packages/nhsuk-frontend/src/nhsuk/components/card/_index.scss
+++ b/packages/nhsuk-frontend/src/nhsuk/components/card/_index.scss
@@ -76,6 +76,7 @@ $card-border-hover-colour: $nhsuk-card-border-hover-colour;
   }
 
   .nhsuk-card__heading-container,
+  // Deprecated, to be removed in v11.0
   .nhsuk-card--care__heading-container {
     @include nhsuk-responsive-padding(5);
   }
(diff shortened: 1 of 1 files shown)


paulrobertlloyd:
Should we add comments where the deprecated class names exist, or is it enough to have them captured elsewhere (and if so, is that place https://github.com/nhsuk/nhsuk-frontend/pull/1553)?

colinrotherham:
Great idea @paulrobertlloyd

I've pushed up https://github.com/nhsuk/nhsuk-frontend/pull/1684/commits/05280879c4bba8367466b7d8eb0e3c54dc8ef297

anandamaryon1:
Would it be too weird to have secondary cards without primary cards? 

Feels slightly odd to me that primary means "with chevron". 

We could maybe remove primary and instead have:
`nhsuk-card--clickable`
`nhsuk-card--with-chevron`

I suppose having it as `primary` communicates that it's mutually exclusive to secondary… So keep as is?

(Sorry, thinking out loud)

colinrotherham:
Maybe, in another PR?

For this PR I've limited the scope to tidying up what we already have

e.g. All cards should set the modifier (primary, secondary etc) at the top-level not the content or heading container

colinrotherham:
@anandamaryon1 Do we definitely need to hold this back for v10.3.0?

It's backwards compatible and no new guidance is needed so it's all good for v10.2.0 already

anandamaryon1:
Fair enough, I was thinking it logically sits with the summary card updates which will likely be 10.3.0 but I suppose it's better to get things out when they're ready, since it's not actually dependent.

PR #1685, opened 2025-11-14
[v10] Add card enhancements to summary list, table and warning callout

colinrotherham:
## Description

This PR includes the following:

1. Add support for card `actions`, e.g. summary card
2. Add support for summary list `card` wrapper
3. Add support for table `card` wrapper
4. Add card `headingVisuallyHiddenText` Nunjucks param
5. Add card `headingSize` Nunjucks param
6. Add hero `headingSize` Nunjucks param
7. Warning callout uses card component

Screenshot 2025-11-14 at 12 01 00

<br>
<br>

Screenshot 2025-11-14 at 12 01 30

## Checklist

- [x] CHANGELOG entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -200,7 +200,7 @@ This change was introduced in [pull request #1655: Add support for Dynamic Type 
 
 For consistency with other components, the HTML and Nunjucks options for tables as a panel have changed. The previous names are deprecated and will be removed in a future release.
 
-If you're using the `table` Nunjucks macro in your service, you should:
+If you're using the `table` Nunjucks macro with the `panel` option, you should migrate to the feature card enhancement:
 
 - replace the `heading` option with the nested `card.heading` option
 - replace the `headingLevel` option with the nested `card.headingLevel` option
@@ -227,7 +227,7 @@ If you're using the `table` Nunjucks macro in your service, you should:
   }) }}
 ```
 
-If you are not using Nunjucks macros, update your HTML markup using the [card examples in the NHS digital service manual](https://service-manual.nhs.uk/design-system/components/card) as follows:
+If you are not using Nunjucks macros, update your HTML markup using the [table as a card (feature) example on the NHS.UK frontend review app](https://nhsuk.github.io/nhsuk-frontend/components/tables/as-a-card-feature/) as follows:
 
 - Add the wrapper `<div class="nhsuk-card nhsuk-card--feature"> </div>`
 - Rename the panel `<div class="nhsuk-table__panel-with-heading-tab"` class attribute to match `<div class="nhsuk-card__content"`


colinrotherham:
Unlike the GOV.UK approach, this just uses our [NHS.UK frontend card component](https://service-manual.nhs.uk/design-system/components/card)

Which means all the usual card options work 😬


Screenshot 2025-11-14 at 12 09 27

<br>
<br>

Screenshot 2025-11-14 at 12 10 08

edwardhorsford:
I think this is a nice, pragmatic addition. Will review thoroughly next week.

I was always unsure about the GDS summary card (and the ability to have generic actions) being tied exclusively to the summary list. What if you want to show a table with actions? or two summary lists?

colinrotherham:
>What if you want to show a table with actions? or two summary lists?

Definitely. I've made sure we can pass the `actions` option to `card()` too

E.g. Have a look at the [**Card basic with summary list and actions** preview](https://nhsuk-frontend-pr-1685.herokuapp.com/nhsuk-frontend/components/card/#basic-with-summary-list-and-actions)

Card wrapping example

<br>
<br>

With the following Nunjucks code:

```njk
{% call card({
  heading: 'Regional Manager',
  headingLevel: 3,
  actions: {
    items: [
      {
        text: 'Delete',
        href: '#'
      },
      {
        text: 'Withdraw',
        href: '#'
      }
    ]
  }
}) %}
  {{ summaryList({
    rows: [
      {
        key: {
          text: 'Name'
        },
        value: {
          text: 'Karen Francis'
        }
      },
      {
        key: {
          text: 'Date of birth'
        },
        value: {
          text: '15 March 1984'
        }
      }
    ]
  }) }}
{% endcall %}
```

anandamaryon1:
Thanks for this @colinrotherham. 

On the design, sharing some options for consideration and discussion.

At a minimum I think we should give slightly less margin on mobile (more padding on desktop) to help group the heading with it's actions. And apply `no-visited-state` to the action links. 

| Before | After |
| -------- | ------- |
| localhost_3000_nhsuk-frontend_examples_summary-cards_ (6) | localhost_3000_nhsuk-frontend_examples_summary-cards_ (5) |

And possibly take it further by adding a background or border to the header:

| Margin | Border-bottom | White background |  Grey background |
| -------- | -------- | ------- | ------- |
| localhost_3000_nhsuk-frontend_examples_summary-cards_ (8) | localhost_3000_nhsuk-frontend_examples_summary-cards_ (10) |  localhost_3000_nhsuk-frontend_examples_summary-cards_ (11) | localhost_3000_nhsuk-frontend_examples_summary-cards_ (12) |

I'm counting out the grey background – too heavy. 

Interested to hear other's thoughts on the options.

edwardhorsford:
Agreed on no-visited status.

I quite like the border-bottom one - though the margins feel too chunky. For the casework service I work on this would end up taking too much space. This could possibly be addressed if we later added general support for compact variants in the design system.

frankieroberto:
@anandamaryon1 the margin tweaks look good to me. The border-bottom option is interesting - it makes the header clearer, but maybe all the different lines get confusing? 

Worth trying having the border-bottom on the header paired with no-borders on the summary list?

Is it also worth considering only supporting a single action on the header? Whilst I can imagine that supporting 2+ actions might be useful in some cases, it’s hard to avoid it wrapping awkwardly depending on text length and viewport width. An alternative would be to have a single header row link which could then lead to additional actions (eg if the label was 'Edit' but that took you to a page where you could edit or delete from).

frankieroberto:
I had a quick play with trying an idea to move the actions up so they're offset from the container box, like we do with the feature card:

Screenshot 2025-11-18 at 12 17 36

(it's not perfectly lined up but you get the idea)

I don't think it quite works though, somehow.

colinrotherham:
Regarding a "no visited" state, what if the card action was a preview link?

You'd no longer be able to see which cards you'd previewed in a previous visit

I'd prefer to support `nhsuk-link--no-visited-state` in the usual way—same as row actions

edwardhorsford:
@colinrotherham you're right, that's probably a good reason not to default it. We could make sure that any guidance notes that you're likely to want to set it for change links though.

anandamaryon1:
Fair point on the `nhsuk-link--no-visited-state`, I suppose in real usage it'll be much rarer to see them in the visited state anyway. 

On the design, I'm thinking the enhanced margins might be enough. The border bottom adds more noise (lots of lines) and takes up more space. 

Happy for me to push some tweaks @colinrotherham?

anandamaryon1:
Would it be too naughty (risky) to put `flex-shrink: 0;` on `.nhsuk-card__actions`? To avoid actions being wrapped when the heading should wrap instead? Obviously if someone used loads of actions, it'd be bad…

image

Is there a less aggressive way to do it, to just _prefer_ giving space to the actions?

anandamaryon1:
Pushed 2 commits to:

* increase card actions margin (decrease on mobile)
* add padding-left to card actions, to prevent heading and actions touching
* add flex basis and grow, to give priority to actions over headings, to prevent needless wrapping of actions

packages/nhsuk-frontend/src/nhsuk/components/card/_index.scss
```diff
@@ -93,6 +95,89 @@ $card-border-hover-colour: $nhsuk-card-border-hover-colour;
...
+    }
+  }
+
+  .nhsuk-card__actions {
+    display: flex;
+
+    flex-wrap: wrap;
+    row-gap: 10px;
+
+    margin: - nhsuk-spacing(1) 0 nhsuk-spacing(3);
```
colinrotherham: Can't remember if this causes issues, do you mind using a negative spacing value?

```suggestion
    margin: nhsuk-spacing(-1) 0 nhsuk-spacing(3);
```

anandamaryon1: Thought I'd tried that first but 🤷  
Works and is nicer to read – done, thanks.

anandamaryon1:
Also want to slightly reduce the margin-top on summary list 'Change' links, to group them better. Is that best in a separate PR? 

| Before    | After |
| -------- | ------- |
| localhost_3000_nhsuk-frontend_components_summary-list_ |  localhost_3000_nhsuk-frontend_components_summary-list_ (1) |

colinrotherham:
@anandamaryon1 Let's do it in this PR

Some [unnecessary margins were removed](https://github.com/nhsuk/nhsuk-frontend/pull/1685/commits/50746ddf407c26eb22d776b5631609d99cd38950) already (e.g. from the inner bottom of cards)

To help with testing I've pushed up all the different card types temporarily:
https://nhsuk-frontend-pr-1685.herokuapp.com/nhsuk-frontend/components/summary-list/#as-a-card-feature-with-actions

I know they won't officially be supported, but we needed to address some knock-on flex issues like:

Feature cards

<br>
<br>

anandamaryon1:
Thanks for those @colinrotherham. Did you fix the flex issue on the feature cards? (I don't see it)

I've pushed some small tweaks based on your examples of different card types. 

@frankieroberto, I'm reluctant to add too much specific styling for unsupported card + actions types, and agree that looks a bit off. I've left it as shown in prior screenshots in this discussion, where the actions just sit in the card, which is a bit odd but works. 

I'm happy with this now but keen for any further thoughts before we say it's done. I'll then do further testing and start thinking about guidance and examples in the design system.

colinrotherham:
>I'm happy with this now but keen for any further thoughts before we say it's done. I'll then do further testing and start thinking about guidance and examples in the design system.

@anandamaryon1 All your tweaks look great and I'm happy to call it done—ready for more testing

In light of https://github.com/nhsuk/nhsuk-frontend/pull/1728 I've tightened the scope of all action styles

Would be great to get this into v10.3.0 along with https://github.com/nhsuk/nhsuk-frontend/pull/1734

colinrotherham:
I've rebased this on top of https://github.com/nhsuk/nhsuk-frontend/pull/1734 so we can test both together

colinrotherham:
We should deprecate `nhsuk-table__panel-with-heading-tab` (i.e. legacy card enhancement for tables) too

anandamaryon1:
Tested on browsers, devices and assistive technologies [listed in our testing guidance](https://github.com/nhsuk/nhsuk-frontend/blob/main/docs/contributing/browser-support.md#browser-support). Found no issues. 

Still to do: test on Dragon and JAWS.

anandamaryon1:
Tested on JAWS and Dragon. 

Only thing we notice is that JAWS reads out "parens" around the heading eg. "Delete _left paren_ Regional Manager _right paren_." 

Which is annoying but down to how verbose JAWS is by default, and can be changed in settings. So I think we can leave as is.

CHANGELOG.md
```diff
@@ -186,6 +196,85 @@ Please carefully review all custom components to make sure they scale dynamicall
...
++     classes: "nhsuk-u-margin-bottom-8"
++   },
+-   tableClasses: "nhsuk-u-margin-bottom-0",
++   classes: "nhsuk-u-margin-bottom-0",
+    head: [],
+    rows: []
+  }) }}
+```
+
+If you are not using Nunjucks macros, update your HTML markup using the [card examples in the NHS digital service manual](https://service-manual.nhs.uk/design-system/components/card) as follows:
```
anandamaryon1: Hmm, we don't actually have examples in the design system of tables in panels. And we don't have any immediate plans to add them, the guidance would be tricky. 

Could perhaps link to the frontend review examples for HTML?

CHANGELOG.md
```diff
@@ -186,6 +196,85 @@ Please carefully review all custom components to make sure they scale dynamicall
 
 This change was introduced in [pull request #1655: Add support for Dynamic Type on Apple devices](https://github.com/nhsuk/nhsuk-frontend/pull/1655).
 
+#### Update the HTML for tables as a panel
+
+For consistency with other components, the HTML and Nunjucks options for tables as a panel have changed. The previous names are deprecated and will be removed in a future release.
+
+If you're using the `table` Nunjucks macro in your service, you should:
```
anandamaryon1:
```suggestion
If you're using the `table` Nunjucks macro for tables as a panel in your service, you should:
```

To make it more obvious that this isn't for all tables.

CHANGELOG.md
colinrotherham: See what you think of https://github.com/nhsuk/nhsuk-frontend/pull/1685/commits/a2cf24e3f3c5a563d253af5a1f3357d8f6a5ca0e

Thanks @anandamaryon1

CHANGELOG.md
colinrotherham: Had a go also in https://github.com/nhsuk/nhsuk-frontend/pull/1685/commits/a2cf24e3f3c5a563d253af5a1f3357d8f6a5ca0e

anandamaryon1 (approved):
Nicely done, I think this is ready now 👏

PR #1689, opened 2025-11-17
Only show header navigation items if not empty

frankieroberto:
This enables inline conditions to be used in the Nunjucks macro, eg:

```njk
{{ header({
  navigation: {
    items: [
      {
        text: "Home",
        href: "/"
      } if signedIn
    ]
  }
}) }}
```

Current if you do this, the macro outputs empty items which cause a visual gap to appear in the interface.

### Before

Screenshot 2025-11-17 at 16 42 42

### After

Screenshot 2025-11-17 at 16 42 54

### DIFF
--- a/packages/nhsuk-frontend/src/nhsuk/components/header/macro-options.mjs
+++ b/packages/nhsuk-frontend/src/nhsuk/components/header/macro-options.mjs
@@ -1440,7 +1440,7 @@ export const examples = {
       viewports: ['desktop']
     }
   },
-  'where some navigation items are false': {
+  'blue with navigation (empty items)': {
     context: {
       logo: {
         href: '#'


colinrotherham (commented):
Lovely stuff!

Changelog entry? Can copy https://github.com/nhsuk/nhsuk-frontend/pull/1653 probably

Any other `{% for item in …` loops need fixing?

frankieroberto:
@colinrotherham changelog item added. The `account.items` in the header already has an `if` condition present. I’ve not checked the rest of the macros though.

CHANGELOG.md
```diff
@@ -90,6 +90,7 @@ This change was introduced in pull requests [#1633: Review global `box-sizing` u
 - [#1653: Only show a task list item if not empty](https://github.com/nhsuk/nhsuk-frontend/pull/1653)
 - [#1656: Improve button text vertical alignment](https://github.com/nhsuk/nhsuk-frontend/pull/1656)
 - [#1657: Fix inconsistent margin in form and button groups](https://github.com/nhsuk/nhsuk-frontend/pull/1657)
+- [#1689: Allow header navigation items to be false](https://github.com/nhsuk/nhsuk-frontend/pull/1689)
```
colinrotherham: We used "if not empty" in https://github.com/nhsuk/nhsuk-frontend/pull/1653 because it's not just `false`

If that's alright?

```suggestion
- [#1689: Only show header navigation items if not empty](https://github.com/nhsuk/nhsuk-frontend/pull/1689)
```

packages/nhsuk-frontend/src/nhsuk/components/header/macro-options.mjs
```diff
@@ -1439,6 +1439,50 @@ export const examples = {
...
+          }
+        ]
+      }
+    },
+    options: {
+      width: false
+    },
+    screenshot: {
+      viewports: ['desktop']
+    }
```
colinrotherham: Don't necessarily worry about another screenshot

This would be one to cull in future

Shall we add it to **header.jsdom.test.mjs** instead?

frankieroberto: @colinrotherham I’ve removed the screenshot.

Feels like this ought to be covered by a `template.test.js` file rather than the javascript ones? We don’t have any of those yet though.

colinrotherham: We do all our Nunjucks and HTML accessibility testing in the JSDOM tests currently

But yeah absolutely, adding `template.test.js` would be even better

Not a blocker so we can do it another day

frankieroberto: I'll leave it for the backlog for now. Editing the existing `header.jsdom.test.mjs` is slightly beyond me (or at least it'd take me a good while to figure out).

packages/nhsuk-frontend/src/nhsuk/components/header/macro-options.mjs
```diff
@@ -1439,6 +1439,47 @@ export const examples = {
     screenshot: {
       viewports: ['desktop']
     }
+  },
+  'where some navigation items are false': {
```
colinrotherham: Sorry @frankieroberto can we rename this?

Just with the examples ordered A-Z with "blue" or "white" suffixes etc

* blue with navigation
* blue with navigation (justified)
* blue with navigation (overflow)
* blue with navigation (overflow, white)

Makes sense to make this:

```suggestion
  'blue with navigation (empty items)': {
```

PR #1697, opened 2025-11-20
Remove unused `topTask` card Nunjucks option

colinrotherham:
## Description

This PR removes the unused `topTask` card Nunjucks option

The top task card only uses `clickable: true` instead and no `--top-task` styles exist
https://service-manual.nhs.uk/design-system/components/card#top-task-card

## Checklist

- [x] CHANGELOG entry


colinrotherham:
>Worth mentioning very briefly in the changelog, or do we think it’s likely never been used in the wild (outside of the NHS.UK website)?

Happy to mention it as a reminder just in case they have their own styles (although no side effects from us)

colinrotherham:
@frankieroberto Changelog entry added in https://github.com/nhsuk/nhsuk-frontend/pull/1697/commits/377177541f29c3a5963363a212aaf7017f3f71e1

PR #1655, opened 2025-10-30
[v10] Add support for Dynamic Type on Apple devices

zoesalmon4-nhs:
## Description
This PR adds support for Dynamic Type on Apple devices for a11y purposes, allowing website text size to be influenced by the OS-level a11y settings.

> [**Supporting Dynamic Type**](https://developer.apple.com/design/human-interface-guidelines/typography#Supporting-Dynamic-Type)
>Dynamic Type is a system-level feature in iOS, iPadOS, tvOS, visionOS, and watchOS that lets people adjust the size of visible text on their device to ensure readability and comfort. For related guidance, see [Accessibility](https://developer.apple.com/design/human-interface-guidelines/accessibility).

These changes are optional and are only recommended where browser text size adjustments are not available. For example, in embedded HTML and Progressive Web Apps (PWAs).

## Implementation

When this feature is enabled, we use an `@supports` feature check for:

1. Dynamic Type `font: -apple-system-body` support
2. Touch screen `-webkit-touch-callout: default` support (iOS, iPadOS but not macOS)

If the above pass, we set the font to `-apple-system-body` to apply the Dynamic Type adjustment (as shown below), but then proceeding to set `font-family` so that `"Frutiger W01"` overrides the Apple system body font.

<br>

iOS Dynamic Type setting

## Feature flag

To enable this feature, set the feature flag variable `$nhsuk-include-dynamic-type` to `true` before you forward NHS.UK frontend in your Sass files:

```scss
@forward "nhsuk-frontend/dist/nhsuk" with (
  $nhsuk-include-dynamic-type: true
);
```

If you use the precompiled CSS from NHS.UK frontend, you can swap to an alternative stylesheet:

```patch
- <link rel="stylesheet" href="/stylesheets/nhsuk-frontend.min.css">
+ <link rel="stylesheet" href="/stylesheets/nhsuk-frontend-dynamic-type.min.css">
```

Please carefully review all custom components to make sure they scale dynamically when `$nhsuk-root-font-size` is set to Apple's ~16px~ 17px default font size on iOS.

For comparison, behaviour is switched between:

### Browser default

* 16px font size at `font-size: 100%` applies by default

```scss
$nhsuk-root-font-size: 16px;
```

```css
h1 {
  font-size: 3rem; // 48px
}
```

### Dynamic Type enabled

* 16px font size at `font-size: 100%` applies by default
* 17px font size at `font-size: 100%` applies on iOS and iPadOS only

[...]

PR #1698, opened 2025-11-21
Fix 2px minimum chevron `outline-width` syntax

colinrotherham:
## Description

Looks like the 2px minimum chevron `outline-width` isn't working

Should just be the single value? Noticed it when looking at https://github.com/nhsuk/nhsuk-frontend/pull/1655

If you could have a look please @paulrobertlloyd

### Safari

Safari web inspector showing invalid outline-width

### Chrome

Chrome web inspector showing invalid outline-width

## Checklist

- [x] CHANGELOG entry


colinrotherham:
Thanks @paulrobertlloyd

Noticed `border-width` was in the `@supports` query so I've updated that too

This might be a good technique to fix the button shadow and border drift (due to pixel snapping) though?

PR #1700, opened 2025-11-23
Select with search (aka autocomplete)

paulrobertlloyd:
## Description

Many services require users to choose from a large but limited number of options; too many to display as radio buttons, and yet selects are to be avoided as they can be difficult to use.

Autocompletes can help. These can display dropdown options in a more accessible way than the default HTML rendering, and allow options to be searched.

For example, in the *Give or refuse consent for vaccinations service*, if a parent wants to say that their child attends a different school, we provide a list of options in an autocomplete:

Searching for a school using a combobox powered buy the accessible-autocomplete component.

### Accessible autocomplete

This uses the GOV.UK [Accessible autocomplete](https://github.com/alphagov/accessible-autocomplete) component. However as of late-2025, this component:

- provides support for user agents no longer supported by NHS.UK frontend
- supports a range of different interactions and use cases, leading to complexity and large file size
- uses CSS that fits with GOV.UK styles but not those of NHS.UK
- can be difficult to get working with the prototype kit (and in production)

Finally, it is not actively maintained by GDS. A replacement component is on their roadmap under [Future plans](https://design-system.service.gov.uk/community/roadmap/#future-plans), but it doesn’t appear to be an immediate priority

That being said, it is exceedingly well battle tested both in the lab and in the wild, and remains the gold-standard for such a component.

### Opportunity

Right now, 2 of the Accessible autocomplete’s original authors (@edwardhorsford and @tvararu) are working within the NHS, and a lot of talented people are actively working on the NHS design system.

While acknowledging that building an autocomplete component is not for the faint-hearted, now seems like a good opportunity to add something akin to Accessible autocomplete as a first-party component to the NHS.UK design system.

### Standing on the shoulders of giants

I really liked @frankieroberto earlier proposals for GOV.UK Frontend. Rather than attempt to create an uber-autocomplete, he instead addressed specific use cases via targeted enhancements to 2 existing components:

- https://github.com/alphagov/govuk-frontend/pull/2453
- https://github.com/alphagov/govuk-frontend/pull/2430

This PR implements the second of these, progressively enhancing a select element so that a user can search the many options it contains.

To achieve this, I started with @adamsilver’s autocomplete control ([blog post](https://adamsilver.io/blog/building-an-accessible-autocomplete-control/), [code](https://github.com/adamsilver/nostyle/tree/master/src/components/autocomplete), [example](https://nostyle.onrender.com/examples/autocomplete)) which he developed over 5 years ago, consulting with Theo.

I removed the jQuery dependency, before then slowly reworking the component to follow naming and coding patterns we use for our components. Here is an example showing hint text and a focused option:

<p>Select showing search options, with one of the options focused.</p>

➡️ [View example of an enhanced select](https://nhsuk-frontend-pr-1700.herokuapp.com/nhsuk-frontend/components/select/default/)

➡️ [View example of an enhanced select with search](https://nhsuk-frontend-pr-1700.herokuapp.com/nhsuk-frontend/components/select/with-search/)

#### Persistent live region

Following the general principle that if something is worth showing to one set of users, it probably makes sense to show it to all, I’ve added persistent, visible assistive hint text below the component.

This live region updates to show either the number matching options found, else falls back to a description on how to use the component (with total number of options available shown). All these messages can be localised.

Needless to say, this needs a lot more work and consideration. Perhaps a similar pattern to that used for the Character Count can be used, where there are both visible and hidden messages, updated at different rates? @tvararu also told me that there can be issues when typing a letter can lead to the same number of results being returned – the fix included in accessible autocomplete is not in this implementation (yet).

#### Indicating when a partial search is invalid

As this is a select component, one of the options must be selected. Therefore, if a user types in a search, and then exits the component before selecting a valid option, the component is shown with a red outline.

Or should this case only be handled via server side validation, and existing error message and summary patterns? 

select-with-invalid-search

(The status message should be updated for this scenario so as to not rely on colour alone, and that’s currently missing, if we continue with this client-side approach.)

#### Hint text

Although possible with Accessible autocomplete via its templating feature, this component supports showing hint text alongside values as a first-party option. Currently this text is not included when searching options – should it?

#### Synonyms

Borrowing from Frankie’s proposal (though not as extensive as [Ed’s extensions to the autocomplete](https://becoming-a-teacher.design-history.education.gov.uk/register-trainee-teachers/autocomplete-improvements/)), I’ve also added support for synonyms.

These are values that are not shown to the user, but can help them find an option when filtering results in the combobox for example, a previous name, official acronym or common mis-spelling.

#### Design

Finally, I’ve made the corners on the right-hand side rounded, and those on the left-hand side square, to reflect the dual-nature of this component. If the size of `$nhsuk-border-radius` is increased via #1600, this will effect will become more pronounced, which I think is a good thing.

## Next steps

This PR is presented as a first step and a foundation to build upon.

It would be good to reach early consensus on the shape and scope of such a component, before digging into the many bugs, issues and complex interactions that’d need testing and unpicking. By having the overall scope of this component be more contained, hopefully it shouldn’t be as difficult as that for the Accessible autocomplete.

## Checklist

- [ ] CHANGELOG entry


frankieroberto:
@paulrobertlloyd ooh, this is a big project to (re-)start, but maybe by keeping it tightly scoped and with the people we have here we can get it across the line! ❤️

One design idea I explored way back in https://github.com/alphagov/govuk-frontend/pull/2430 was to make it work a bit more like regular Selects by having a ✔️ to show the currently selected item. It does have the downside of indenting all the text to make space, but it could help to make things clearer the component is used on an edit screen.

I’ve also since had the thought that the component could also have an option where the search feature is not enabled, and it effectively becomes just a much more accessible Select component? Then it could be a replacement for any regular dropdowns, in the same way as the [enhanced File upload](https://design-system.service.gov.uk/components/file-upload/#using-the-improved-file-upload-component)?

colinrotherham:
Can you try rebasing this with `main` please @paulrobertlloyd?

Very odd visual regression failure showing `hashFiles('package-lock.json')` failed

paulrobertlloyd:
> I’ve also since had the thought that the component could also have an option where the search feature is not enabled, and it effectively becomes just a much more accessible Select component?

@frankieroberto I had the very same idea! I actually tried to get that working by adding `read-only` attribute to the input, but that caused focus issues. Thinking about it afresh today, perhaps that was the wrong approach. Instead, when search isn’t enabled rather than add an `<input>` we instead add a `<button>`. I’ll have a play with that idea.

frankieroberto:
> As this is a select component, one of the options must be selected. Therefore, if a user types in a search, and then exits the component before selecting a valid option, the component is shown with a red outline.

@paulrobertlloyd I think (but I'll have to try and get it running again to be sure) that when I last looked at this I tried setting the behaviour so that if you search for something but don't select an option (eg because there are no matches), then when you blur the field it reverts back to what is currently actually selected (which could be blank if that was the default value). This felt a bit more conceptually closer to how a Select works, in that the field can't be left in an invalid state, and searching is a temporary mode to help you find the right option to select? (Native selects have this too, but just don't show you what you've typed in, and just jump to the first match rather than filtering the options)

Not without its challenges though!

paulrobertlloyd:
I’ve done quite a bit of a rewrite, with the following changes:

- All selects are now enhanced. Due the nature of the menu, this means selects behave like inputs in that they are 100% width by default, and need a width applied (either directly or to their container) to constrain its size
- The selected option is now marked with a checkmark
- If you exit the component having entered an invalid value, the search input (and therefore the underlying value) will revert to the previous valid value
- The new ‘divider’ option is supported by the enhanced select design, too
- Rebased with `main` 

Spotted a few weird behaviours, but the bones of a component that can wholesale replace the native select is now here.

Select without search and showing a divider between items.

frankieroberto:
@paulrobertlloyd looking good!

The trickiest bit is how best to try and signal to users that they can search rather than just scroll through the list and find the option that way. The half-squared, half-rounded corners is probably too subtle for that (although I kinda like it). When you click the box you get a caret with the search option, but that's also quite easily missed.

Which kinda leaves us relying on text. I can see why you've added "Select an option or search" to the default status text, but perhaps we should recommend that this be included as hint text above the search box instead, to make it more visible?

It's particularly an issue on mobile, as the status text disappear below the on-screen keyboard (although at least the keyboard is more of a hint that you can search):

A8B63BF2-D4EF-454A-9550-91036D572DD2_1_102_o

frankieroberto:
One bug I spotted: if you select an option in the combo box, and then search again (eg removing the last character), the checked icon disappear from the currently selected item. I think it should remain in place until you either select a different option?

paulrobertlloyd:
@frankieroberto 

- Now persist selected (checked) option until you select a new option
- Showing hint text above the combobox input
- Use new larger border radius for this component (until #1600 gets merged)

There’s a bug where the input gets reset to the previous value when you delete all the text… tried a fix but that caused other issues, so need to dig into this behaviour a bit more.

anandamaryon1:
Sorry I've not chimed in until now, this is brilliant! I really like the idea of narrowing the scope to split it up. 

Design wise, I wonder how we could better connect the dropdown box with the input, perhaps just making the border-bottom on the input thinner when it's open and lining up the right-hand borders? 

Not sure on how best to communicate the dual nature of search and select, it's a challenge. The different borders makes some sense but it's very subtle and a bit odd looking. Search inputs often have placeholder text, perhaps that could be used? Or, if we're using these because we expect the list to be too long to scroll, perhaps we don't show the chevron and it looks more like a text input and therefore users expect to type into it? Presents issues if we then add the autocomplete with suggestions though…

On enhancing the select menu by default. I worry about mobile, on mobile OS the select menus are quite different and I expect users are more used to them, they have nice big hit areas etc. Perhaps on mobile it can defer to native? Or, if using the autocomplete, it can go full screen to allow space for just searching and selecting? 

Thanks for this, it's a huge step forward and I'm keen to get it published. I'm going to discuss this at the next design system review panel next week, I'll come back with any further thoughts.

anandamaryon1:
Some thoughts from discussion with the Design System Review Panel:

* Mix of border-radius on one side and not on the other looks very strange
* Worth mapping ‘esc’ key to close the drop-down
* In guidance, need to clearly delineate between dropdown with search and search with autocomplete / autosuggest
* Appears to search only main text, not hint text, eg. Fibula
* Scroll bars look very narrow, which can be an accessibility issue, can we make them wider?
* Searching “di” on medical condition example shows 4 results but only two contain that pair of letters. – Think this is the synonyms working. Is there a way to communicate this to a user? 
* Should search match letters in the middle of a word?

frankieroberto:
> Some thoughts from discussion with the Design System Review Panel:

🙌 

> * Mix of border-radius on one side and not on the other looks very strange

I think the idea here was to offer an extra clue (as well as the "Select an option or search" hint text) that you can search. Otherwise it looks identical to regular dropdowns and users might spend ages scrolling through the list not realising they can search.

Maybe it’s too subtle though? Other options:

* both edges could be squared (if searchable) vs rounded (non-searchable)
* keep it rounded and rely on the hint text + caret appearing when focusing the box
* something else?

> * Worth mapping ‘esc’ key to close the drop-down

This is already working for me? Although I think there's a bug: as well as closing the dropdown, pressing Esc should revert the displayed value to the currently selected item.

> * In guidance, need to clearly delineate between dropdown with search and search with autocomplete / autosuggest

Yeah, that's tricky! Especially as we don’t (yet) have a search with autocomplete component. I think in guidance we can differentiate them as closed question vs open question?

Thinking ahead for the search with autocomplete component design, my hunch is that it would be different by:

* not having the down-arrow on the right
* having a vertical gap between the input box and the suggestions
* including the word 'Suggestions' above the suggestions

> * Appears to search only main text, not hint text, eg. Fibula

Good point. Should probably search both?

> * Scroll bars look very narrow, which can be an accessibility issue, can we make them wider?

👍 

> * Searching “di” on medical condition example shows 4 results but only two contain that pair of letters. – Think this is the synonyms working. Is there a way to communicate this to a user?

Could display synonym in brackets if the text matches the synonym but not the main text? That’s what the GOV.UK design system search box does:

Screenshot 2026-03-11 at 15 00 33

On the other hand, if the synonym you’ve added is a common typo or mis-spelling of the word (eg "Astma" for "Asthma") then perhaps seeing that in brackets would be weird?

> * Should search match letters in the middle of a word?

Ah I think this is a bug - should only match the start letters (of any word)?

anandamaryon1:
I've been looking at our search input in response to [some recent feedback](https://github.com/nhsuk/nhsuk-service-manual/issues/2631) and it's led me to thinking about select with search, search inputs and related variants. Based on this I've been looking again at this PR and have some thoughts on design.

I found at this GOV.UK publishing component: [select with search](https://components.publishing.service.gov.uk/component-guide/select_with_search.). Which behaviour matches more closely to a regular select, forcing a default selection and then enhancing the dropdown with a search. This feels like a good balance to me. I think it helps to have the search separate from the select input, to signal that it's only filtering the available options and to differentiate the search term from the input value. 

Below I've explored this (2. & 3.) alongside adding a search icon (1.) to a design more in line with current the PR. 

image

I'm leaning toward (2.) which will allow us to use our search (icon/no-button) input rather than more custom search styles. And, the non-progressively enhanced select will look the same as the enhanced (before clicking).

Keen for other's thoughts. 

And finally, a quick update on where this sits in our backlog. We're currently focussed on:
* Section nav component and v11 (tweaking)
* Product pages pattern (UR and tweaking)
* Common error page patterns 

And then:
* Updating Expanders and Details in line with gov accordions
* Accessible autocomplete (this!)
* Address lookup pattern (gone quiet at the mo)

frankieroberto:
@anandamaryon1 interesting! 

I discovered today that the 'intelligent navigation' team in the NHS App have been looking at a similar autocomplete/multi-select component to let users search for and select multiple symptoms. Will encourage them to contribute!

PR #1702, opened 2025-11-24
[v10] Expose NHS.UK frontend Node.js render API

colinrotherham:
## Description

This PR adds a new package entry point `nhsuk-frontend/lib` to render components without Nunjucks code

For example, using `components.render()` in Node.js to output HTML for the input component:

```mjs
import { components } from 'nhsuk-frontend/lib/index.mjs'

const html = components.render('input', {
  context: {
    label: {
      text: 'Mobile phone number'
    },
    name: 'mobile',
    type: 'tel'
  }
})
```

This feature is intended as an internal API to render components for the Design System, but it also helps with [**Test if your HTML matches NHS.UK frontend**](https://github.com/nhsuk/nhsuk-frontend/blob/main/docs/contributing/testing-your-html.md) where the hard-coded per-component `fixtures.json` may not be enough

## Checklist

- [ ] CHANGELOG entry

### DIFF
--- a/packages/nhsuk-frontend/package.json
+++ b/packages/nhsuk-frontend/package.json
@@ -45,6 +45,7 @@
   "files": [
     "dist",
     "src",
+    "!**/fixtures",
     "!**/*.test.*"
   ],
   "scripts": {
--- a/packages/nhsuk-frontend/src/nhsuk/lib/components.unit.test.mjs
+++ b/packages/nhsuk-frontend/src/nhsuk/lib/components.unit.test.mjs
@@ -1,0 +1,74 @@
+import { join } from 'node:path'
+
+import { render } from './components.mjs'
+import { configure } from './nunjucks.mjs'
+
+describe('Node.js API: Components', () => {
+  const context = {
+    label: {
+      text: 'Mobile phone number'
+    },
+    name: 'mobile',
+    type: 'tel'
+  }
+
+  const output = `
+    "<div class="nhsuk-form-group">
+      <label class="nhsuk-label" for="mobile">
+        Mobile phone number
+      </label>
+      <input class="nhsuk-input" id="mobile" name="mobile" type="tel">
+    </div>
+    "
+  `
+
+  const outputCustom = `
+    "<p class="app-example">Custom example</p>
+    "
+  `
+
+  describe('render', () => {
+    it('returns component HTML', () => {
+      const html = render('input', { context })
+      expect(html).toMatchInlineSnapshot(output)
+    })
+
+    it('returns component HTML (custom env)', () => {
+      const html = render('input', { context, env: configure() })
+      expect(html).toMatchInlineSnapshot(output)
+    })
+
+    it('returns custom component HTML (custom env)', () => {
+      const viewsPath = join(import.meta.dirname, 'fixtures/components')
+
+      const html = render('example1', {
+        context: { text: 'Custom example' },
+        env: configure(viewsPath)
+      })
+
+      expect(html).toMatchInlineSnapshot(outputCustom)
+    })
+
+    it("returns custom component HTML with 'app' prefix (custom env)", () => {
+      const viewsPath = join(import.meta.dirname, 'fixtures/components')
+
+      const html = render('example2', {
+        context: { text: 'Custom example' },
+        env: configure(viewsPath),
+        prefix: 'app'
+      })
+
+      expect(html).toMatchInlineSnapshot(outputCustom)
+    })
+
+    it('throws with missing component macro', () => {
+      const renderHtml = () => render('example', { context })
+
+      expect(() => renderHtml()).toThrowErrorMatchingInlineSnapshot(`
+        "(unknown path)
+          Template render error: (unknown path)
+          Error: template not found: example/macro.njk"
+      `)
+    })
+  })
+})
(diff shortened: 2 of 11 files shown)


packages/nhsuk-frontend/src/nhsuk/lib/nunjucks.mjs
```diff
@@ -1,25 +1,24 @@
...
+import { resolve } from 'node:path'
+import { fileURLToPath } from 'node:url'
 
-import { environment, paths } from '@nhsuk/frontend-config'
 import nunjucks from 'nunjucks'
 import { outdent } from 'outdent'
 
+// Node.js environment with default
+// https://nodejs.org/en/learn/getting-started/nodejs-the-difference-between-development-and-production
+export const { NODE_ENV = 'development' } = process.env
```
MatMoore: why the default to development if the link above says to use production everywhere?

packages/nhsuk-frontend/src/nhsuk/lib/README.md
```diff
@@ -0,0 +1,19 @@
+# NHS.UK frontend Node.js API
+
+Import the NHS.UK frontend Node.js API to render components without Nunjucks views.
```
MatMoore, line 3: Should this specify that it is using nunjucks under the hood, and the params are those of the nunjucks components?

There is also an asymmetry here: if I define a custom component in my app, and use the base class provided by nhsuk-frontend, I might expect to be able to call this function, like I can for any of the packaged components. 

Maybe that's fine, but I wonder if there should be a way to configure the component such that this render function still works.

packages/nhsuk-frontend/src/nhsuk/template.njk
```diff
@@ -7,10 +7,10 @@
 {% set themeColor = themeColor | default("#005eb8", true) -%}
 
 <!DOCTYPE html>
-  <html lang="{{ htmlLang | default("en", true) }}" {{- nhsukAttributes(htmlAttributes) }}>
+  <html lang="{{ htmlLang | default("en", true) }}" {%- if htmlClasses %} class="{{ htmlClasses }}"{% endif %} {{- nhsukAttributes(htmlAttributes) }}>
```
MatMoore: is this commit required for the rest of the PR?

packages/nhsuk-frontend/src/nhsuk/lib/nunjucks.mjs
colinrotherham: One for another PR? The comment has been moved (it's not new)

That said, the guidance says the absence of `NODE_ENV` would (to some libraries) default to `development`, but the stricter heading "[**Why is NODE_ENV considered an antipattern?**](https://nodejs.org/en/learn/getting-started/nodejs-the-difference-between-development-and-production#why-is-node_env-considered-an-antipattern)" [was not present in previous versions](https://web.archive.org/web/20240714052821/https://nodejs.org/en/learn/getting-started/nodejs-the-difference-between-development-and-production).

Our library only uses `NODE_ENV=production` and `NODE_ENV=test`

We should follow the refreshed guidance and drop the default value

```suggestion
export const { NODE_ENV } = process.env
```

packages/nhsuk-frontend/src/nhsuk/template.njk
colinrotherham: No, good point—thank you! I found these omissions by working on this PR but they need splitting out

colinrotherham:
>This looks reasonable, I'm just a bit confused about the intended use case. Is it mainly going to be for prototype kit users? other node projects?

It's intended use is as an internal API to render components for the Design System, but it also helps with [**Test if your HTML matches NHS.UK frontend**](https://github.com/nhsuk/nhsuk-frontend/blob/main/docs/contributing/testing-your-html.md) where the hard-coded per-component `fixtures.json` may not be enough

packages/nhsuk-frontend/src/nhsuk/lib/README.md
colinrotherham: You can pass in the `env` option as a custom Nunjucks environment with different load paths

It's not clearly documented but I'm not sure how much I should add [as it's primarily for internal use](https://github.com/nhsuk/nhsuk-frontend/pull/1702#issuecomment-3834810331)

packages/nhsuk-frontend/src/nhsuk/lib/nunjucks.mjs
colinrotherham: Resolved in https://github.com/nhsuk/nhsuk-frontend/pull/1702/commits/83ee7ec0f13c0a58a8eb7d38acdca09f3e5a3740

packages/nhsuk-frontend/src/nhsuk/template.njk
colinrotherham: Split out into https://github.com/nhsuk/nhsuk-frontend/pull/1795

packages/nhsuk-frontend/package.json
```diff
@@ -34,9 +44,8 @@
   "types": "dist/nhsuk/index.d.mts",
   "files": [
     "dist",
+    "lib",
```
colinrotherham: Note to self, need to remove this (moved it to `src/nhsuk/lib`)

PR #1705, opened 2025-11-25
Add dependabot cooldown

frankieroberto:
See https://blog.yossarian.net/2025/11/21/We-should-all-be-using-dependency-cooldowns

It's possibly less needed here as Dependabot is on a monthly update schedule, but I guess we could be unlucky and it includes a just-released malicious update?

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -77,7 +77,7 @@ updates:
     schedule:
       interval: monthly
     cooldown:
-      default-days: 4
+      default-days: 7
 
     versioning-strategy: increase
 
@@ -98,3 +98,5 @@ updates:
 
     schedule:
       interval: 'monthly'
+    cooldown:
+      default-days: 7


.github/dependabot.yml
```diff
@@ -76,6 +76,8 @@ updates:
 
     schedule:
       interval: monthly
+    cooldown:
+      default-days: 4
```
colinrotherham: What's the thinking behind 4 days?

frankieroberto: I just copied that from another team. 🤷  I guess it's based on 4 days being longer than all the "windows of opportunity" listed in https://blog.yossarian.net/2025/11/21/We-should-all-be-using-dependency-cooldowns ?

colinrotherham: Our Jinja friends have gone with 7 days but included GitHub Actions too

Shall we do the same? https://github.com/NHSDigital/nhsuk-frontend-jinja/pull/76

frankieroberto: Sure. I assumed GitHub actions are safer as we’re only using the ones authored by GitHub. But I guess a cooldown is probably fine there too?

frankieroberto: Updated.

PR #1707, opened 2025-11-26
[v10] Add checkbox "all" option

frankieroberto:
This updates the Checkbox component with a new "all" option using using `inclusive: true`.

The new "all" checkbox will select all the others when checked.

The existing "none" option deselects all the others when checked.


frankieroberto:
I've found myself using this in a prototype this week, for a checkbox listing pharmacies (where there can be a lot of them):

Screenshot 2025-12-09 at 12 39 51

We’re aiming to user test this in the coming weeks.

frankieroberto:
@colinrotherham I’ve updated this (with some help from AI) so I think it’s a reviewable state now.

I made one behavioural change which is that the “select all” checkbox becomes automatically checked if you individually check all the other checkboxes.

frankieroberto:
One difference with the GOVUK implementation of this (which I originally wrote) is that that uses `behaviour: "exclusive"` whereas the NHS one uses `exclusive: true`.

The advantage of the GOVUK one is that you can’t specify both inclusive and exclusive on same input! (Not that they support inclusive behaviour yet)

We could either choose to align with the GOVUK version and switch to a `behaviour` param (which could be backwards-compatible), or keep with `exclusive: true` and `inclusive: true` but make a decision about which one takes priority if both are specified?

frankieroberto:
@anandamaryon1 @colinrotherham @sarawilcox any chance we could review this for the next release? I think we have enough evidence that it would be useful?! (It's been designed into at least 2 vaccination services).

anandamaryon1:
Sorry for not replying on this one. It makes total sense. Did you get to run your user testing with it? 

I think we'd want to add guidance and an example in the design system too. Any insight as to why gov haven't added it?

colinrotherham:
Related issues etc

* https://github.com/alphagov/govuk-design-system-backlog/issues/37#issuecomment-429334420
* https://github.com/alphagov/govuk-design-system-backlog/issues/292
* https://github.com/alphagov/govuk-frontend/issues/4653

frankieroberto:
I've asked in the cross-gov Slack about it too.

frankieroberto:
We tested the interface in https://github.com/nhsuk/nhsuk-frontend/pull/1707#issuecomment-3632066144 in our last round of user research, with staff who use Record a vaccination across multiple pharmacies. We didn’t find any issues with the interaction.

In our current prototype I’m always including the select-all option so long as there’s 2+ options (if there’s only a single option they skip the screen entirely).

It might be that it only makes sense to include the select-all option if there are 5+ or some other number of options though.

colinrotherham:
I'd like us to [support `item.behaviour` to set `data-behaviour="{{ item.behaviour }}"` like GOV.UK Frontend](https://github.com/nhsuk/nhsuk-frontend/pull/1707#issuecomment-3746973873)

We'd want backwards compatibility with our NHS specific `item.exclusive` option in v10.x too

But what about multiple _inclusive_ groups—would we support these too?

## Multiple named groups

Unlike GOV.UK Frontend we also support multiple named groups via `item.exclusiveGroup`

E.g. Try the 3 named exclusive groups:
https://nhsuk.github.io/nhsuk-frontend/examples/conditional-reveals/

* None of the primary colours
* None of the secondary colours
* None of the above

Should we rename to ~`item.exclusiveGroup`~ `item.behaviourGroup` to pair with `item.behaviour`?

frankieroberto:
@colinrotherham maybe initially we add this in a backwards-compatible way, by adding `inclusive: true` and `inclusiveGroup` (see the PR as it is now)?

Then could consider switching to `behaviour: "inclusive"` and `behaviour: "exclusive"` in the next breaking change release, along with either `behaviourGroup` or removing the group functionality (depending on whether we still think there’s a good use case for it)?

colinrotherham:
@frankieroberto I've had a go at [deduplicating the spike code](https://github.com/nhsuk/nhsuk-frontend/pull/1707/commits/92c862058191033a7b6e97199139884a75ee06a6) I added back in https://github.com/nhsuk/nhsuk-frontend/pull/1707/changes/70ee9ba95cec370e41a5da5ea11408e5076ccfd2

>Then could consider switching to behaviour: "inclusive" and behaviour: "exclusive" in the next breaking change release, along with either behaviourGroup or removing the group functionality (depending on whether we still think there’s a good use case for it)?

Rather than add new options, only to remove them later, I've pushed up changes to:

* Add Nunjucks options `behaviour` and `behaviourGroup`
* Deprecate Nunjucks options `exclusive: true` and `exclusiveGroup`

This gives us both `behaviour: "inclusive"` and `behaviour: "exclusive"` without breaking changes

colinrotherham:
@frankieroberto Also to discuss, whether we should hide "Select all" (or even "None of the above") without JS?

See commits https://github.com/nhsuk/nhsuk-frontend/pull/1707/commits/895e722cd9481ba8b45c449277852a10b70f549d and https://github.com/nhsuk/nhsuk-frontend/pull/1707/commits/77d95e4997ae359619bbed48d16f66c624b36ab5 – thoughts?

sarawilcox:
Hi @frankieroberto, do we need something adding to the guidance in the design system about this?

frankieroberto:
@colinrotherham I think it makes sense to not show the 'select all' checkbox where javascript is unavailable, and treat it as progressive enhancement? Otherwise, we'd have to infer that if someone ticked 'select all' but none (or only some) of the other checkboxes, which feels slightly risky?

This is different from the 'none' checkbox, as that still has a benefit with no javascript - and we have some guidance suggesting showing an error if they selected both another checkbox _and_ 'none':

> If JavaScript is unavailable, and a user selects both the "none" checkbox and another checkbox, display an error message.

colinrotherham:
Thanks @frankieroberto, what did you think about the implementation?

> See commits https://github.com/nhsuk/nhsuk-frontend/commit/895e722cd9481ba8b45c449277852a10b70f549d and https://github.com/nhsuk/nhsuk-frontend/commit/77d95e4997ae359619bbed48d16f66c624b36ab5 – thoughts?

frankieroberto:
@colinrotherham I'm happy with the implementation. Adding `behaviour: 'inclusive'` to the divider to also hide that when javascript is disabled makes sense to me, and I can’t think of any other clean way to infer when dividers should be hidden, without a ton of extra logic.

colinrotherham:
@frankieroberto Brilliant. I'll need to rebase with `support/10.x` instead of `main`

Will pick it up when I get chance, thank you 🙏

frankieroberto:
@colinrotherham this is currently a breaking change, so should target `main` instead? 

It's breaking as it removes the existing `exclusive: true`, renaming it to `behaviour: "exclusive"`.

We could make it backwars compatible, and support both `exclusive: true` AND `behaviour: "exclusive"` for "none" options, but perhaps only `behaviour: "inclusive"` for the "all" option? But then that does pose the question of what happens if you use `exclusive: true` and `behaviour: "inclusive"` on the same item... 😬 

If v11 is not far away, it might be simpler to wait until then? But if it's a way off, we could put the effort in to make it backwards-compatible to enable a sooner release?

colinrotherham:
Wasn't it backwards compatible until https://github.com/nhsuk/nhsuk-frontend/pull/1707/changes/0709320caf14b872132a210723cfe8178fe01ddf when `exclusive` and `exclusiveGroup` were removed?

Only because this PR has the v10.6.0 milestone

frankieroberto:
@colinrotherham at the moment this PR drops support for `exclusive: true`, by the looks of it, so that's not backwards-compatible, but the JavaScript is: https://github.com/nhsuk/nhsuk-frontend/pull/1707/changes#diff-f3029a8bee8b96cb9d973cff3a8a09ae04dd4b7059d231834c1f00487a773654R312

frankieroberto:
ok I think we've decided to make this a non-breaking change now, so the older params are still supported but deprecated.

colinrotherham:
@frankieroberto All green now and ready for v10.6.0

I'll do some more testing of this before we merge it in

Thanks 🙏

frankieroberto:
@colinrotherham thank you! I’ll draft some guidance.

frankieroberto:
Added draft guidance in https://github.com/nhsuk/nhsuk-service-manual/pull/2583

colinrotherham:
I've had a little go at updating the examples in the changelog entry now the guidance is coming along

Review app is enabled again too

colinrotherham:
@frankieroberto We previously added `nhsuk-frontend-supported-only` to this PR

It's now been renamed to `nhsuk-u-hidden-not-supported` and split out into https://github.com/nhsuk/nhsuk-frontend/pull/1972

colinrotherham:
I've renamed our example names to use "none" and "all" (to match the guidance) in https://github.com/nhsuk/nhsuk-frontend/pull/1707/commits/d4d0a1a2d8cddcce518a8e54c46eb3b020d6f177

@frankieroberto Shall we change the colours example to a vaccines one?

colinrotherham (approved):
Brilliant work! 🙌

Just that question about maybe a more NHS-specific example @frankieroberto

colinrotherham:
Also flagging that we don't have Puppeteer tests for "all" if you think we should

colinrotherham (approved):

packages/nhsuk-frontend/src/nhsuk/components/checkboxes/checkboxes.jsdom.test.mjs
```diff
@@ -269,7 +269,7 @@ describe('Checkboxes', () => {
 
       expect($inputBehaviour).not.toBeChecked()
 
-      // Check "None of the above"
+      // Check "None"
```
colinrotherham: Should this be `Check "I do not want to be contacted"`?

Otherwise maybe `Check "none" option` to keep it generic

packages/nhsuk-frontend/src/nhsuk/components/checkboxes/checkboxes.mjs
```diff
@@ -104,7 +104,7 @@ export class Checkboxes extends ConfigurableComponent {
    * Uncheck other checkboxes
    *
    * Find any other checkbox inputs with the checkbox group value, and uncheck them.
-   * This is useful when a “None of the above" checkbox is checked.
+   * This is useful when a “None of these" checkbox is checked.
```
colinrotherham: I think we used to say "None of these" before I changed it to "None of the above"

Shall we refer to the "none" option here too?

packages/nhsuk-frontend/src/nhsuk/components/checkboxes/checkboxes.jsdom.test.mjs
frankieroberto: generic sounds good to me. Changed in 930436c465d83b8553c36caacb98755351437065

packages/nhsuk-frontend/src/nhsuk/components/checkboxes/checkboxes.mjs
frankieroberto: I think "None of these" is ok here? 🤷  Maybe it could be:

```suggestion
   * This is useful when a "None of these" type of checkbox is checked.
```

to make it clear that you don’t have to use that exact wording?

frankieroberto: Or this?

```suggestion
   * This is useful when a "none" checkbox option is checked.
```

colinrotherham: Last one but lowercase "none" like in the guidance?

frankieroberto: done

packages/nhsuk-frontend/src/nhsuk/components/checkboxes/checkboxes.mjs
```diff
@@ -104,80 +104,180 @@ export class Checkboxes extends ConfigurableComponent {
...
 
-    const selectorGroup = exclusiveGroup
-      ? `[data-checkbox-exclusive-group="${exclusiveGroup}"]`
-      : `[name="${$input.name}"]`
+  /**
+   * Uncheck inputs
+   *
+   * Find any checkbox inputs with the same checkbox group value and behaviour
+   * and uncheck them. This helps prevent checking both a regular checkbox and
+   * "None of these" at the same time
```
colinrotherham: Got one more "None of these here" 🙈

packages/nhsuk-frontend/src/nhsuk/components/checkboxes/checkboxes.mjs
```diff
@@ -104,80 +104,180 @@ export class Checkboxes extends ConfigurableComponent {
...
+    }
+  }
 
-    allInputsWithSameName.forEach(($inputWithSameName) => {
-      const hasSameFormOwner = $input.form === $inputWithSameName.form
+  /**
+   * Check other checkboxes
+   *
+   * Find any other checkbox inputs with the checkbox group value, and check them.
+   * This is useful for when a "Select all" checkbox is checked.
```
colinrotherham:
```suggestion
   * This is useful for when an "all" checkbox option is checked.
```

packages/nhsuk-frontend/src/nhsuk/components/checkboxes/checkboxes.mjs
```diff
@@ -104,80 +104,180 @@ export class Checkboxes extends ConfigurableComponent {
...
+    for (const $input of $groupInputs) {
+      this.setInputState($input, true, behaviourGroup)
+    }
+  }
+
+  /**
+   * Check inputs
+   *
+   * Find any checkbox inputs with the same checkbox group value and behaviour
+   * and check them. This helps prevent "Select all" being left unchecked when
```
colinrotherham:
```suggestion
   * and check them. This helps prevent "all" options being left unchecked when
```

frankieroberto:
@colinrotherham @anandamaryon1 all issues resolved. Ok to merge this?

CHANGELOG.md
```diff
@@ -3993,7 +4094,7 @@ After:
 
 ### :new: **New features**
 
-- Add "None of these" JavaScript behaviour to checkboxes - allowing all checkboxes in a group to be automatically unchecked when the "None of the above" option is checked ([PR 756](https://github.com/nhsuk/nhsuk-frontend/pull/756))
+- Add "None of these" JavaScript behaviour to checkboxes - allowing all checkboxes in a group to be automatically unchecked when the "None of these" option is checked ([PR 756](https://github.com/nhsuk/nhsuk-frontend/pull/756))
```
colinrotherham: We should leave old changelog entries as they were, I'll edit this

CHANGELOG.md
```diff
@@ -1812,9 +1913,9 @@ The file structure for the stylesheets has changed. You will have to make differ
 
 For precompiled CSS files, note the following path changes:
 
-- copy or serve `node_modules/nhsuk-frontend/dist/nhsuk/nhsuk-frontend.min.css`  
+- copy or serve `node_modules/nhsuk-frontend/dist/nhsuk/nhsuk-frontend.min.css`
```
colinrotherham: We've lost the double spaces (to preserve line breaks) at the end of a few of these lines

I'll sort it

colinrotherham:
@frankieroberto I've updated the macro option descriptions to use "all" and "none" in https://github.com/nhsuk/nhsuk-frontend/pull/1707/commits/2fb2291408cd877489adbc77b91da9552da0c3ce  and added a test for the deprecated method in https://github.com/nhsuk/nhsuk-frontend/pull/1707/commits/416836be0e7e715706df3530716045acd1260eb4. Think we're good to go now

PR #1670, opened 2025-11-05
[v10] Add label and legend `caption`, `headingLevel` and other options

colinrotherham:
# Description

This PR addresses two conflicting decisions:

* Headings can have multiple heading levels (h1, h2, h3 etc) with optional captions
* Labels and legends can optionally be headings (h1 only) but never a caption

This causes problems when [an `<aside>` with a search input](https://nhsuk-frontend-pr-1670.herokuapp.com/nhsuk-frontend/examples/small-form-controls/) needs an associated heading:

Screenshot 2026-08-19 at 12 11 02

<br>
<br>

See related PR and issue over on GOV.UK Frontend

* https://github.com/alphagov/govuk-frontend/pull/5240
* https://github.com/alphagov/govuk-frontend/issues/6471

## Nunjucks heading levels

We've decided to deprecate the `isPageHeading` option and replace it with `heading` and `headingLevel`

For three main reasons:

* It does not support heading levels other than `<h1>`
* It does not follow our other boolean flags, e.g. we use `card: true` not `isCard: true`
* We already use `heading.level` in other components' nested heading

Similar to the `size: "m"` option, we can choose to restrict the allowed heading levels values in future.

From the PR changelog entry:

#### Rename the label and legend `isPageHeading` option

We've deprecated the `isPageHeading` Nunjucks option for labels and legends and added the `heading` option.

If you're using any form component Nunjucks macro with the `isPageHeading` option, you should:

- rename `text` to `heading`
- remove `isPageHeading: true`

```patch
  {{ characterCount({
    label: {
-     text: "Can you provide more detail?",
+     heading: "Can you provide more detail?",
-     isPageHeading: true,
      size: "l"
    },
    name: "more-detail",
    maxlength: 350,
    countType: "characters"
  }) }}
```

Other heading options can be configured if necessary. For example, using `heading.id` and `heading.level` when a label acts as an `<h2>` heading for a landmark:

```patch
- <aside>
+ <aside aria-labelledby="search-heading">
    {{ searchInput({
      label: {
        text: "Search",
-       size: "s"
+       size: "s",
+       heading: {
+         id: "search-heading",
+         level: 2
+       }
      }
    }) }}

    <!-- // … -->
  </aside>
```

## Nunjucks heading options

For consistency with other components, legends now support the following Nunjucks options:

- `id`
- `caption`
- `visuallyHiddenText`
- `heading`
- `attributes`

Labels now also support:

- `caption`
- `visuallyHiddenText`
- `heading`

[...]

PR #1708, opened 2025-11-26
Add label, legend and table caption size option

colinrotherham:
## Description

This PR adds support for:

1. Separate `legend()` component (for internal use)
1. Nunjucks `call` for labels and legends
1. Nunjucks `size` param for label and legend (`s`, `m`, `l`, `xl`)
1. Nunjucks `captionSize` param for tables (`s`, `m`, `l`, `xl`)

Split out from https://github.com/nhsuk/nhsuk-frontend/pull/1670

### Labels

```patch
  {{ input({
    label: {
      text: 'What is your full name?',
-     classes: "nhsuk-label--l"
+     size: "l"
    }
  }) }}
```

### Legends

```patch
  {{ radios({
    fieldset: {
      legend: {
        text: "How do you want to be contacted about this?",
-       classes: "nhsuk-fieldset__legend--l"
+       size: "l"
      }
    },
    items: []
  }}
```

### Labels

```patch
  {{ input({
    label: {
      text: "What is your full name?",
-     classes: "nhsuk-label--l"
+     size: "l"
    }
  }) }}
```

### Tables

We've added a new `captionSize` Nunjucks option to tables as an alternative to the caption modifier classes. For example:

```patch
  {{ table({
    caption: "Skin symptoms and possible causes",
-   captionClasses: "nhsuk-table__caption--l",
+   captionSize: "l",
    rows: []
  }) }}
```

## Checklist

- [x] CHANGELOG entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -98,6 +98,50 @@ We've added a new `divider` Nunjucks option on select items to support this feat
 
 This was added in [pull request #1701: Support showing dividers between select options](https://github.com/nhsuk/nhsuk-frontend/pull/1701).
 
+#### Add a 'size' Nunjucks option to labels and legends
+
+We've added a new `size` Nunjucks option to labels and legends as a simpler alternative to the size modifier classes. For example:
+
+```patch
+  {{ input({
+    label: {
+      text: 'What is your full name?',
+-     classes: "nhsuk-label--l"
++     size: "l"
+    }
+  }) }}
+```
+
+```patch
+  {{ radios({
+    fieldset: {
+      legend: {
+        text: "How do you want to be contacted about this?",
+-       classes: "nhsuk-fieldset__legend--l"
++       size: "l"
+      }
+    },
+    items: []
+  }}
+```
+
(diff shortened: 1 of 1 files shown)


colinrotherham:
>What's the rationale behind extract Legend into a component? I know we have Label as a component, but I’ve always assumed that was because it’s used by multiple other components, whereas Legend is only used by Fieldset?

It's not for separate use (same as the `errorMessage` macro) but makes testing and documentation easier

E.g. I'd like to start [using the `isComponent: true` config like GOV.UK Frontend does](https://github.com/nhsuk/nhsuk-service-manual/issues/2371)

But it's also partly a baby step towards:

* https://github.com/alphagov/govuk-frontend/issues/1079
* https://github.com/alphagov/govuk-frontend/issues/1166

Where maybe a future `formGroup()` macro could glue it all together

We compose custom groups of fields a lot on our screening service

frankieroberto:
@colinrotherham I'm slightly wary of creating components _just_ for dev/testing reasons, particularly where we don't expect/want users to use them directly. (React seems to take the opposite approach with lots of nesting of components)

But sounds like there's some reasonable rationale for it here. And we do already have `hint()` as a component with a documentation for it (unlike the GOVUK design system).

PR #1711, opened 2025-11-27
Review global `box-sizing` usage for v11

colinrotherham:
This PR was recreated from https://github.com/nhsuk/nhsuk-frontend/pull/1633

It was reverted from v10 and is now ready for v11

---

## Description

Globally setting `box-sizing` using best practice from a decade ago is causing problems in components, not least checkboxes and radios inside `nhsuk-details__text` wrappers.

Best practice as of 2025 is to [not set this value](https://www.oddbird.net/2025/09/04/box-model/) globally — certainly if starting a project from scratch, and using intrinsic layouts which don't use fixed widths.

An alternative approach is set this value globally using the `*` selector, and not attempt to force inheritance. However, as this doesn’t encompass pseudo elements, this then means adding this value explicitly in such cases where it is needed.

~~I tried removing the global `box-sizing` reset, but a number of tests failed. Although not beyond the realms of fixing, I have kept the global default for `box-sizing` as `border-box`, removed it from rules where it is inherited and kept it for the pseudo elements used to draw checkboxes and radios.~~

~~The only visual differences are the space between the underlined NHS logo.~~

I’ve chosen the former approach; not setting a global `box-sizing` value and instead setting it where necessary. This is especially important as the global `box-sizing` was never imported on a per-component basis. 

I’ve also added an example of small radios and checkboxes inside details elements… arguably this example page can now be removed, but maybe we want to keep it around for a while.

## Checklist

- [x] CHANGELOG entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,14 @@ If you're using the following tablet width utility classes, you must:
 
 This change was introduced in [pull request #1296: Apply grid classes from tablet (not desktop)](https://github.com/nhsuk/nhsuk-frontend/pull/1296).
 
+#### Remove global box sizing reset
+
+We have removed the global `box-sizing` reset and added `box-sizing: border-box` only where necessary.
+
+Please review any custom styles, especially those with defined widths, to make sure they have a correctly calculated box size.
+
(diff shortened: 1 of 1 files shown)


colinrotherham:
We'll need to edit the changelog and move the entry to **Breaking changes** before merging

Blocked by https://github.com/nhsuk/nhsuk-frontend/pull/1631 when all the v11 things can be merged

colinrotherham:
Ready for review

colinrotherham:
This can be approved and merged now @paulrobertlloyd 🙏

Can't approve my own PR 🤦‍♂️

paulrobertlloyd (approved):

PR #1712, opened 2025-11-27
[v10] Add support for icon buttons

colinrotherham:
## Description

This PR [adds icons](https://service-manual.nhs.uk/design-system/styles/icons) to buttons using the `icon` Nunjucks options.

For example, using `icon: "search"` to add an icon-only search button to a text input:

```njk
{{ input({
  formGroup: {
    afterInput: {
      html: button({
        ariaLabel: "Find",
        icon: "search",
        small: true
      })
    }
  }
}) }}
```

With support for both an icon and text shown together:

```patch
  html: button({
-   ariaLabel: "Find",
+   text: "Find",
    icon: "search",
    small: true
  })
```

Or to change the icon and adjust placement:

```patch
 html: button({
   text: "Find",
-   icon: "search",
+   icon: {
+     name: "arrow-right",
+     placement: "end"
+   }
 })
```

It was split out from https://github.com/nhsuk/nhsuk-frontend/pull/1660

Inputs and icon buttons

<br>
<br>

Inputs and icon buttons (mobile)

<br>
<br>

Inputs and icon buttons (watch)

## Checklist

- [x] CHANGELOG entry

### DIFF
--- a/packages/nhsuk-frontend/src/nhsuk/components/button/_index.scss
+++ b/packages/nhsuk-frontend/src/nhsuk/components/button/_index.scss
@@ -223,7 +223,7 @@ $button-shadow-size: $nhsuk-button-shadow-size;
     align-items: center;
 
     // Add minimum width for icon-only buttons
-    min-width: nhsuk-px-to-rem(40px);
+    min-width: nhsuk-px-to-rem(44px);
 
     // Prevent users from selecting icon button text
     // e.g. When double clicking the button
@@ -241,7 +241,7 @@ $button-shadow-size: $nhsuk-button-shadow-size;
 
   .nhsuk-button:not(.nhsuk-button--small) {
     @include nhsuk-media-query($from: tablet) {
-      min-width: nhsuk-px-to-rem(62px);
+      min-width: nhsuk-px-to-rem(66px);
     }
   }
 


anandamaryon1:
Reviewing this, oddly on iOS the text inputs with icon buttons lack background white. Other inline inputs are fine. 

image

anandamaryon1:
NVDA - icon only buttons announced as eg. 'Find graphic find button' and 'Accept graphic accept button'. 

Voiceover is fine, just, 'Find, button' and 'Accept, button'.

anandamaryon1:
Aside form the above, all looking good.

I'm thinking we merge this but don't (yet at least) publish them in the design system, to avoid encouraging icon-only buttons. If we have some good use-cases, ideas for guidance and testing in future we could add them. But merging them at least means we can use them in a search component.

colinrotherham:
>NVDA - icon only buttons announced as eg. 'Find graphic find button' and 'Accept graphic accept button'.

Ah, even with `aria-hidden="true"`?

I wonder if this is unique to buttons and we need `aria-label` instead (rather than trust the inner text)

anandamaryon1:
Had a play on NVDA, it seems to work OK if the aria-label is on the button instead of the SVG. Saying just "Find, button" for the icon-only button. 

Maybe it get confused if it's `aria-hidden` but also has an `aria-label`?

anandamaryon1:
Any ideas on the [greyed out inputs on iOS](https://github.com/nhsuk/nhsuk-frontend/pull/1712#issuecomment-3848075067)? Very odd…

colinrotherham:
>Any ideas on the https://github.com/nhsuk/nhsuk-frontend/pull/1712#issuecomment-3848075067? Very odd…

Yeah it's automatic but not sure why

It's some sort of feature where it tries to auto-match the background colour, seemed to come in Safari 15 when support for alternative theme colours via `prefers-color-scheme: dark` landed

The header search was always grey too until we fixed that recently in https://github.com/nhsuk/nhsuk-frontend/commit/8d5084dcbf510b692f8f4ba5c7a8ab1adb107400

Search for `$nhsuk-input-background-colour` and you'll see we set a white background for:

* Checkboxes
* Radios
* Header input
* Select

colinrotherham:
>Had a play on NVDA, it seems to work OK if the aria-label is on the button instead of the SVG. Saying just "Find, button" for the icon-only button.
>
>Maybe it get confused if it's `aria-hidden` but also has an `aria-label`?

Makes sense

Knowing [GOV.UK Frontend icons had similar issues](https://github.com/alphagov/govuk-frontend/pull/1724) let's keep `aria-hidden="true"` set on the nested SVG and move the text label to the button via `aria-label` instead

colinrotherham:
@anandamaryon1 Examples updated with `aria-label`:
https://nhsuk-frontend-pr-1712.herokuapp.com/nhsuk-frontend/examples/form-alignment/

anandamaryon1:
Nice, works well for me on NVDA now 👍

anandamaryon1:
> > Any ideas on the [#1712 (comment)](https://github.com/nhsuk/nhsuk-frontend/pull/1712#issuecomment-3848075067)? Very odd…
> 
> Yeah it's automatic but not sure why
> 
> It's some sort of feature where it tries to auto-match the background colour, seemed to come in Safari 15 when support for alternative theme colours via `prefers-color-scheme: dark` landed
> 
> The header search was always grey too until we fixed that recently in [8d5084d](https://github.com/nhsuk/nhsuk-frontend/commit/8d5084dcbf510b692f8f4ba5c7a8ab1adb107400)
> 
> Search for `$nhsuk-input-background-colour` and you'll see we set a white background for:
> 
> * Checkboxes
> * Radios
> * Header input
> * Select


Ahh yes, iOS is adding:
image
for `type="search"`. So it seems we need to override that.

colinrotherham:
This is the iOS default for regular text inputs:

```css
background-color: canvas;
color: canvastext
```

The GOV.UK Frontend uplift added this warning:

https://github.com/nhsuk/nhsuk-frontend/blob/49de5273743ee66528ef905f5cec862b3353d910/packages/nhsuk-frontend/src/nhsuk/components/input/_index.scss#L22-L25

So we should test our select menu fix which goes against that:

https://github.com/nhsuk/nhsuk-frontend/blob/49de5273743ee66528ef905f5cec862b3353d910/packages/nhsuk-frontend/src/nhsuk/components/select/_index.scss#L30-L31

Setting `background-color: canvas;` works the same in iOS but (in theory) is dark mode friendly

anandamaryon1:
> This is the iOS default for regular text inputs:
> 
> ```css
> background-color: canvas;
> color: canvastext
> ```
> 
> The GOV.UK Frontend uplift added this warning:
> 
> https://github.com/nhsuk/nhsuk-frontend/blob/49de5273743ee66528ef905f5cec862b3353d910/packages/nhsuk-frontend/src/nhsuk/components/input/_index.scss#L22-L25
> 
> So we should test our select menu fix which goes against that:
> 
> https://github.com/nhsuk/nhsuk-frontend/blob/49de5273743ee66528ef905f5cec862b3353d910/packages/nhsuk-frontend/src/nhsuk/components/select/_index.scss#L30-L31
> 
> Setting `background-color: canvas;` works the same in iOS but (in theory) is dark mode friendly

Looks like that Firefox bug was fixed 5 years ago. Fine in latest but…

Firefox 60, it is broken, select text is unreadable. Along with primary icon buttons.
image

```
background-color: canvas;
color: canvastext;
```

Fixes the iOS input background display issue. 👍

colinrotherham:
@anandamaryon1 I've updated the changelog entry regarding `"aria-label"` and the header search

Shall we pick up the search input background colour issue in https://github.com/nhsuk/nhsuk-frontend/pull/1660?

We should raise a separate issue for the Firefox 60 custom colours too

This PR now targets v10.x

colinrotherham:
I've updated the [**All buttons** example page](https://nhsuk-frontend-pr-1712.herokuapp.com/nhsuk-frontend/examples/all-buttons/) to test buttons, links and disabled states:

All buttons example page

edwardhorsford:
Great to see this capability coming in!

Design wise, I reckon when full width like this, the icon and text should stay together, centred. I find the separation of the icon left aligned somewhat weird.

Screenshot 2026-04-02 at 16 30 36

colinrotherham:
@edwardhorsford Thankfully, inline buttons only display like that below 320px (same with input prefix/suffix)

From mobile and up this is what you'll see:

Inline button with icon, mobile

<br>
<br>

On top of this comes [search input](https://github.com/nhsuk/nhsuk-frontend/pull/1660) and [stepper input component](https://github.com/nhsuk/nhsuk-frontend/pull/1719) where we'd only planned for icons to be shown inline with text inputs. I'll have a look at non-inline buttons, but this could come in a later PR

frankieroberto:
Looks good to me! 🙌 

It works with the 'user' icon too. Is that something we want to support? 🤷 

Screenshot 2026-04-08 at 14 32 10

colinrotherham:
>It works with the 'user' icon too. Is that something we want to support? 🤷

Thanks @frankieroberto

This PR primarily adds the ability to add icons rather than a specific feature

* For us, it unblocks the [search input](https://github.com/nhsuk/nhsuk-frontend/pull/1660) release
* For users, it unblocks [stepper inputs](https://github.com/nhsuk/nhsuk-frontend/pull/1719) and similar features

It was a suggestion from @paulrobertlloyd but you'll see not only `icon` or `icon.name` enabling support our own icons, but also `icon.html` bringing support for custom icons where the SVG markup can be passed in.

>Design wise, I reckon when full width like this, the icon and text should stay together, centred. I find the separation of the icon left aligned somewhat weird.

@edwardhorsford Your feedback to keep the text and icon together was applied too:

Inputs and icon buttons (watch)

edwardhorsford:
Nice @colinrotherham, I think that looks much nicer / more balanced.

colinrotherham:
No worries @edwardhorsford, always appreciate your input

PR #1643, opened 2025-10-24
Add small buttons and inline form groups

colinrotherham:
## Description

This PR extracts small button styles from https://github.com/nhsuk/nhsuk-frontend/pull/1600 without refactoring borders or focus styles

* [Button examples](https://nhsuk-frontend-pr-1643.herokuapp.com/nhsuk-frontend/examples/all-buttons/)
* [Button groups](https://nhsuk-frontend-pr-1643.herokuapp.com/nhsuk-frontend/examples/button-group/)
* [Form alignment](https://nhsuk-frontend-pr-1643.herokuapp.com/nhsuk-frontend/examples/form-alignment/)

4 of 4 stacked PRs in a 4-part mini series:

* https://github.com/nhsuk/nhsuk-frontend/pull/1657
* https://github.com/nhsuk/nhsuk-frontend/pull/1659
* https://github.com/nhsuk/nhsuk-frontend/pull/1656
* https://github.com/nhsuk/nhsuk-frontend/pull/1643 (this PR)

## Inputs with buttons

Inline buttons are now supported by text inputs using the Nunjucks option `afterInput`

Inputs with buttons

## Selects with buttons

Inline buttons are now supported by select menus using the Nunjucks option `afterInput`

Selects with buttons

## Inline form group support

Inline form groups are now supported via `.form-group--inline`

For example, from the NHS [Calculate your body mass index (BMI)](https://www.nhs.uk/health-assessment-tools/calculate-your-body-mass-index/) service:

Inline form groups showing pound and stone inputs

## Input wrapper renamed

To fix a slight BEM issue, the input wrapper used by prefixes and suffixes has been renamed:

```patch
  <div class="nhsuk-form-group">
    <h1 class="nhsuk-label-wrapper">…</h1>
-   <div class="nhsuk-input__wrapper">
+   <div class="nhsuk-input-wrapper">
```

This change is backwards compatible and we can remove the old class in a future release

## Checklist

- [x] Tested against our [testing policy](https://github.com/nhsuk/nhsuk-frontend/blob/master/docs/contributing/testing.md) (Resolution, Browser & Accessibility)
- [x] Follows our [coding standards and style guide](https://github.com/nhsuk/nhsuk-frontend/blob/master/docs/contributing/coding-standards.md)
- [x] CHANGELOG entry

PR #1719, opened 2025-11-28
[WIP] Add stepper input component

colinrotherham:
## Description

This PR adds a new stepper input component with +/- stepper buttons

It was built using prototype from @edwardhorsford and utilises:

* https://github.com/nhsuk/nhsuk-frontend/pull/1643
* https://github.com/nhsuk/nhsuk-frontend/pull/1712

**To do:** Add screen reader status announcements ([see prototype code](https://github.com/NHSDigital/manage-breast-screening-prototype/blob/main/app/assets/javascript/stepper-input.js))

Number input with stepper buttons

## Checklist

- [ ] CHANGELOG entry


colinrotherham:
I've compared the prototype icons and they're slightly smaller than our default size

E.g. Here's the "Close" icon in red, rotated to compare with the new "Plus" icon

Icon size difference

edwardhorsford:
This is awesome to see @colinrotherham!

Some thoughts:
* I reckon we should avoid `numberInput` - to me this implies it's a general component for numbers, when in reality it's a subset. I think stepper / stepperInput might be the more conventional name?
* I didn't do it in my prototype, but you could conceive that it could optionally accept text to go inside each button - either alongside the icons or instead of.
* I reckon these icons are a bit large / weighty (very subjective though)
* In my prototype I used centred text for the input. That's probably unconventional but I felt made the component as a whole seem more balanced - would be interested in what others think.

anandamaryon1:
Very cool and works well. 

Agree with the decision to centre the number in the input, to better relate to to both stepper buttons. 

Being able to use arrow keys is a nice touch, guessing that's just the underlying input type number though.

Struggle to think of many use-cases for this though, don't think we have many services where you fill a basket of items. I suppose it helps in situations where the user is primarily using a mouse and can avoid breaking flow to use the keyboard? Or on touch devices where the keyboard is a little more unwieldy?

edwardhorsford:
@anandamaryon1 

There was a [good article by Nielson Norman](https://www.nngroup.com/articles/input-steppers/) group about them. Agree the usages are limited. Besides the mouse/tablet only scenario, I'd see them as being most effective where users may need to make small adjustments to a value relative to a current value.

You could speculate for NHS that might be adjusting dosages - 0, 25ml, 50ml, 75ml, etc. Or 'how many times a week do you do x' type surveys. Agreed it's probably not that common though!

anandamaryon1:
A stepper in the wild: https://ukhsa-dashboard.data.gov.uk/cover/childhood-vaccinations

image

PR #1734, opened 2025-12-04
[v10] Fix appearance of summary lists alongside other elements within card content

paulrobertlloyd:
## Description

We currently force the removal of a bottom margin on summary lists within cards. However, if there is an element that follows it, they now collide:

Example of a summary list within a card, followed by a button.

<br>
<br>

`.nhsuk-card__content` already removes the bottom margin from the last item in a card, so zeroing out `margin-bottom` on `.nhsuk-card__content .nhsuk-summary-list` isn’t necessary.

In addition, we should only apply the last row adaptions when the summary list is the last item within `.nhsuk-card__content`.

This PR fixes these issues, and adds examples and visual regression test images.

## Checklist

- [x] CHANGELOG entry

### DIFF
--- a/packages/nhsuk-frontend/src/nhsuk/components/summary-list/_index.scss
+++ b/packages/nhsuk-frontend/src/nhsuk/components/summary-list/_index.scss
@@ -189,7 +189,8 @@
       margin-bottom: 0;
     }
 
-    .nhsuk-summary-list:last-child .nhsuk-summary-list__row:last-of-type > :last-child {
+    .nhsuk-summary-list--no-border:last-child .nhsuk-summary-list__row:last-of-type > :last-child,
+    .nhsuk-summary-list:last-child .nhsuk-summary-list__row--no-border:last-of-type > :last-child {
       @include nhsuk-media-query($until: tablet) {
         margin-bottom: 0;
       }


colinrotherham:
I've rebased this onto v10 and added https://github.com/nhsuk/nhsuk-frontend/pull/1734/commits/eab3332e0e40c34df605b263395fb33ddcd66f56 to prefer `--no-border`

colinrotherham:
Changelog entry added

paulrobertlloyd:
Ah, now that you bring the `--no-border` modifiers into it…

In 3f339a8, I’ve changed this so that we only remove the bottom margin on the last child of the last row of the last summary list if:

- the entire summary list has no borders
- the last summary row has no border

This means that, in effect:

- We remove the bottom margin of the last summary row in the last summary list in a card (overriding the bottom margin that gets added to all summary list rows below the tablet breakpoint)
- We remove the bottom margin of the last child of the last summary row in the last summary list in a card, if that row does not have a bottom border.

Make sense? 🤪

packages/nhsuk-frontend/src/nhsuk/components/card/fixtures.mjs
```diff
@@ -71,7 +72,109 @@ export const examples = {
...
+      viewports: ['mobile', 'tablet']
+    }
+  },
+  'basic with summary lists': {
+    context: {
+      heading: 'Regional Managers',
+      headingLevel: 3
+    },
+    callBlock: outdent`
+      <h4 class="nhsuk-heading-s nhsuk-u-margin-bottom-1">East</h4>
```
colinrotherham, line 86: Does this bottom margin tweak need turning into a proper fix too?

colinrotherham:
Shall we get this merged @anandamaryon1?

packages/nhsuk-frontend/src/nhsuk/components/card/fixtures.mjs
paulrobertlloyd: Let’s punt on this for now… I’m forever changing the bottom margin of headings, so it might be a broader change.

frankieroberto (approved):

PR #1737, opened 2025-12-10
Use consistent inset padding across components

### LINKED ISSUE #1735: Ensure inset padding is consistent across components
We have a number of components that have inset padding, but this ranges from 24px to 32px:

| Component | Border | Padding | Total distance from left edge |
| - | - | - | - |
| (Care) card | 1px | 32px | **33px** |
| Contents list (item) | - | 32px | **32px** |
| Error summary | 4px | 24px | **28px** |
| Expander | 1px | 32px | **33px** |
| File upload (upcoming, see #1556) | 2px | 26px | **28px** |
| Inset text | 8px | 24px | **32px** |
| Notification banner | 4px | 24px | **28px** |
| Pagination | - | 32px | **32px** |
| Tabs (panel) | 1px | 24px | **25px** |
| Warning callout | 1px | 32px | **33px** |

Here is an image that shows these components together:

Image

There may be further differences at narrower breakpoints.

It would be good to standardise the padding we use across components for visual consistency, if nothing else.

paulrobertlloyd:
## Description

Fixes #1735.

Using **24px** across the following components across all breakpoints:

- (Care) card
- Contents list
- Error summary
- Expander
- Form group object
- Inset text
- Notification banner
- Pagination
- Tabs (panel)
- Warning callout

File upload (#1556) will need the same values applied, once that is merged.

24px is a size that works well across all different components. It also works for components such as details and pagination where text with left padding can align with the global value, but not if it gets much bigger. While cards have reduced padding, this works better for cards within card groups with many columns.

I’ve added this as a global setting: `$nhsuk-inset-size`.

Take a look at the updated reference images for comparison.

In #1740 I added a page with components that have inset padding, and in this PR you can [see this page with the inset spacing highlighted](https://nhsuk-frontend-pr-1737.herokuapp.com/nhsuk-frontend/examples/inset#show-guide).

I’ve also aligned the design for mobile tab links to match that of the contents list; we may want to pull this out into a separate PR.

## Checklist

- [ ] CHANGELOG entry


packages/nhsuk-frontend/src/nhsuk/components/tabs/_index.scss
```diff
@@ -33,13 +33,21 @@
...
+
+      display: block;
+
+      position: absolute;
+      top: nhsuk-em(12px);
+      left: 0;
+
+      width: nhsuk-spacing(3);
+
+      border-top: 1px solid $nhsuk-border-hover-colour;
```
paulrobertlloyd: This is the change added to align mobile tab list items with that of the contents list component.

anandamaryon1:
Love the rationalisation, it's been needed for a long time ❤️ 

And love the factoring of border sizes to get things to line up properly. 

On the actual value, I like 24px on small screens, it matches our current card padding. But on bigger screens, I feel we should retain the baseline of 32px for cards, and therefore other components too. 24px feels a bit tight and busy for content-focussed layouts on bigger screens, eg.

image

I accept that in general our spacing is quite generous for more complex UIs so perhaps we can attempt to solve this with a global modifier that will adjust paddings and margins for 'dense' interfaces? As a sister PR to this or as a follow up later on.

paulrobertlloyd:
@anandamaryon1 I was worried you’d say that! Perhaps cards (and cards alone?) need a different behaviour above, say for above the desktop breakpoint?

Making the size any larger and applying it to all components isn’t tenable as for some components, like form fields with errors, and pagination arrows, greater spacing means the relationship is lost between the different elements of the design.

anandamaryon1:
Feel like I'm always befouling your neat work, sorry! 

On form fields with errors, yes 32px is too much. 24px is about ok. I think it's OK that they don't align with the other components, they are temporary / situational and might even benefit from standing out of line to get noticed. 

I'd argue pagination doesn't need to be included in this consolidation, since the padding isn't empty space and the subheading breaks though, making it visually a block in itself. 

Tabs, it's slightly regrettable that the tab buttons get a little wider but the contents within seems about right at 32px, particularly when compared with cards which are visually very similar. 

I propose that we exclude pagination, details and form field errors from this, and go with 32px on desktop for the remaining components. Edit: and exclude details, where the arrow plus label can be seen as a block itself too. Edit 2: and action links. So essentially, only apply this consistent padding to boxy / card / panel things. 

Happy for others thoughts on this too though.

packages/nhsuk-frontend/src/nhsuk/components/tabs/_index.scss
paulrobertlloyd: Split out into a separate PR: https://github.com/nhsuk/nhsuk-frontend/pull/1761

colinrotherham:
@paulrobertlloyd I'm just giving this a rebase

It's hit an interesting point now that https://github.com/nhsuk/nhsuk-frontend/pull/1655 has merged

For action links, details and primary cards with chevrons we actually can't have a fixed inset size. Otherwise as text-only scaling kicks in (and the associated icon scales with it) we see the inset become too small 😬

Action link and details component with larger text

<br>
<br>

For the chevron "safe area" we can keep your approach but I'll use rems for the icon width:

Card primary with chevron and larger text

colinrotherham:
@paulrobertlloyd I do really love this PR

I've converted the left/right inset values to rems to solve https://github.com/nhsuk/nhsuk-frontend/pull/1737#issuecomment-3728193752

### The danger of rems versus pixels

Using rems for everything would prevent text-only zoom from doing its job

i.e. You'd increase the text size and the page would zoom instead (eventually you'd get the mobile layout)

To avoid this I've only used rems for the left/right inset padding

Perhaps a nice enhancement would be to add `$units: "rem"` support to our spacing helpers:

```scss
@include nhsuk-responsive-padding(3, "left", $unit: "rem");
@include nhsuk-responsive-padding(3, "right", $unit: "rem");
```

colinrotherham:
I've pushed up the suggestion from https://github.com/nhsuk/nhsuk-frontend/pull/1737#issuecomment-3728824905

E.g. Here's how to set 24px mobile, 32px tablet, complete with border offset

```scss
@include nhsuk-responsive-padding(5, ("top", "bottom"), $adjustment: - $nhsuk-border-width);
@include nhsuk-responsive-padding(5, ("left", "right"), $adjustment: - $nhsuk-border-width, $unit: "rem");
```

So we don't actually need `$nhsuk-inset-size`?

anandamaryon1:
I've made the non-card components all have 24px padding. All components have 24 on mobile. And now only card components have 32px on desktop. Also added and aligned file upload and panel components. Tested with font-scaling. 

So what we have now, on tablet and above:
### Not cards
* action link _-does not comply-_
* pagination _-does not comply-_ 
* contents list 24
* details 24
* error message 24

### Cards or card-adjacent 
* error summary 24/32
* notification banner 24/32
* file upload 24/32
* expander 24/32
* cards, inc. care cards, do don't cards, feature cards 24/32
* panel 24/32
* warning 24/32
* tabs 24/32
* inset text, (controversially? but keeping as-is) 24/32

anandamaryon1:
Happy with this as it stands @paulrobertlloyd ? 

I'm thinking about getting this into v11.

paulrobertlloyd:
> Happy with this as it stands @paulrobertlloyd?

Mostly. But one last thing… shouldn’t the padding for inset text align with that for details content and fields with an error, which share a very similar design?

In which case, this would mean the left padding for inset text would be reduced to 16px.

## Before

before

## After

after

---

Also, should we update the togglable overlay guides on this page so that they also visibly show the 24px inset size too?

colinrotherham:
Rebased, but a reminder to say the wider spacing (tablet) isn't great for tabs

Tabs_tablet

anandamaryon1:
> > Happy with this as it stands @paulrobertlloyd?
> 
> Mostly. But one last thing… shouldn’t the padding for inset text align with that for details content and fields with an error, which share a very similar design?

Looking at them together, yes. But, when looked at alone, or alongside other components, I feel 16px is too close for the inset text. With such a heavy blue border the added empty space is needed to make the text not feel cramped, in my opinion. And I feel this trumps consistency with those two specific other components because they are less likely to be used right next to an inset text. 

Eg.
| 16px      | 24px |
| ----------- | ----------- |
| image      | image       |




> Also, should we update the togglable overlay guides on this page so that they also visibly show the 24px inset size too?

Yeah that'd be helpful 👍 

--

> Rebased, but a reminder to say the wider spacing (tablet) isn't great for tabs

This isn't great is it, thanks @colinrotherham. Maybe it wouldn't be crazy to have the tabs not align, to maximise available space.

For example, 16px, similar to buttons: 
image

colinrotherham:
I've rebased this against https://github.com/nhsuk/nhsuk-frontend/pull/1961 and had another go at making pagination comply

We've still got [issues with Tabs spacing](https://github.com/nhsuk/nhsuk-frontend/pull/1737#issuecomment-4104290788) and I'm still not sure about [inset text versus error summary spacing](https://github.com/nhsuk/nhsuk-frontend/pull/1737#issuecomment-4104290788)

Tweaks to pagination

colinrotherham:
The narrower error summary spacing looks nice and compact for small form controls though:

Compact nested errors

PR #1741, opened 2025-12-15
[v10] Update contents list dash colour to improve contrast

anandamaryon1:
## Description
Fixes https://github.com/nhsuk/nhsuk-service-manual-community-backlog/issues/508

Open to counter suggestions on how to make the checklist dashes `grey-2`. 

## Checklist

- [x] CHANGELOG entry

### DIFF
--- a/packages/nhsuk-frontend/src/nhsuk/components/contents-list/_index.scss
+++ b/packages/nhsuk-frontend/src/nhsuk/components/contents-list/_index.scss
@@ -5,9 +5,6 @@
 ////
 /// Contents list component
 ///
-/// 1. Creates a grey line before each list
-///    item using a ASCII number for the symbol.
-///
 /// @group components/contents-list
 ////
 


packages/nhsuk-frontend/src/nhsuk/core/settings/_colours-applied.scss
```diff
@@ -107,6 +107,14 @@ $nhsuk-success-colour: nhsuk-colour("green") !default;
 
 $nhsuk-border-colour: nhsuk-colour("grey-4") !default;
 
+/// Border dark colour
+///
+/// Passes 3:1 on page background. Used in contents list.
+///
+/// @type Colour
+
+$nhsuk-border-dark-colour: nhsuk-colour("grey-2") !default;
```
colinrotherham: Thanks @anandamaryon1

I feel like this should be `nhsuk-border-text-colour` to pair with `nhsuk-text-colour`

```suggestion
$nhsuk-border-text-colour: nhsuk-colour("grey-2") !default;
```

colinrotherham:
@anandamaryon1 I've changed the target branch to `support/10.x`

~Think we need a (maybe "Fixes") changelog entry?~

I've updated the failing unit tests, plus added a changelog entry and new reference images

packages/nhsuk-frontend/src/nhsuk/core/settings/_colours-applied.scss
colinrotherham: I've applied this suggestion but tell me if you preferred it before

colinrotherham (approved):

anandamaryon1:
> @anandamaryon1 I've changed the target branch to `support/10.x`
> 
> ~Think we need a (maybe "Fixes") changelog entry?~
> 
> I've updated the failing unit tests, plus added a changelog entry and new reference images

Appreciated, thank you

packages/nhsuk-frontend/src/nhsuk/core/settings/_colours-applied.scss
anandamaryon1: I'm not sure, text implies it's to be used on text whereas this shouldn't be, since it's only 3:1 contrast on page background.

colinrotherham: I initially suggested a flipped ~`$nhsuk-border-text-colour`~ `$nhsuk-text-border-colour` if that's better?

As in, it's purpose is for borders that divide or decorate text content

Reasoning behind not using "dark" is that actually it's quite light

Alternatively we could skip the Sass global and just use `nhsuk-colour("grey-2")` directly?

anandamaryon1: Since it's currently only used in the contents-list component, perhaps we should just use the colour directly for now, and extract it out in future if needed elsewhere too.

colinrotherham: No problem, sorted it. Merge if you're happy

Thanks

packages/nhsuk-frontend/src/nhsuk/components/contents-list/_index.scss
colinrotherham: @anandamaryon1 What about this comment?

Maybe we used to use a symbol before `border-top`

anandamaryon1: Well spotted. Agree, it is referring to an [older implementation](https://github.com/nhsuk/nhsuk-frontend/commit/d23969935eb359166046ed8297ef874c49f23bfa#diff-989970a19fff9b59309a3cf64f5c710e2e1bd5bbcf7f70bfbd0e9655e0001bb0). Let's remove.

PR #1367, opened 2025-06-06
Updates to link styles and link hover states

frankieroberto:
This updates links (and link-like elements such as details) so that the underline is consistently thinner and offset further away from the link text, improving readability.

Links also have a clearer hover state using dark blue (`#003087`) instead of dark pink (`#7c2855`), and the underline gets thicker instead of disappearing, to make the link stand out more to users.

It also introduces new modifier classes for:

- text colour links (`nhsuk-link--text-colour`)
- links without underlines (`nhsuk-link--no-underline`)

<br>

| Before | After |
|----|-----|
| Links before  | Links after |

## Checklist

- [ ] Decide which components should have underlines and which should not
- [ ] Tested against our [testing policy](https://github.com/nhsuk/nhsuk-frontend/blob/master/docs/contributing/testing.md) (Resolution, Browser & Accessibility)
- [x] Follows our [coding standards and style guide](https://github.com/nhsuk/nhsuk-frontend/blob/master/docs/contributing/coding-standards.md)
- [x] CHANGELOG entry

PR #1746, opened 2025-12-18
Sortable table tweaks

frankieroberto:
This suggests some tweaks to the work-in-progress sortable table design in #1654.

Main changes:

* fix for the bug where the click area doesn't match the hover colour area
* fix for bug showing hover colour on non-sortable column header
* updated with the latest link styles from #1367 (thicker underline and darker blue on hover)
* switch to the background hover colour from the Task list component (a lighter blue rather than light grey)
* the blue border on the bottom of the currently-sorted header column now only extends as far as the content in the column (which has a responsive padding on the right, except for the last column where it’s on the left)

The last bullet is perhaps the most experimental, as it means the blue border doesn't match the width of the hover colour, which looks a bit odd. However when not hovering it perhaps looks better as it matches the content instead of going right up to the next column?

### DIFF
--- a/packages/nhsuk-frontend/src/nhsuk/i18n.mjs
+++ b/packages/nhsuk-frontend/src/nhsuk/i18n.mjs
@@ -141,7 +141,7 @@ export class I18n {
   hasIntlPluralRulesSupport() {
     return Boolean(
       'PluralRules' in window.Intl &&
-        Intl.PluralRules.supportedLocalesOf(this.locale).length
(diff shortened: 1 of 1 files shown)


anandamaryon1:
Thanks for these @frankieroberto they're great. 

Only bit I'm unsure of is the blue border on the bottom of the currently-sorted header column. It looks a little odd to me and it won't work always on the last column because it lacks left padding, so the content may sometimes extend beyond the blue border. 

e.g 
image

I'll merge this but feel free to reopen the branch and PR for any further tweaks.

PR #1764, opened 2026-01-12
Align pagination component styles and markup with equivalent GOV.UK Frontend component

### LINKED ISSUE #1632: Pagination component uses different markup from equivalent component in GOV.UK Frontend
There are a number of differences in the numbered pagination markup compared to that used in the equivalent component used for GOV.UK Frontend.

Replacing `nhsuk` and `govuk` with `brand` so as to reduce the number of differences gives the following diff:

```diff
- <nav class="brand-pagination" aria-label="Pagination">
+ <nav class="brand-pagination nhsuk-pagination--numbered" role="navigation" aria-label="Pagination">
-   <div class="brand-pagination__prev">
-     <a class="brand-link brand-pagination__link" href="#" rel="prev">
+     <a class="brand-pagination__previous" href="#" rel="prev">
-       <svg><!--GOV.UK left arrow icon--></svg>
+       <svg><!--NHS.UK left arrow icon--></svg>
-       <span class="brand-pagination__link-title">
+       <span class="brand-pagination__title">
-         Previous<span class="brand-u-visually-hidden"> page</span>
+         Previous<span class="brand-visually-hidden"> page</span> 
        </span>
      </a>
-   </div>
    <ul class="brand-pagination__list">
      <li class="brand-pagination__item">
-       <a class="brand-link brand-pagination__link" href="#" aria-label="Page 1">
+       <a class="brand-pagination__link" href="#" aria-label="Page 1">
          1
        </a>
      </li>
      <li class="brand-pagination__item brand-pagination__item--ellipsis">
        &ctdot;
      </li>
      <li class="brand-pagination__item">
-       <a class="brand-link brand-pagination__link" href="#" aria-label="Page 6">
+       <a class="brand-pagination__link" href="#" aria-label="Page 6">
          6
        </a>
      </li>
      <li class="brand-pagination__item brand-pagination__item--current">
        <a class="brand-pagination__link" href="#" aria-label="Page 7" aria-current="page">
          7
        </a>
      </li>
      <li class="brand-pagination__item">
-       <a class="brand-link brand-pagination__link" href="#" aria-label="Page 8">
+       <a class="brand-pagination__link" href="#" aria-label="Page 8">
          8
        </a>
      </li>
      <li class="brand-pagination__item brand-pagination__item--ellipsis">
        &ctdot;
      </li>
      <li class="brand-pagination__item">
-       <a class="brand-link brand-pagination__link" href="#" aria-label="Page 42">
+       <a class="brand-pagination__link" href="#" aria-label="Page 42">
          42
        </a>
      </li>
    </ul>
-   <div class="brand-pagination__next">
-     <a class="brand-link brand-pagination__link" href="#" rel="next">
+     <a class="brand-pagination__next" href="#" rel="prev">
-       <svg><!--GOV.UK right arrow icon--></svg>
+       <svg><!--NHS.UK right arrow icon--></svg>
-       <span class="brand-pagination__link-title">
+       <span class="brand-pagination__title">
-         Next<span class="brand-u-visually-hidden"> page</span>
+         Next<span class="brand-visually-hidden"> page</span> 
        </span>
      </a>
-   </div>
  </nav>
```

The main differences are that the NHS component:

- Requires a `--numbered` modifier to enable numbered pagination styles
- Includes extraneous `role="navigation"`
- Uses a `div` around previous and next links
- Uses `__previous` element class instead of `__prev`
- Uses `__title` element class instead of `__link-title`
- Doesn’t use `nhsuk-link` for pagination links

Also different, but unavoidable, are the SVG icons used for previous and next arrows (and the different visually hidden class name).

I’m not sure what differences there are for the block pagination variant, but a number of the above will still apply.

How many of these differences are we happy with, and how many do we want to align in a future upcoming breaking release?

paulrobertlloyd:
## Description

Fixes #1632.

## Checklist

- [ ] CHANGELOG entry


paulrobertlloyd:
@colinrotherham Try it now!

frankieroberto:
I'm not sure what alignment with GOVUK buys us with pagination, given that the embedded SVG element is always going to be different?

Using `nhsuk-pagination--numbered` was a bit of an annoying hack though, as I couldn’t find a way to otherwise make the numbered component backwards-compatible for the non-numbered variant. So would be happy to drop that in a v11.

for `__prev` vs `__previous` I wonder if we can persuade the GOVUK team to switch to `previous` (as a non-breaking change, still supporting `prev`)? The macro option is called `previous` so it’s a bit less confusing if both use the same term.

paulrobertlloyd:
With regards to `prev`, would referring you to https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/rel convince you that this is okay?

paulrobertlloyd:
Selfishly, beyond changing the icons, it would mean Mavis having to employ fewer monkey patches to get NHS frontend pagination styles working with the markup output by https://govuk-components.netlify.app/components/pagination/

frankieroberto:
@paulrobertlloyd ha, well if we’re going by the W3C specs then we should be using `color` and `referer`... 😉

But in general I support aligning with govuk frontend. However this will be a slightly painful breaking change for anyone who’s _only just_ implemented pagination...

Are there any bits of it we could do in smaller increments, or in a backwards-compatible way (changing the HTML but having the CSS still support the old class names)?

colinrotherham:
>Are there any bits of it we could do in smaller increments, or in a backwards-compatible way (changing the HTML but having the CSS still support the old class names)?

If you can weigh this against all the other deprecated things we're flagged for removal?

* https://github.com/nhsuk/nhsuk-frontend/pull/1553

frankieroberto:
@colinrotherham #1553 removes things already marked as deprecated for a little while? Maybe we could use a similar approach for pagination - make changes to the HTML, but have the CSS still support the old HTML and mark it as deprecated for a while (v11?) and then remove it in a future release (v12?).

colinrotherham:
In v10.3.0 we deprecated `imgURL`, `imgALT` and moved `caption` to `caption.text` and `caption.html`

Classes wise, we deprecated:

```
nhsuk-input__prefix
nhsuk-input__suffix
nhsuk-header__search-submit
nhsuk-table__panel-with-heading-tab
nhsuk-table__heading-tab
```

Is there any reason not to dual-support the necessary classes (and options?) in v10.3.1 instead of v11?

paulrobertlloyd:
Also, bearing in mind our outreach/contributions to the React and Jinja ports, teams using those frameworks won’t need to change anything.

frankieroberto:
dual-support for a transition period sounds good to me.

I think there enough teams using hardcoded HTML in templates, or other ports and implementations, that it’s worth trying to smooth the upgrade path, if possible?

PR #1765, opened 2026-01-13
Fix #1619: Implement grapheme cluster counting for character count co…

### LINKED ISSUE #1619: Character count component counts code points, not characters
# Bug Report

## What is the issue?

moshaid:
This PR implements grapheme cluster counting for the character count component to ensure client-side validation matches Python's `len()` behaviour for server-side consistency:

- Add graphemeCount utility function with Intl.Segmenter support and fallback
- Update character count component to use grapheme counting instead of UTF-16 code units
- Add comprehensive tests for Unicode, emoji, and complex characters
- Update component documentation with counting methodology
- Ensures client-side validation matches Python len() behaviour for server-side consistency


colinrotherham:
Thanks for raising this @moshaid

Better compatibility with Python `len()` and our [nhsuk-frontend-jinja](https://github.com/NHSDigital/nhsuk-frontend-jinja) port would be great

Knowing we only have [`Intl.Segmenter` in **Baseline 2024**](https://caniuse.com/wf-intl-segmenter), what should we do about the support gap?

* [Browserslist `supports es6-module` browsers](https://browsersl.ist/#q=supports+es6-module), versus
* [Browserslist `supports es6-module and not baseline 2024` browsers](https://browsersl.ist/#q=supports+es6-module+and+not+baseline+2024)

### Server-side consistency

Because the error message set server-side might include a different count to client-side, should we:

1. Support grapheme counting with string counting fallback
1. Support grapheme counting with no fallback, textarea only
1. Support grapheme counting via config only

Have you tried this already on your service?

Thanks

moshaid:
Appreciate you reviewing this @colinrotherham 

I have added a fall back to support the gap.
For the server consistency, i tried the 3rd option which i consider the best.

moshaid:
@colinrotherham i just ran a prettier check on the file and it is well formatted.  I am stuck on what could be wrong with the prettier check fail. Is there any option you could recommend on this?.

Thanks

colinrotherham:
Can you run `npm install` and try again?

I'm getting the following change on `npm run lint:prettier:fix`

Thanks @moshaid

```patch
--- A/packages/nhsuk-frontend/src/nhsuk/i18n.mjs
+++ B/packages/nhsuk-frontend/src/nhsuk/i18n.mjs
@@ -141,7 +141,7 @@
   hasIntlPluralRulesSupport() {
     return Boolean(
       'PluralRules' in window.Intl &&
+      Intl.PluralRules.supportedLocalesOf(this.locale).length
-        Intl.PluralRules.supportedLocalesOf(this.locale).length
     )
   }
```

moshaid:
Thanks for the tip @colinrotherham

packages/nhsuk-frontend/src/nhsuk/components/character-count/README.md
```diff
@@ -7,3 +7,84 @@ See the [main README quick start guide](https://github.com/nhsuk/nhsuk-frontend#
 ## Guidance and examples
 
 To learn more about the character count component and when to use it, visit the [design system in the NHS digital service manual](https://service-manual.nhs.uk/design-system/components/character-count) for guidance, examples and options.
+
+## How characters are counted
+
+By default, the character count component uses **code point counting**, which matches Python's `len()` function for Unicode strings. This ensures consistency between client-side (JavaScript) and server-side (Python) validation in `nhsuk-frontend-jinja`, preventing mismatched error messages.
```
MatMoore, line 13: Not everyone is using python and jinja, so I don't know if this paragraph should be geared at a broader audience than teams using python.

nhsuk-frontend-jinja is a port of the nunjucks components, so it's not actually the thing providing server-side validation, but wherever the template relies on the `len` filter, it will be using the code point counting, so it's good that its consistent.

This is a great contribution btw 👍🏻

colinrotherham: What do you both think about a customisable count function?

* https://github.com/alphagov/govuk-frontend/issues/1104
* https://github.com/alphagov/govuk-frontend/issues/1364

It might need to support promises (should a server-side count be needed)

Similar to a config option, you could pass in pre-exported count functions that we publish. Or alternatively use your own if necessary

packages/nhsuk-frontend/src/nhsuk/components/character-count/README.md
```diff
@@ -7,3 +7,84 @@ See the [main README quick start guide](https://github.com/nhsuk/nhsuk-frontend#
...
+
+You can optionally enable **grapheme cluster counting** (user-perceived characters) by setting `useGraphemeCounting: true` in the component configuration. This provides more accurate counting for:
+
+- **Emoji and emoji sequences**: Emoji like 👋, 👋🏼 (with skin tone), and 👨‍👩‍👧‍👦 (family emoji) are each counted as a single character
+- **Characters with combining marks**: Accented characters like é, ñ, and ü are counted correctly regardless of whether they're stored as a single code point or as a base character plus combining mark
+- **Complex scripts**: Non-Latin scripts (Chinese, Japanese, Korean, Arabic, etc.) are counted accurately
+
+**Important**: Only enable grapheme counting if your server-side validation also uses grapheme counting. Otherwise, you may see different counts between client and server validation messages.
```
MatMoore, line 21: Would it be helpful to signpost the way to do this in javascript and the various other languages we use on the backend? (In python I think you probably need a 3rd party library like https://grapheme.readthedocs.io/en/latest/grapheme.html)

packages/nhsuk-frontend/src/nhsuk/components/character-count/README.md
MatMoore: I like the idea in theory. I think calling the server for the count is a bit excessive though. Just being able to specify a count function on the client side to match what the server does seems like it would be flexible enough.

Counting bytes seems like the main use case other than the options we've got now (graphemes, codepoints, words). The byte count would depend on the encoding the server/database uses, so if we export a function for this it probably shouldn't assume utf-8.

I'm sure there are plenty of cases where we're constrained by some legacy backend thing, but I'm not sure the ideal UX in that case. With graphemes - and to a lesser extent, the existing codepoint counting - the display will track with what the user is entering, but it will lie about whether they are over the limit if they enter any multi-byte characters. But if they customise the count function to use byte counting, the count will jump by more than 1 when they type multi-byte characters, which might confuse users as well. 🤷🏻

frankieroberto: I wonder if we can partially mitigate mis-matches by:

* recommending that you set quite a high limit where possible
* recommending that teams use a `threshold`, so that most users don’t see the character count at all
* even if both client side and server-side are counting characters, you could allow an extra 5 or so characters server-side vs client-side, just in case there’s a discrepancy in the counting functions...

colinrotherham:
Thanks again for this @moshaid but we're going to close this and bring it into the team

With our work to align with GOV.UK Frontend we'd like to adhere closely to their approach ([see comment above](https://github.com/nhsuk/nhsuk-frontend/pull/1765#discussion_r2767751176)) especially with an [implementation plan added recently](https://github.com/alphagov/govuk-frontend/issues/1364#issuecomment-4161213597) by @romaricpascal

PR #1783, opened 2026-01-22
[v10] Add support for inline header search or account

colinrotherham:
## Description

This PR adds a new `nhsuk-header--inline` class

Header with inline log in

<br>
<br>

Header with inline log out

## Checklist

- [x] CHANGELOG entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,14 @@
 
 Note: This release was created from the `support/10.x` branch.
 
+### :new: **New features**
+
+#### Add a modifier class for header inline search or account links
+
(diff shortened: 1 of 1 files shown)


anandamaryon1:
I like this and think it'd be a good addition to the header - to make it work nicer on mobile for specific setups (one account link and shorter service names). 

Currently the changes seems to break the bigger account sections: 
image

colinrotherham:
>Currently the changes seems to break the bigger account sections:

@anandamaryon1 I've fixed this, I forgot to restrict it to `--inline`

Over to you for review

anandamaryon1:
Nice, tested on different browsers. 

I think this is useful. Will need to add something in the design system to document it. I'll open an issue for that.

colinrotherham:
@anandamaryon1 Should we add `inline: true` as an option?

Or just the `nhsuk-header--inline` class

PR #1796, opened 2026-02-04
[v10] Hide character count message until threshold is reached

colinrotherham:
## Description

The character count reserves space for the "visible count message" even when it's empty (e.g. due to threshold)

This PR hides the count message until the threshold is reached

<br>

Character count threshold

## Checklist

- [x] CHANGELOG entry

### DIFF
--- a/packages/nhsuk-frontend/src/nhsuk/components/character-count/character-count.mjs
+++ b/packages/nhsuk-frontend/src/nhsuk/components/character-count/character-count.mjs
@@ -236,7 +236,7 @@ export class CharacterCount extends ConfigurableComponent {
     // If input is over the threshold, show the count message
     if (this.isOverThreshold()) {
       this.$visibleCountMessage.removeAttribute('hidden')
-    } else if (!this.$visibleCountMessage.hasAttribute('hidden')) {
+    } else {
       this.$visibleCountMessage.setAttribute('hidden', '')
     }
 


packages/nhsuk-frontend/src/nhsuk/components/character-count/character-count.mjs
```diff
@@ -231,13 +231,21 @@ export class CharacterCount extends ConfigurableComponent {
...
-      !this.isOverThreshold()
-    )
+    const isOverThreshold = this.isOverThreshold()
+
+    // If input is over the threshold, show the count message
+    if (this.config.visibleCountMessageOptional) {
+      if (isOverThreshold) {
+        this.$visibleCountMessage.removeAttribute('hidden')
+      } else {
+        this.$visibleCountMessage.setAttribute('hidden', '')
```
MatMoore: if javascript is not available, should it start out hidden or disabled? (Does it matter?)

anandamaryon1:
I agree that it's probably better most of the time if you only have a threshold, to not reserve the space and accept the little layout bump when the text is added.

colinrotherham:
>I agree that it's probably better most of the time if you only have a threshold, to not reserve the space and accept the little layout bump when the text is added.

@anandamaryon1 Nice one. Happy to file it into "Changes" (non-breaking) and avoid the extra option?

packages/nhsuk-frontend/src/nhsuk/components/character-count/character-count.mjs
colinrotherham: It should start out hidden only when rendering an:

- empty visible count message with a `threshold` set but not yet reached
- empty textarea description with neither `maxlength` nor `maxwords` set

I've handled that 2nd case in `template.njk`

But we unfortunately can't assume hidden-by-default for `threshold` because a textarea description is still necessary if JavaScript is not available

What we can do though is remove the flicker when the textarea description is swapped over to a now-empty count message due to `threshold` which I've done in https://github.com/nhsuk/nhsuk-frontend/pull/1796/commits/8881ef51c27b01459bcc345e6e3af9b0732cda28

anandamaryon1:
> @anandamaryon1 Nice one. Happy to file it into "Changes" (non-breaking) and avoid the extra option?

Yes let's go with that, thanks.

colinrotherham:
@MatMoore @anandamaryon1 Ready for review again now it's a (non-breaking) change

colinrotherham:
Thanks @MatMoore

Are you happy with this approach @anandamaryon1?

anandamaryon1:
Could be smart but not sure it's needed, since it'd be reacting to what you are doing ie. deleting and typing text, I don't think it'd be too jarring. 

Revisit if we get any reports of issues.

PR #1797, opened 2026-02-05
[v10] Ensure components correctly pass label.id down to the label

MatMoore:
## Description

Previously label.id was being ignored.

This change is split off from https://github.com/nhsuk/nhsuk-frontend/pull/1719

For context, we have added a version of the stepper component to our service, manage breast screening. Without this fix we can't refer to the input's label with aria-described-by attributes.

## Checklist

- [x] CHANGELOG entry

### DIFF
--- a/packages/nhsuk-frontend/src/nhsuk/components/checkboxes/macro-options.mjs
+++ b/packages/nhsuk-frontend/src/nhsuk/components/checkboxes/macro-options.mjs
@@ -148,6 +148,11 @@ export const params = {
           'Subset of options for the label used by each checkbox item within the checkboxes component.',
         isComponent: true,
         params: {
+          id: {
+            type: 'string',
+            required: false,
+            description: 'The ID of the label tag.'
+          },
           classes: {
             type: 'string',
             required: false,
--- a/packages/nhsuk-frontend/src/nhsuk/components/radios/macro-options.mjs
+++ b/packages/nhsuk-frontend/src/nhsuk/components/radios/macro-options.mjs
@@ -136,6 +136,11 @@ export const params = {
           'Subset of options for the label used by each radio item within the radios component.',
         isComponent: true,
         params: {
+          id: {
+            type: 'string',
+            required: false,
+            description: 'The ID of the label tag.'
+          },
           classes: {
             type: 'string',
             required: false,


colinrotherham:
Could you edit the commit (new SHA) and push again please @MatMoore?

It's a known GitHub Actions issue when changing the target branch after pushing

MatMoore:
@colinrotherham thanks, I have pulled that commit in.

Regarding the fixtures, I didn't think it was worth adding any new ones just for this, but for a few examples where the `label` param was specified, I've added in the `id`. This will help with testing that the jinja version is consistent.

packages/nhsuk-frontend/src/nhsuk/components/file-upload/fixtures.mjs
```diff
@@ -8,6 +8,7 @@ export const examples = {
     context: {
       label: {
         text: 'Upload a file',
+        id: 'file-upload-label',
```
colinrotherham: Thanks for all this @MatMoore

We might one day use the `'default'` examples as our exemplar best showcase, especially as the first component example on the design system pages

So could we remove the `id` here as it's unnecessary?

Saves it being accidentally copied and pasted

Could I suggest we find the label variants, and add a new one? They're present in lots of the examples to quickly output different sizes, but something like:

```patch
      {
        description: 'with size XL',
        context: {
          label: {
            size: 'xl'
          }
        }
-     }
+     },
+     {
+       description: 'with id attribute',
+       context: {
+         label: {
+           id: 'custom-id'
+         }
+       }
+     }
```

packages/nhsuk-frontend/src/nhsuk/components/file-upload/fixtures.mjs
```diff
@@ -107,6 +107,14 @@ export const examples = {
             size: 'xl'
           }
         }
+      },
+      {
+        description: 'with id attribute',
```
colinrotherham: We'd best explain that this is the label id

Are we setting the `hidden: true` option here as well?

```suggestion
        description: 'with label id attribute',
```

packages/nhsuk-frontend/src/nhsuk/components/input/fixtures.mjs
```diff
@@ -248,6 +248,17 @@ export const examples = {
             size: 'xl'
           }
         }
+      },
+      {
+        description: 'with id attribute',
```
colinrotherham:
```suggestion
        description: 'with label id attribute',
```

packages/nhsuk-frontend/src/nhsuk/components/select/fixtures.mjs
```diff
@@ -347,6 +347,17 @@ export const examples = {
             size: 'xl'
           }
         }
+      },
+      {
+        description: 'with id attribute',
```
colinrotherham:
```suggestion
        description: 'with label id attribute',
```

packages/nhsuk-frontend/src/nhsuk/components/textarea/fixtures.mjs
```diff
@@ -22,6 +22,7 @@ export const examples = {
     context: {
       label: {
         text: 'Can you provide more detail?',
+        id: 'with-hint-label',
```
colinrotherham: We can remove this now we've split out a separate fixture

```suggestion
```

packages/nhsuk-frontend/src/nhsuk/components/textarea/fixtures.mjs
```diff
@@ -76,6 +77,17 @@ export const examples = {
             size: 'xl'
           }
         }
+      },
+      {
+        description: 'with id attribute',
```
colinrotherham:
```suggestion
        description: 'with label id attribute',
```

colinrotherham (approved):
Nicely done, thanks @MatMoore

PR #1801, opened 2026-02-10
[v10] Add Nunjucks options for components with modifier classes

colinrotherham:
## Description

This PR adds support for Nunjucks component `variant` and other modifiers to close https://github.com/nhsuk/nhsuk-frontend/issues/1852

## Checklist

- [x] CHANGELOG entry

### DIFF
--- a/packages/nhsuk-frontend-review/src/examples/code-blocks.njk
+++ b/packages/nhsuk-frontend-review/src/examples/code-blocks.njk
@@ -1,7 +1,6 @@
-{% extends "layouts/page.njk" %}
+{% extends "layouts/page-full-width.njk" %}
 
 {% set pageName = "Code blocks" %}
-{% set mainClasses = "nhsuk-main-wrapper--s" %}
 
 {% block beforeContent %}
   {{ breadcrumb({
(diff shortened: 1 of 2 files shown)


anandamaryon1:
This is testing our gov alignment…

It'd be a bit messier but we could perhaps defer to gov where the component is shared and only add flags like this for our own things, eg. small buttons. 

Or, if we're diverging here, we already have the `size` in use for headings etc. so we could go whole hog and use that for all size options.

colinrotherham:
I prefer `size: "s"` but `small: true` aligns with NHS.UK React components

We support `secondary: true` on cards, yet only the `--secondary` class modifier on buttons

packages/nhsuk-frontend/src/nhsuk/components/button/macro-options.mjs
```diff
@@ -59,6 +59,11 @@ export const params = {
     description:
       'The URL that the button should link to. If this is set, `element` will be automatically set to `"a"` if it has not already been defined.'
   },
+  small: {
+    type: 'boolean',
+    required: false,
+    description: 'If set to `true`, smaller button size will be used.'
+  },
```
colinrotherham, line 83: If this was `size` instead of `small` here's an alternative for guidance

Effectively `"s"` would work but `"m"` (default) would be intentionally ignored

```mjs
size: {
  type: 'string',
  required: false,
  description: 'Size of the button – `"s"` or `"m"`. Defaults to `"m"`'
},
```

---

This is a bit like `.nhsuk-body` and `.nhsuk-body-m` being aliases

frankieroberto: I think I prefer `small: true` here. It's inconsistent with `size: "s"` however it’s consistent with the terms used in the actual CSS classes.

And I think on balance it’s better to align to CSS, to make the param names easier to remember if you already know the CSS modifier. Also may help the designer-developer handover if the terms are aligned?

frankieroberto:
Alternative idea: `variant`.

Eg, `variant: "secondary"`, `variant: "secondary-solid"`, `variant: "warning"`, etc

One possible advantage is that it helps to imply that the variants are mutually exclusive, eg you can’t have a button that is both `secondary` and `warning`.

This wouldn't work for `small` though, as you can have small secondary buttons... 🤔

colinrotherham:
Or `--small` and `--inline` radios?

Bit like some of the card combos with `clickable` too

packages/nhsuk-frontend/src/nhsuk/components/button/macro-options.mjs
colinrotherham: Good point, I like this thinking

Can I ask how you'd configure a hypothetical large search input?

Large search input

<br>
<br>

Bearing in mind:

* Text input is large
* Primary button default size

Would you use Nunjucks to pass `large: true` in `params`, only `params.input`, or something else?

frankieroberto: Is it a cop out to suggest that that could be a separate `search` component (which includes both the input and the button)?

colinrotherham: Yeah, this PR for `small: true` was formed to solve that question regarding:

* https://github.com/nhsuk/nhsuk-frontend/pull/1660

What do you think?

frankieroberto:
@colinrotherham yeah, `variant` would have to be only for mutually-exclusive modifiers. Not sure how many of those we have!

Potentially `variant: "inline", small: true` would work? (or `variant: "inline", size: "small")`.

edwardhorsford:
> It'd be a bit messier but we could perhaps defer to gov where the component is shared and only add flags like this for our own things, eg. small buttons.

We're in a tricky case where the current NHS default sizings are often way bigger - and can look stupidly spacious in services. And probably a greater percentage of our stuff is (or could be?) business facing - so I see more of a need for compact variants of many things.

A separate way of tackling it could be some kind of global modifier or config - generally setting the design system in to compact mode. That's probably not incompatible with individual components supporting a size though.

packages/nhsuk-frontend/src/nhsuk/components/button/macro-options.mjs
colinrotherham: Any thoughts on this @frankieroberto?

If we're happy with `small: true` do you reckon we could use `large: true` in https://github.com/nhsuk/nhsuk-frontend/pull/1660?

anandamaryon1: I vote for `small:true` and `large: true`. And `size` being reserved for things that have a scale of sizes.

colinrotherham: Thanks @anandamaryon1, let's call this ready for review then

anandamaryon1:
Happy with this but think it needs to be v11 now, given it's a breaking change for copy and pasting Nunjucks.

Unless we implement the [version number designs](https://github.com/nhsuk/nhsuk-service-manual/pull/2432) at the same time? Would be a good test for it I suppose.

colinrotherham:
> Happy with this but think it needs to be v11 now, given it's a breaking change for copy and pasting Nunjucks.

It's a non-breaking change because you don't have to use `small: true`

You can continue using `classes: "nhsuk-radios--small"`

No rush to update the design system examples

anandamaryon1:
Ah yes makes sense, added a ticket for myself to update design system examples only in v11.

packages/nhsuk-frontend-review/src/examples/components/index.njk
```diff
@@ -589,6 +589,7 @@
       }) }}
 
       {{ summaryList({
+        lastRowBorder: false,
```
frankieroberto, line 592:
```suggestion
        lastRowHasBorder: false,
```

Maybe? 🤔

packages/nhsuk-frontend/src/nhsuk/components/back-link/macro-options.mjs
```diff
@@ -28,6 +28,11 @@ const options = {
     required: false,
     description: 'The value of the link `href` attribute.'
   },
+  variant: {
+    type: 'string',
+    required: false,
+    description: 'Optional variant of back link to render – `"reverse"`.'
```
frankieroberto:
```suggestion
    description: 'Optional variant of back link – currently only `"reverse"` is supported.'
```

packages/nhsuk-frontend/src/nhsuk/components/action-link/macro-options.mjs
```diff
@@ -33,6 +33,11 @@ const options = {
     required: false,
     description: 'If set to `true`, then the link will open in a new window.'
   },
+  variant: {
+    type: 'string',
+    required: false,
+    description: 'Optional variant of action link to render – `"reverse"`.'
```
frankieroberto: Avoid use of `render` as that's quite technical?

```suggestion
    description: 'Optional variant of action link – currently only `"reverse"` is supported.'
```

packages/nhsuk-frontend/src/nhsuk/components/card/template.njk
```diff
@@ -25,8 +28,32 @@
...
+    not params.classes or (
+      not "nhsuk-card--feature" in params.classes and
+      not "nhsuk-card--primary" in params.classes and
+      not "nhsuk-card--secondary" in params.classes and
+      not "nhsuk-card--warning" in params.classes
+    )
+  )
+%}
+  {% set classNames = classNames ~ " nhsuk-card--" ~ variant %}
+{% endif %}
```
frankieroberto, line 42: Worth abstracting this logic to some kind of helper filter, as it’s used in a few places?

```suggestion
{% set supportedVariants = ["feature", "primary", "secondary", "warning"] %}
{%- if (variant in supportedVariants) and not (params.classes | includesVariant(variant)) %}
  {% set classNames = classNames ~ " nhsuk-card--" ~ variant %}
{% endif %}
```

packages/nhsuk-frontend/src/nhsuk/components/do-dont-list/template.njk
```diff
@@ -1,33 +1,38 @@
-{% from "nhsuk/macros/attributes.njk" import nhsukAttributes %}
+{% from "nhsuk/components/card/macro.njk" import card %}
 {% from "nhsuk/macros/icon.njk" import nhsukIcon %}
 
-{%- set type = "cross" if params.type == "cross" else "tick" %}
-{%- set headingLevel = params.headingLevel if params.headingLevel else 3 %}
+{%- set icon = "cross" if (params.icon or params.type) == "cross" else "tick" %}
```
frankieroberto, line 5: Add a note here to suggest that `params.type` is deprecated?

packages/nhsuk-frontend/src/nhsuk/components/task-list/macro-options.mjs
```diff
@@ -107,6 +107,12 @@ const options = {
             type: 'string',
             required: false,
             description: 'Classes to add to the status container.'
+          },
+          variant: {
+            type: 'string',
+            required: false,
+            description:
+              'Optional variant of status to render – `"completed"` or `"cannot-start-yet"`.'
```
frankieroberto: Not totally sure about this being a 'variant' - it's quite different from the other examples.

`status.state` maybe? Or is that too confusing?

packages/nhsuk-frontend-review/src/examples/components/index.njk
colinrotherham: Ah this came up previously and `firstCellIsHeader` was flagged too

Should we review that one?

---

**Update:** Ahh I'd forgotten [I suggested `lastRowHasBorder` on Slack](https://screening-discovery.slack.com/archives/C08QG748NDC/p1770819199021189?thread_ts=1770818944.166379&cid=C08QG748NDC) but @edwardhorsford preferred `lastRowBorder` and we already approved it in https://github.com/nhsuk/nhsuk-frontend/pull/1803#top

>I prefer `lastRowBorder` over `lastRowHasBorder`. but could also have remove in there since it’s overriding a default

Let's consider "is" and "has" prefixes unnecessary like we discussed in:

* https://github.com/nhsuk/nhsuk-frontend/issues/1852#issuecomment-4056666261

packages/nhsuk-frontend/src/nhsuk/components/task-list/macro-options.mjs
colinrotherham: Yeah I was torn too. If we're not sure for now, can leave it as classes?

packages/nhsuk-frontend/src/nhsuk/components/card/template.njk
colinrotherham: Unfortunately we can't add new mandatory filters like `includesVariant()`

It would be a breaking change to make sure all users ran `env.addFilter()` to add it

We could add a macro similar to `nhsukAttributes()` to write out classes and toggle variants though? Sounds like a nice enhancement for another PR in future?

packages/nhsuk-frontend/src/nhsuk/components/task-list/macro-options.mjs
colinrotherham: I've put it back to `classes` and we can have a think in another PR

Maybe we can make the classes optional and solve this by:

* Adding styles for `:not(:has(.nhsuk-tag))` to set the font size automatically
* Adding Nunjucks option `colour: "secondary"` for the text colour

packages/nhsuk-frontend/src/nhsuk/components/do-dont-list/template.njk
colinrotherham: Done in https://github.com/nhsuk/nhsuk-frontend/pull/1801/commits/5a3fb43644d7bacde2762afc3d9e948c4f38a8e9

packages/nhsuk-frontend/src/nhsuk/components/action-link/macro-options.mjs
colinrotherham: Done in https://github.com/nhsuk/nhsuk-frontend/pull/1801/commits/cbd9348f806b810425d5cf2c7090b141dd732a5f

packages/nhsuk-frontend/src/nhsuk/components/back-link/macro-options.mjs
colinrotherham: Done in https://github.com/nhsuk/nhsuk-frontend/pull/1801/commits/cbd9348f806b810425d5cf2c7090b141dd732a5f

CHANGELOG.md
```diff
@@ -8,9 +8,109 @@ Note: This release was created from the `support/10.x` branch.
 
 #### Add a modifier class for header inline search or account links
 
-We've added a new `.nhsuk-header--inline` class for the [header](https://service-manual.nhs.uk/design-system/components/header) component. This positions the search bar (or account links) inline with the NHS logo, depending on the length of your service name.
+We've added a new `.nhsuk-header--inline` class and `inline` Nunjucks option for the [header](https://service-manual.nhs.uk/design-system/components/header) component. This positions the search bar (or account links) inline with the NHS logo, depending on the length of your service name. For example:
```
anandamaryon1:
```suggestion
We've added a new `.nhsuk-header--inline` class and `inline` Nunjucks option for the [header](https://service-manual.nhs.uk/design-system/components/header) component. This positions the search bar (or account links) inline with the NHS logo on small screens, depending on the length of your service name. For example:
```

anandamaryon1 (approved):

colinrotherham:
@frankieroberto I've added a [`"reverse"`](https://nhsuk-frontend-pr-1801.herokuapp.com/nhsuk-frontend/examples/code-blocks/) and `border: false` option to code blocks as I forgot [you use these on the NHS prototype kit site](https://prototype-kit.service-manual.nhs.uk/guides/git/#1-check-if-you-have-git-installed). They look really smart 😎

~You reckon code blocks should be font size 19 not 16?~

**Update:** [I've matched the font size](https://nhsuk-frontend-pr-1801.herokuapp.com/nhsuk-frontend/examples/code-blocks/)

packages/nhsuk-frontend/src/nhsuk/components/task-list/macro-options.mjs
frankieroberto: Yea ok let’s leave this for a future PR

PR #1803, opened 2026-02-11
[v10] Add summary list `border: false` and `lastRowBorder: false` Nunjucks options

colinrotherham:
## Description

This PR proposes the following new Nunjucks options for summary lists:

1. Set `border: false` to remove all separating borders (also works on specific rows)
1. Set `lastRowBorder: false` to remove the last row separating border automatically

Why are these needed?

When row data is provided by external sources, it's incredibly difficult to set classes manually on specific rows

## Checklist

- [ ] CHANGELOG entry

### DIFF
--- a/packages/nhsuk-frontend-review/src/examples/components/index.njk
+++ b/packages/nhsuk-frontend-review/src/examples/components/index.njk
@@ -589,6 +589,7 @@
       }) }}
 
       {{ summaryList({
+        lastRowBorder: false,
         rows: [
           {
             key: {
@@ -642,7 +643,6 @@
             }
           },
           {
-            classes: "nhsuk-summary-list__row--no-border",
             key: {
               text: "Contact details"
             },
@@ -668,7 +668,7 @@
       }) }}
 
       {{ summaryList({
-        classes: "nhsuk-summary-list--no-border",
+        border: false,
         rows: [
           {
             key: {
--- a/packages/nhsuk-frontend/src/nhsuk/components/card/fixtures.mjs
+++ b/packages/nhsuk-frontend/src/nhsuk/components/card/fixtures.mjs
@@ -68,7 +68,7 @@ const fixtures = {
               }
             },
             {
-              classes: 'nhsuk-summary-list__row--no-border',
+              border: false,
               key: {
                 text: 'Date of birth'
               },
@@ -127,7 +127,7 @@ const fixtures = {
               }
             },
             {
-              classes: 'nhsuk-summary-list__row--no-border',
+              border: false,
               key: {
                 text: 'Date of birth'
               },
@@ -213,7 +213,7 @@ const fixtures = {
               }
             },
             {
-              classes: 'nhsuk-summary-list__row--no-border',
+              border: false,
               key: {
                 text: 'Date of birth'
               },
@@ -256,7 +256,7 @@ const fixtures = {
               }
             },
             {
-              classes: 'nhsuk-summary-list__row--no-border',
+              border: false,
               key: {
                 text: 'Date of birth'
               },
@@ -296,7 +296,7 @@ const fixtures = {
               }
             },
             {
-              classes: 'nhsuk-summary-list__row--no-border',
+              border: false,
               key: {
                 text: 'Date of birth'
               },
@@ -328,7 +328,7 @@ const fixtures = {
               }
             },
             {
-              classes: 'nhsuk-summary-list__row--no-border',
+              border: false,
               key: {
                 text: 'Date of birth'
               },
@@ -390,7 +390,7 @@ const fixtures = {
               }
             },
             {
-              classes: 'nhsuk-summary-list__row--no-border',
+              border: false,
               key: {
                 text: 'Date of birth'
               },
@@ -447,7 +447,7 @@ const fixtures = {
               }
             },
             {
-              classes: 'nhsuk-summary-list__row--no-border',
+              border: false,
               key: {
                 text: 'Date of birth'
               },
@@ -529,7 +529,7 @@ const fixtures = {
               }
             },
             {
-              classes: 'nhsuk-summary-list__row--no-border',
+              border: false,
               key: {
                 text: 'Date of birth'
               },
@@ -573,7 +573,7 @@ const fixtures = {
               }
             },
             {
-              classes: 'nhsuk-summary-list__row--no-border',
+              border: false,
               key: {
                 text: 'Date of birth'
               },
@@ -614,7 +614,7 @@ const fixtures = {
               }
             },
             {
-              classes: 'nhsuk-summary-list__row--no-border',
+              border: false,
               key: {
                 text: 'Date of birth'
               },
@@ -647,7 +647,7 @@ const fixtures = {
               }
             },
             {
-              classes: 'nhsuk-summary-list__row--no-border',
+              border: false,
               key: {
                 text: 'Date of birth'
               },
@@ -848,7 +848,7 @@ const fixtures = {
               }
             },
             {
-              classes: 'nhsuk-summary-list__row--no-border',
+              border: false,
               key: {
                 text: 'Date of birth'
               },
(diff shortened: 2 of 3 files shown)


packages/nhsuk-frontend/src/nhsuk/components/summary-list/template.njk
```diff
@@ -14,18 +25,34 @@
...
+{%- if params.lastRowBorder === false and rows | length %}
+  {%- set index = rows | length - 1 %}
+  {%- set rows = (rows.splice(index, 1, {
+    classes: rows[index].classes,
+    border: false,
+    key: rows[index].key,
+    value: rows[index].value,
+    actions: rows[index].actions
+  }), rows) -%}
+{% endif %}
```
colinrotherham: @MatMoore From [your old issue](https://github.com/MatMoore/nhsuk-frontend-jinja/issues/2) we flagged that [Nunjucks does not support modifying an object](https://github.com/mozilla/nunjucks/issues/313)

But is there a Jinja friendly alternative to this `.splice()` workaround?

We're replacing the last item in the `rows` array to set `border: false`

```jinja
{% do rows.pop() %}
{% do rows.append({ … }) %}
```

MatMoore: I can't think of a direct equivalent no. Normally in python you would use slice syntax for this but I don't think Jinja supports that.

Would it work if you change the `row.border === false` check to something like `loop.last and params.lastRowBorder === false`?

This isn't always a reliable way of doing things though - I remember running into a bug previously where first/last of the loop were not actually the first/last displayed, due to filtering out empty objects or something.

colinrotherham: >This isn't always a reliable way of doing things though - I remember running into a bug previously where first/last of the loop were not actually the first/last displayed, due to filtering out empty objects or something.

Yep! You guessed the main reason why I pulled `set rows` up though, because the last item could be `null`, `false`, `undefined` etc so we can’t rely on indexes unless that `select("mapping")` happens up front

```njk
{%- set rows = params.rows | select("mapping") if params.rows | length else [] %}
```

>Would it work if you change the `row.border === false` check to something like `loop.last and params.lastRowBorder === false`?

Yeah that would work if that [`{% do %}` expression](https://jinja.palletsprojects.com/en/stable/templates/#expression-statement) I saw suggested is a no go?

I'd prefer to have `rows` prepared and ready to pass to another macro in future (where `params.lastRowBorder` might not be available) but we can make your suggestion if it works in both Nunjucks and Jinja

MatMoore: The do expression works, but it's an extension rather than core jinja, and we don't currently use it in nhsuk_frontend_jinja. I'd be open to doing it that way, but we'd need to make it a breaking change and update the recommended configuration in the documentation.

colinrotherham: Gotcha, let's keep Jinja happy then. I've [pushed up a `loop.last` version](https://github.com/nhsuk/nhsuk-frontend/compare/f8262e1a8234aa97955063dfc77756698e52fc80..35cfe1c9002ed369d84c855d90ae6279da9cd5c5)

edwardhorsford:
Really glad to see this!

I recently added a filter to my service to achieve a similar thing - as we often have summary lists within wrapping containers and the last row border doesn't look correct. Like Colin mentioned - our summary list rows are often dynamically generated so it's not so easy to apply the existing `no-border` class to the right item.

For the first option, I wonder if it should be `borders: false`?

colinrotherham:
Thanks @edwardhorsford, I've brought it out of draft

> For the first option, I wonder if it should be `borders: false`?

I think singular `border: false` matches the modifier `---no-border` better but it's a good point

edwardhorsford:
Does it need to match?

I'm thinking `borders` might align to our other properties which are pluralised - items, values, etc. 

You have the items arrays, but the actual class of each thing is singular...

colinrotherham:
>Does it need to match?

No it doesn't need to match, but [it was some recent feedback](https://github.com/nhsuk/nhsuk-frontend/pull/1801#discussion_r2790168240) from @frankieroberto

For example:

* If you already know `--small` you'd assume `small: true`
* If you already know `--no-border` you'd (maybe) assume `border: false`

Both ideas make sense and maybe the modifier should be ~`--no-border`~ `--no-borders`

edwardhorsford:
I don't _think_ I agree with the reasoning. Small isn't singular or plural. The existing class makes sense as it applies to a single thing - but this option is applying to a group. @sarawilcox @anandamaryon1 @frankieroberto any views?

Do we have other options that are named singularly but actually are disabling a plurality of things?

colinrotherham:
>Both ideas make sense and maybe the modifier should be ~`--no-border`~ `--no-borders`

Did you see this bit @edwardhorsford?

edwardhorsford:
> > Both ideas make sense and maybe the modifier should be ~`--no-border`~ `--no-borders`
> 
> Did you see this bit @edwardhorsford?

If it's a single class applied to the component, that sounds good? if it's per item, probably not?

colinrotherham:
The only other singular one is `.nhsuk-list--border` where multiple borders are added

See the **A to Z page** pattern
https://service-manual.nhs.uk/design-example/patterns/a-to-z-page/default

Possibly also `--inline` for radios, where in a non-CSS sense all radios are individually inline

colinrotherham:
Can we unblock this one with a decision @anandamaryon1?

anandamaryon1:
@colinrotherham are you saying if we want to change this to `borders: true` we should also change the CSS class name to match? 

This makes sense to me. Technically borders plural is more accurate. Can't really think why it was singular to begin with.

Ok to make this change and mark `--no-border` as deprecated @colinrotherham? (If I've understood correctly)

colinrotherham:
@anandamaryon1 I see the case for plural but I'm also happy with singular

We follow a "trickle down" approach when naming, so a singular top-level flag applies to all the things

* Adding `.nhsuk-radios--inline` means "Make all the radios inline"
* Adding `.nhsuk-radios--small` means "Make all the radios small"

Similarly with lists:

* Adding `.nhsuk-list--border` means "Make all the list items have borders"

But yes, Nunjucks options (e.g. `border: true`) should match the modifier `--border`

## Singular or plural?

To help us decide, we only have one use of plural currently:

```css
.nhsuk-summary-list__row--no-actions
```

But when it comes to borders we have two singular uses:

```scss
.nhsuk-list--border {}
.nhsuk-summary-list--no-border {}
```

So if we choose to switch to plural flags, we'd need to deprecate the singular and temporarily support both:

```scss
.nhsuk-list--borders,
// Deprecated, to be removed in v11.0
.nhsuk-list--border {}
```

```scss
.nhsuk-summary-list--no-borders,
// Deprecated, to be removed in v11.0
.nhsuk-summary-list--no-border {}
```

I'm happy with this too, that alright?

---

**Update:** We also have the Nunjucks `firstCellIsHeader: true` top-level table option as singular

It would be a bit awkward in plural as `firstCellsAreHeader: true`

anandamaryon1:
Hmm ok, the trickle-down naming makes sense now you've explained it…

All this extra refactoring seems over the top. So, let's just go with the Nunjucks option being `border: false` and the class stays `--no-border`. In the context of the component, its appearance and options, I don't think this will confuse. 

So it's a border list or a no-border list, rather than a borders list or a no-borders list. 

Hope you're OK with this @edwardhorsford

edwardhorsford:
It's not a hill I'm going to die on, but it feels a bit backwards to me that our public api wouldn't use a plural form for an option like this. I don't particularly think it _needs_ to match the classname. If anything it's the classname that's a bit weird given it applies to multiple items, not one.

colinrotherham:
Thanks @edwardhorsford

>…If anything it's the classname that's a bit weird given it applies to multiple items, not one.

Let's merge this and discuss plurals in future

Renaming `.nhsuk-list--border` to `.nhsuk-list--borders` would also be a bit weird, yet consistent

colinrotherham:
I'll merge this into https://github.com/nhsuk/nhsuk-frontend/pull/1801 and we can add the changelog entry there

PR #1804, opened 2026-02-11
[v10] Update card states for consistency

anandamaryon1:
## Description

Adds a hover state which alters the card border colour on clickable cards and adds an active state for a 'pressed' style. 

Pressed styling is maybe too much? And can be reserved for buttons only. Added because the styling seems to suggest it should be so. 

Also alters primary card with chevron's icon style, to inherit hover and active colours. 

### Before
| Hover    | Active |
| -------- | ------- |
|  nhsuk github io_nhsuk-frontend_examples_card-group_ (7) |  nhsuk github io_nhsuk-frontend_examples_card-group_ (6) |

### After
| Hover    | Active |
| -------- | ------- |
|  localhost_3000_nhsuk-frontend_examples_card-group_ (5)  |  localhost_3000_nhsuk-frontend_examples_card-group_ (3) |

## Checklist

- [ ] CHANGELOG entry

### DIFF
--- a/packages/nhsuk-frontend/src/nhsuk/components/card/_index.scss
+++ b/packages/nhsuk-frontend/src/nhsuk/components/card/_index.scss
@@ -197,11 +197,17 @@ $card-border-hover-colour: $nhsuk-card-border-hover-colour;
     margin-bottom: 0;
   }
 
+  // Card shadow shared styles
+
+  .nhsuk-card--clickable,
+  .nhsuk-card--secondary {
+    box-shadow: 0 $nhsuk-card-border-bottom-width 0 $nhsuk-border-colour;
+    @include nhsuk-responsive-margin(7, "bottom", $adjustment: $nhsuk-card-border-bottom-width); // [18]
+  }
+
   // Clickable card
 
   .nhsuk-card--clickable {
-    box-shadow: 0 $nhsuk-card-border-bottom-width 0 $nhsuk-border-colour;
-
     .nhsuk-card__heading a,
     .nhsuk-card__link {
       &::before {
@@ -279,13 +285,28 @@ $card-border-hover-colour: $nhsuk-card-border-hover-colour;
       margin-bottom: nhsuk-spacing(5);
     }
 
+    .nhsuk-card--clickable,
+    .nhsuk-card--secondary {
+      margin-bottom: nhsuk-spacing(5) + $nhsuk-card-border-bottom-width; // [18]
+    }
+
     @include nhsuk-media-query($until: desktop) {
       .nhsuk-card {
         margin-bottom: nhsuk-spacing(3);
       }
 
+      .nhsuk-card--clickable,
+      .nhsuk-card--secondary {
+        margin-bottom: nhsuk-spacing(3) + $nhsuk-card-border-bottom-width; // [18]
+      }
+
       &:last-child .nhsuk-card {
         margin-bottom: 0;
+      }
+
+      &:last-child .nhsuk-card--clickable,
+      &:last-child .nhsuk-card--secondary {
+        margin-bottom: $nhsuk-card-border-bottom-width; // [18]
       }
     }
   }
@@ -542,7 +563,6 @@ $card-border-hover-colour: $nhsuk-card-border-hover-colour;
     border-right: 0;
     border-left: 0;
     background: transparent;
-    box-shadow: 0 $nhsuk-card-border-bottom-width 0 $nhsuk-border-colour;
   }
 
   .nhsuk-card--secondary > .nhsuk-card__heading-container,


colinrotherham:
Changelog entry and update the screenshots and we can get this in?

Would be nice to see the changes side-by-side

anandamaryon1:
Added screens to description. 

Fixed a bug with the active area extending below the card (because of box sizing I think). 

Added a changelog entry.

anandamaryon1:
Noticed that now I've added `border-box` to the card content, the card shadow is not clickable. Probably the wrong fix. 

One you might be able to investigate @colinrotherham? It's a bug on the current v11 branch. Click in the space under one of the 'halves' cards: https://nhsuk.github.io/nhsuk-frontend/examples/card-group/. Maybe should be in a separate issue / PR?

packages/nhsuk-frontend/src/nhsuk/components/card/_index.scss
colinrotherham, line 535: Adding `box-sizing: border-box;` mostly fixes the issue by the way 🙌

Just need another 4px adding on to the height

```patch
-     height: 100%;
+     height: calc(100% + $nhsuk-card-shadow-size);
```

anandamaryon1: Hmm, that kinda fixes the click styling but doesn't actually click the link…

Think I may have done this wrong to begin with and the states should be tied to the actual anchor link, but not sure how to do that since it's nested in the card. 

Looking again, it's a live bug with the expander too, you can't click the shadow…

colinrotherham: Ah you're spot on, I forgot the link has a `::before` to spread the clickable area out

So you'll want to ignore https://github.com/nhsuk/nhsuk-frontend/pull/1804#discussion_r2879463191 and do this instead:

```patch
  position: absolute;
  top: 0;
  right: 0;
- bottom: 0;
+ bottom: -$nhsuk-card-border-bottom-width;
  left: 0;
```

colinrotherham: Regarding states, yeah if you manually set `.nhsuk-card__link:active` nothing happens

Same thing on sortable tables where `th:hover` should really be `th:has(button:hover)`

Adding both would be nice for browsers without [`:has()` support](https://caniuse.com/css-has), like this?

```patch
- &:active {
+ &:active,
+ &:has(.nhsuk-card__heading a:active),
+ &:has(.nhsuk-card__link:active) {
    bottom: -$nhsuk-card-border-width; // [3]
    border-color: $nhsuk-card-border-hover-colour;
  }
```

anandamaryon1: Fixed the target area for clickable cards.

I'll now do the same for the expander

colinrotherham: Ahh brill we're doing the border offset too

Worth a reminder then about https://github.com/nhsuk/nhsuk-frontend/pull/1737 where @paulrobertlloyd has deducted the border width from the padding. Both are great to see

colinrotherham: Are you alright adding the `:has()` change too?

It's likely important for touch screens (e.g. simulated hover/active via stylus) where the event might go straight to the link rather than the card

I can demo it if you're not sure

anandamaryon1: Hmmph, expander is quite different, might leave that as a separate issue.

anandamaryon1: The `has:` for clickable card, is that necessary? Since the states are tied to the actual `<a>`?

colinrotherham: States are not on the `<a>` that's the problem

>Regarding states, yeah if you manually set `.nhsuk-card__link:active` nothing happens
>
> Same thing on sortable tables where `th:hover` should really be `th:has(button:hover)`

Want a demo?

anandamaryon1: Ah got you, thanks. Added the `:has()` selectors

packages/nhsuk-frontend/src/nhsuk/components/card/_index.scss
```diff
@@ -28,12 +28,13 @@
...
 /// 16. Emergency care card <a> needs to be white because the background colour is black.
 /// 17. Trick IE10 into rendering the invisible clickable area. Without this it does not.
+/// 18. Increase the clickable area to the border of the card and the shadow.
 ///
 /// @group components/card
 ////
 
 $nhsuk-card-border-width: 1px;
-$nhsuk-card-border-bottom-width: nhsuk-spacing(1);
+$nhsuk-card-shadow-size: nhsuk-spacing(1) - 1px;
```
colinrotherham: In v10 we didn't rename these Sass variables without keeping the old one deprecated

Could do with a follow up PR to (perhaps) add underscores to make them private in v11

```patch
- $nhsuk-card-border-bottom-width: nhsuk-spacing(1);
+ $_nhsuk-card-border-bottom-width: nhsuk-spacing(1);
```

anandamaryon1: Ahh yes, whoops. 

Can't we do it in this PR?

colinrotherham: With a changelog entry to tell people to stop using `$nhsuk-card-border-bottom-width`?

colinrotherham:
@anandamaryon1 If this is going into v10.4.0 we'll need two PRs:

1. Fix for the `box-sizing` issue in v11
2. Fix for the shadow issue in v10 + v11

anandamaryon1:
> @anandamaryon1 If this is going into v10.4.0 we'll need two PRs:
> 
> 1. Fix for the `box-sizing` issue in v11
> 2. Fix for the shadow issue in v10 + v11

@colinrotherham 
Added a PR to fix `box-sizing` issue in v11: https://github.com/nhsuk/nhsuk-frontend/pull/1847

colinrotherham:
@anandamaryon1 Switching from `border-bottom` to `box-shadow` has removed 3px card height

I've pushed up a [similar fix used for buttons](https://github.com/nhsuk/nhsuk-frontend/pull/1804/commits/153a1814321abf1ecf58d4f09593f541ea075ee2) to adjust the margin to compensate:

```scss
@include nhsuk-responsive-margin(7, "bottom", $adjustment: $nhsuk-card-border-bottom-width); // [18]
```

colinrotherham (approved):

PR #1829, opened 2026-02-25
[v10] Automatically validate component config schema

colinrotherham:
## Description

This PR fixes an issue where `Component.schema` validation rules are not automatically checked

We can now move these lines from **character-count.mjs** to **configurable-component.mjs**

```mjs
// Check for valid config
const errors = validateConfig(CharacterCount.schema, this.config)
if (errors[0]) {
  throw new ConfigError(formatErrorMessage(CharacterCount, errors[0]))
}
```

For components with mandatory options I've ported https://github.com/alphagov/govuk-frontend/pull/4791

Code coverage for `validateConfig()` using both `'anyOf'` and `'allOf'` has been added

## Checklist

- [x] CHANGELOG entry

### DIFF
--- a/packages/nhsuk-frontend/src/nhsuk/common/configuration/validate-config.mjs
+++ b/packages/nhsuk-frontend/src/nhsuk/common/configuration/validate-config.mjs
@@ -12,11 +12,11 @@
  * @returns {string[]} List of validation errors
  */
 export function validateConfig(schema, config) {
-  const validationErrors = []
+  const validationErrors = /** @type {string[]} */ ([])
 
   // Check errors for each schema
   for (const [name, conditions] of Object.entries(schema)) {
-    const errors = []
+    const errors = /** @type {string[]} */ ([])
 
     // Check errors for each schema condition
     if (Array.isArray(conditions)) {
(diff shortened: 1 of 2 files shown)


packages/nhsuk-frontend/src/nhsuk/lib/fixtures/configuration/mock-component.mjs
```diff
@@ -156,6 +156,49 @@ export class MockConfigurableComponentNoDefaults extends ConfigurableComponent {
...
+/**
+ * @augments {ConfigurableComponent<MockConfigOptional>}
+ */
+export class MockConfigurableComponentAnyOf extends ConfigurableComponent {
+  static moduleName = 'mock-component'
+
+  /**
+   * @satisfies {Schema<MockConfigOptional>}
+   */
+  static schema = {
```
MatMoore, line 211: if this is just used in the one test file, I think it would be a bit more readable to inline this schema into that test. I don't know if we need the whole mock component?

colinrotherham: Fine by me, makes sense

I should probably add a mock component with both `anyOf` and `allOf` combined too

Thanks Mat

colinrotherham: Moved the mock schema back 👍

colinrotherham:
@MatMoore I've added some mixed examples so we can see "and" with "or" together

* `'"example1" and "example2" must be provided'`
* `'"example3" or "example4" must be provided'`

PR #1832, opened 2026-02-26
Add support for server-side table sorting

frankieroberto:
This updates the sortable tables PR (#1654) with support for server-side table sorting.

This does not use AJAX to fetch the new rows and update the table (as we said that was out of scope for now), but instead just inserts plain `<a href="">` links within the headings, which would then do a full page reload, like using query strings to specify the new sort order.

The main use case for this would be where there are more rows than would reasonably fit on the screen, and so the table is only the first X rows, with pagination links. In this scenario, client-side sorting would be misleading as the first X rows are likely to be different depending on the sort order.

Within the Nunjucks macro, server-side sorting could be added by setting an `href` on the header cell, along with an `aria-sort` attribute.

### DIFF
--- a/packages/nhsuk-frontend/src/nhsuk/components/tables/table.mjs
+++ b/packages/nhsuk-frontend/src/nhsuk/components/tables/table.mjs
@@ -60,12 +60,11 @@ export class Table extends ConfigurableComponent {
     // Count headings which have an aria-sort attribute but which
     // do not already have a link inside them (for server-side
     // sorting)
-    const sortableHeadingsCount = this.$headings.filter(
-      (heading) =>
-        heading.getAttribute('aria-sort') && !heading.querySelector('a')
-    ).length
+    const hasSortableHeadings = this.$headings.some((heading) => {
+      return heading.getAttribute('aria-sort') && !heading.querySelector('a')
+    })
 
-    if (sortableHeadingsCount > 0) {
+    if (hasSortableHeadings) {
       this.$root.classList.add('nhsuk-table--with-sortable-columns')
       this.createHeadingButtons()
       this.updateCaption()


packages/nhsuk-frontend/src/nhsuk/components/tables/fixtures.mjs
```diff
@@ -400,6 +400,85 @@ const fixtures = {
       ]
     }
   },
+  'with names and times, sortable server-side': {
+    context: {
+      caption: 'Appointments',
+      head: [
+        {
+          html: '<a href="#" class="">Time <svg width="22" height="22" focusable="false" aria-hidden="true" role="img" viewBox="0 0 22 22" fill="none" xmlns="http://www.w3.org/2000/svg"><path d="M16.6 13.7L11.6 18.7C11.4 18.9 11.2 19 10.9 19C10.6 19 10.4 18.9 10.2 18.7L5.2 13.7C5.05557 13.5074 4.98545 13.2692 5.00252 13.0291C5.01958 12.789 5.12268 12.5631 5.29289 12.3929C5.4631 12.2227 5.689 12.1196 5.92911 12.1025C6.16922 12.0854 6.40743 12.1556 6.6 12.3L9.9 15.6V4C9.9 3.73478 10.0054 3.48043 10.1929 3.29289C10.3804 3.10536 10.6348 3 10.9 3C11.1652 3 11.4196 3.10536 11.6071 3.29289C11.7946 3.48043 11.9 3.73478 11.9 4V15.6L15.2 12.3C15.286 12.1854 15.3956 12.0906 15.5213 12.022C15.6471 11.9534 15.7862 11.9127 15.9291 11.9025C16.072 11.8924 16.2154 11.913 16.3497 11.9631C16.4839 12.0132 16.6058 12.0916 16.7071 12.1929C16.8084 12.2942 16.8868 12.4161 16.9369 12.5503C16.987 12.6845 17.0076 12.828 16.9975 12.9709C16.9873 13.1138 16.9466 13.2529 16.878 13.3787C16.8094 13.5044 16.7146 13.614 16.6 13.7Z" fill="currentColor"/></svg></a>',
```
frankieroberto: @colinrotherham @anandamaryon1 having to manually insert the right SVG icon is slightly annoying. The Nunjucks macro could do this for somehow.

Alternatively maybe the SVG could be added using CSS as a background image, and then you’d only have to specify a class on the link element. The downside to this is that I think you can no longer use `fill="currentColor"`, which can be useful for high contrast modes?

frankieroberto:
The table could look identical to the client side version.

Imagine this, but with pagination links below it:

Screenshot 2026-02-26 at 13 49 23

packages/nhsuk-frontend/src/nhsuk/components/tables/fixtures.mjs
colinrotherham: Ah nice. Yeah it is a pain having to render all that

Could we take the password input approach and render everything we need in advance?

But with the `hidden` attribute set

frankieroberto: Updated in 98f65abd4671addf3cc2fea2d800efa17ddb1de7 to use the nunjucks macro to insert the SVG.

frankieroberto:
@colinrotherham would you mind helping me bring this back up to date with the base branch, which has been updated?

And are you happy with the approach?

colinrotherham:
@frankieroberto Done

Yeah approach is brilliant, can get it merged in and carry on in https://github.com/nhsuk/nhsuk-frontend/pull/1654 ?

We've got [hard-coded SVGs in **table.mjs**](https://github.com/nhsuk/nhsuk-frontend/blob/sortable-tables-links/packages/nhsuk-frontend/src/nhsuk/components/tables/table.mjs#L11-L28) to resolve at some point

anandamaryon1:
Nice, I'm happy for this to be merged and work to continue on https://github.com/nhsuk/nhsuk-frontend/pull/1654

Will that mean a new preview-branch SHA and updates to anyone using it? 

We'll also need to think about how to document this in the guidance, but that can come later and maybe be left to just the Nunjucks option description.

colinrotherham:
>Will that mean a new preview-branch SHA and updates to anyone using it?

It depends which branch we run `npx publish-preview` from

If we run it from `sortable-table` yeah it recreates branch `preview-sortable-table`

Once merged, we can make sure we create previews from another branch?

frankieroberto:
Thanks both. Let’s merge and continue the work in the other PR

PR #1886, opened 2026-04-13
Add section navigation component

edwardhorsford:
## Description

This introduces a new component 'section navigation' which is a general purpose list of links that might be used within a service. This pattern is commonly used as separate 'tabs' in a section, or filtered views of the same section.

Screenshot 2026-06-01 at 16 20 37

The styles and html are primarily inspired by [Manage vaccinations in schools' code](https://github.com/NHSDigital/manage-vaccinations-in-schools-prototype), with some changes adopting patterns from the new header component which has very similar list of links with active state.

Also supports a reverse option
Screenshot 2026-06-01 at 16 20 53

Vertical option
Screenshot 2026-06-01 at 16 21 07

## Badge
Including a count is a common need for such navigation, which I've added by way of a separate `badge` component. This is really simple currently and just accepts text/html/classes, etc, but in the future could have more styling options.
Screenshot 2026-06-01 at 16 21 56
Screenshot 2026-06-01 at 16 22 05

The api of the badge is like this:
```njk
{{ badge({
  text: "12",
  visuallyHiddenText: "Count:"
}) }}
```

But this can be provided directly to the params of this component. A nicety of doing this rather than providing raw html is we can make sure the badge is spaced correctly and a hidden space added so the nav label text flows correctly with the badge text.

Screenshot 2026-06-01 at 16 24 47

Other details:
* Supports a vertical param to force a vertical layout - useful for side nav
* Progressively enhances with js so that:
* * It'll avoid vertical stacking if there's still space horizontally
* * It'll stack vertically sooner if there's loads of links and they'd otherwise wrap

Not done / not prioritised:
* Support for itemStart / itemEnd - a future component could support arbitrary html for the start and end like our buttons inputs soon will. Not essential for this version - badge covers the common need.
* Alternative handling for many items and smaller screens. The header supports using js to collapse later items in to a menu. We could consider this for a future enhancement.
* Support for disabled items. For casework systems teams may want to show but visually disable an item. This can still be achieved with custom styles and not providing an href.


packages/nhsuk-frontend/src/nhsuk/components/section-navigation/fixtures.mjs
```diff
@@ -0,0 +1,244 @@
...
+  'with html': {
+    context: {
+      items: [
+        {
+          href: '#',
+          html: 'Overview'
+        },
+        {
+          href: '#',
+          html: 'Vaccinations<span class="nhsuk-section-navigation__suffix"><strong class="nhsuk-tag nhsuk-tag--green">New</strong></span>',
```
frankieroberto: Should we support a `suffix` param, given the CSS supports it? A bit like how the Input component already does? 🤔

edwardhorsford: Discussed with @anandamaryon1 that we could later support an `itemEnd` and `itemStart` params, but that we didn't need them in a first version. Potentially this class could be renamed though.

packages/nhsuk-frontend-review/src/examples/section-navigation.njk
```diff
@@ -0,0 +1,268 @@
...
+    <div class="nhsuk-grid-column-full">
+      <h2 class="nhsuk-heading-m">With counts</h2>
+      <p class="nhsuk-body">Use the <code>count</code> option to show a numeric badge alongside each item — for example, the number of results in a filtered view.</p>
+
+      {{ sectionNavigation({
+        items: [
+          {
+            href: "#",
+            text: "All",
+            count: 12
```
frankieroberto: This needs updating to use the badge component.

edwardhorsford: D'oh! thanks! fixed.

paulrobertlloyd:
Have looked at implementing this design in Mavis: https://github.com/NHSDigital/manage-vaccinations-in-schools-prototype/pull/317

One thing that looks a bit odd to my eyes is having a bottom border when the navigation is displayed vertically:

Screenshot of vertical section navigation component.

Screenshot of vertical section navigation component with multiple headings.

In that last screenshot, I’ve made links use a small 16px font size.

I wonder if there needs to be an explicit small variant, or if links should always appear smaller when in the vertical orientation?

If you look at the Service manual website for example, I suspect larger links are going to appear too dominant on pages?

edwardhorsford:
@paulrobertlloyd your reasoning for smaller links on vertical makes sense to me. @anandamaryon1 any thoughts? could swap to smaller when vertical, and/or support an explicit small option.

I don't have a strong view on underlines, but think the first version of this I did did not have them, and on the call (6 weeks ago?) the panel thought they should be included.

anandamaryon1:
Thanks for testing this out in situ @paulrobertlloyd – very helpful to see. 

Border-bottom on vertical: agree that this should be removed. 

Smaller text on vertical: good point and agree that it works better when used as a sidebar nav. I wonder whether if it were used in the main body, you'd want it body text size instead? That thought makes me err towards making it an option, perhaps with small as the default for vertical.

paulrobertlloyd:
> I don't have a strong view on underlines

Just to be clear, I was talking about the light grey bottom border that is shown below the navigation.

edwardhorsford:
>Just to be clear, I was talking about the light grey bottom border that is shown below the navigation.

Ah I missed that this had slipped in! hard to see in the mini frontend previews. Agree it shouldn't be there.

Fixed in [3263dff](https://github.com/nhsuk/nhsuk-frontend/pull/1886/commits/3263dffe40ba4ec5047bca2430b0917a4c808adb)

edwardhorsford:
Have updated the PR to add a new `small: true` variant which uses 16px font and smaller spacing at all sizes. This is equivalent to the sizing you already get at mobile breakpoint. This was discussed as likely relevant for the stacked variant - the design system nav is already 16px.

paulrobertlloyd:
@edwardhorsford The SCSS currently uses and references the deprecated `nhsuk-link-style-white` mixin, it should now use `nhsuk-link-style-reverse`.

anandamaryon1:
Comparing the focus state for the header nav overflow menu and this vertical nav:

image

image

I think we should try to align them a bit, i.e. make the focus background extend only to the edge of the text. Removes the slightly odd looking L shape black border we currently get on focus too.

paulrobertlloyd:
I quite like the ‘┗’…, looks more polished.

edwardhorsford:
Adding some videos because it's probably worth seeing the interaction with focus on non current items.

service-manual-vertical-focus
section-nav-vertical-focus

anandamaryon1:
> Adding some videos because it's probably worth seeing the interaction with focus on non current items.
> 

Any opinion @edwardhorsford ?

I prefer the focus styles to stay inside the grid, leaving only the active indicator to sit outside in the gutter.

edwardhorsford:
I think like @paulrobertlloyd I mildly prefer the connected L with expanded padding for the yellow background. But I don't have a strong view on it.

I'll also note that the component's padding gives it an increased click target compared to the service manual. With the small variant on vertical, this component has item heights of 40px vs 19px on the service manual. And 44px height for non small variant.

colinrotherham:
This is what the header does (at all sizes) for comparison:

Header with active item and focus (mobile)

<br>
<br>

Header with active item and focus (tablet)

<br>
<br>

Header with active item and focus

colinrotherham:
My view is to follow our current focus style for text, placing the thick black line below only

For the header this meant pushing the active indicator progressively more left as the viewport got wider

edwardhorsford:
Ok, updated so the focus is strictly behind the link. This was done by reducing padding, so decreases the horizontal click area.

|Before| After|
|-|-|
|section-nav-vertical-focus|section-navigation-vertical-focus_new

anandamaryon1:
Should we include a skip link within this nav? Or maybe at least a nunjucks option to add it?

edwardhorsford:
> Should we include a skip link within this nav? Or maybe at least a nunjucks option to add it?

Could you explain a bit more about what you mean? a hidden link at the start that anchors to something at the very end?

anandamaryon1:
@edwardhorsford 
Yeah a hidden link at start that skips past the section nav's links. Given that this can be placed down in the content of a page, within the `<main>` section that the header's skip link points to, then it seems this should have its own one.

edwardhorsford:
We can do, but you're probably not skipping very much - 3-8 links normally? maybe more on casework, but the thing itself probably doesn't scale to huge numbers of links (horizontally at least). It feels different than faceted serach where there might be a load of filters or content to get past. Happy to add if you think we should though.

colinrotherham:
I've done a bit of browser testing, fixed some bugs, and rebased with `main`

* https://nhsuk-frontend-pr-1886.herokuapp.com/nhsuk-frontend/components/section-navigation/

We could do with simplifying the CSS by adding `:not(.nhsuk-section-navigation--reverse)` etc, then we don't have duplicate styles when reversed. Especially true for the current item's left or bottom marker

I'll create a new preview branch shortly

colinrotherham:
Duplicate styles removed

colinrotherham:
```console
✅ Success! To install the branch preview, run:
npm install --save github:nhsuk/nhsuk-frontend#f61cf8490
```

colinrotherham:
Could do with having a `horizontal: true` option

Mainly so avoid the flicker as a vertical navigation switches layout via JavaScript

That way if you know that it'll fit horizontally, you'll only get a flicker if it doesn't (fit)

colinrotherham:
>Could do with having a `horizontal: true` option

I've replaced `vertical: true` with `direction: "vertical"` (or optionally `direction: "horizontal"`)

We can now avoid a flash of vertical styling by setting `direction: "horizontal"` to set the initial layout before the JavaScript runs. If horizontal space is limited, JavaScript will still remove the `--horizontal` modifier

colinrotherham:
```console
✅ Success! To install the branch preview, run:
npm install --save github:nhsuk/nhsuk-frontend#87ca95d59
```

anandamaryon1:
Thanks @colinrotherham 

Any thoughts on adding a skip link option?

colinrotherham:
>Any thoughts on adding a skip link option?

@anandamaryon1 Yeah sounds great, although not via the component

i.e. The section navigation doesn't need to concern itself with the ID of the thing that comes after it

For example, you'd add a skip link before and add an ID after:

```patch
  {% from "section-navigation/macro.njk" import sectionNavigation %}
+ {% from "skip-link/macro.njk" import skipLink %}
  
  <div class="nhsuk-width-container">
    <div class="nhsuk-grid-row">
      <div class="nhsuk-grid-column-one-third">
        <h1 class="nhsuk-heading-l">Care record</h1>
+
+       {{ skipLink({
+         href: "#section-summary",
+         text: "Skip to section: Summary"
+       }) }}
  
        {{ sectionNavigation({
          direction: "vertical",
          items: [
            {
              href: "#section-summary",
              text: "Summary",
              current: true
            },
            {
              href: "#section-observations",
              text: "Observations"
            },
            {
              href: "#section-medicines",
              text: "Medicines"
            },
            {
              href: "#section-allergies",
              text: "Allergies"
            },
            {
              href: "#section-vaccinations",
              text: "Vaccinations"
            },
            {
              href: "#section-referrals",
              text: "Referrals",
              badge: {
                text: "New"
              }
            }
          ]
        }) }}
  
      </div>
-     <div class="nhsuk-grid-column-two-thirds">
+     <div class="nhsuk-grid-column-two-thirds" id="section-summary">
        <h2 class="nhsuk-heading-m">Summary</h2>
        <p>Content for the summary page.</p>
      </div>
    </div>
  </div>
```

colinrotherham:
@anandamaryon1 Worth a mention (from Slack) that our examples now include:

* [Section navigation **with active item**](https://nhsuk-frontend-pr-1886.herokuapp.com/nhsuk-frontend/components/section-navigation/#with-active-item)
* [Section navigation **with current item**](https://nhsuk-frontend-pr-1886.herokuapp.com/nhsuk-frontend/components/section-navigation/#with-current-item)
* [Section navigation **with current (linked) item**](https://nhsuk-frontend-pr-1886.herokuapp.com/nhsuk-frontend/components/section-navigation/#with-current-linked-item)

This gives us the same approach as the header component where:

* Enabling `item.current` adds `aria-current="page"` (current page)
* Enabling `item.active` adds `aria-current="true"` (current page is within this section)

Styling works for either, whether linked or not

With emphasis on our usual disclaimer:

>These examples are used for automated tests and may not follow service manual best practice.

colinrotherham:
@anandamaryon1 Should we automatically add the skip link when `item.current` is true?

**Update:** It would rely on us having the item `href: "#section-summary"` to link to

anandamaryon1:
> > Any thoughts on adding a skip link option?
> 
> @anandamaryon1 Yeah sounds great, although not via the component
> 
> i.e. The section navigation doesn't need to concern itself with the ID of the thing that comes after it

Ah right yeah that makes sense. One to add to design system as an example to show how to do it then. 

Not sure what you mean for automatically adding skip link?

colinrotherham:
>Not sure what you mean for automatically adding skip link?

As in, we could automatically insert [all those green lines](https://github.com/nhsuk/nhsuk-frontend/pull/1886#issuecomment-5495454390)

anandamaryon1:
> > Not sure what you mean for automatically adding skip link?
> 
> As in, we could automatically insert [all those green lines](https://github.com/nhsuk/nhsuk-frontend/pull/1886#issuecomment-5495454390)

I'm thinking making it an option would be better, given that we don't know where section nav will be used on the page, and it might be placed before the `#maincontent` and therefore the normal page template's skip link will jump past it anyway?

edwardhorsford:
I don't think it should automatically be inserted unless there's a way to turn it off. If the section nav is just 2 items, it would seem more noisy than helpful to have a skip link. If a team is using it for side nav and has several section navs, it's likely better to have one skip link that skips all of the section navs rather than one per section nav.

colinrotherham:
Agree with @edwardhorsford, it should be a manual consideration

I'd forgotten that we already have `{% block skipLink %}` for skip links

If you insert them elsewhere they'd only be visible (due to `position: absolute`) at the top of the page

```njk
{% block skipLink %}
  {{ skipLink({
    href: "#maincontent",
    text: "Skip to main content"
  }) }}

  {{ skipLink({
    href: "#section-summary",
    text: "Skip to section: Summary"
  }) }}
{% endblock %}
```

>…If a team is using it for side nav and has several section navs, it's likely better to have one skip link that skips all of the section navs rather than one per section nav.

@edwardhorsford Any thoughts on how we can support a single `<nav>` with multiple `<ul>` within?

For example, the service manual navigation:
https://service-manual.nhs.uk/design-system/components

Something we can perhaps support when vertical?

edwardhorsford:
>Something we can perhaps support when vertical?

I somewhat assumed we'd look at multi-level navigation as a feature to add once this first version is released.

frankieroberto:
I've had a quick experiment using the section navigation for a second-level horizontal navigation for the NHS App design system website at https://github.com/nhsuk/nhsapp-frontend/pull/584

Here’s how it looks ([preview](https://deploy-preview-584--nhsapp-frontend.netlify.app/get-started/)):

Screenshot 2026-09-09 at 16 20 10

Overall I think it works ok visually, but it's a little jarring that the font size is slightly different between the 2 navs (16px vs 19px) and the vertical height is slightly different (56px vs 60px). Should we align the 2?

I know it wasn’t necessarily designed to go immediately below the header, but this feels like a reasonable use-case? This could also work on the NHS Service manual website itself, so that the second level nav within the Design system section contains "Styles", "Components" and "Patterns", to make it easier to move between those sub-sections?

anandamaryon1:
@frankieroberto there is a small variant which should help. But good point on matching the vertical size. We should probably squash this section nav to 56px to match header nav.

colinrotherham:
Similarly if we're proposing it sits outside of a grid/row/column, it'll need `containerClasses` support

Along with the default `nhsuk-width-container` like we have in the header:

```njk
<div class="nhsuk-header__container nhsuk-width-container {%- if params.containerClasses %} {{ params.containerClasses }}{% endif %}">
```

This allows teams to override the container width on specific pages without Sass config changes:

```
{% set containerClasses = "app-width-container--wide" %}
```

We can override and suppress the container styles when within a column layout

anandamaryon1:
To do:
- [ ] Support `visuallyHiddenText` positioning
- [x] Consider height matching to primary nav bar (edit: matched `small` variant height to header nav)
- [x] Fixes for in or out of page width container
- [ ] Add support for arrays of lists and headings

Skip links: we'll not add these specifically to this component, they can be added by adding another link within the usual skip link component

colinrotherham:
I've rebased this to pick up the new review app layouts

Would be great to add some full page examples to fully demonstrate the horizontal layout below the header, but not within a column. Bear in mind this requires a slight refactor to support container classes, like the header:

```njk
<div class="nhsuk-header__container nhsuk-width-container
  {%- if params.containerClasses %} {{ params.containerClasses }}{% endif %}">
```

We'd then need to override the redundant width container when within a column

colinrotherham:
New preview build:

```console
✅ Success! To install the branch preview, run:
npm install --save github:nhsuk/nhsuk-frontend#091bc51c2
```

PR #1889, opened 2026-04-15
Keep hidden components hidden

frankieroberto:
Browsers by default will hide any element with a `hidden` attribute using `display: none`.

However if you set a different `display` property on a class then the item will no longer be hidden even if the attribute is set.

This updates our CSS so that we don't set a display property on components which have a hidden attribute.

This is mainly useful for any pages using JavaScript to dynamically show or hide elements.

Also adds a test page with all the components hidden.

### DIFF
--- a/packages/nhsuk-frontend/src/nhsuk/components/header/_index.scss
+++ b/packages/nhsuk-frontend/src/nhsuk/components/header/_index.scss
@@ -79,7 +79,7 @@ $nhsuk-header-reverse-item-active-colour: $nhsuk-reverse-text-colour;
     background-color: $nhsuk-brand-colour;
 
     // Add shadow to bottom of header when open
-    &:has(.nhsuk-header__menu-list:not([hidden])) {
+    &:has(.nhsuk-header__menu-list) {
       border-bottom: nhsuk-spacing(1) solid $nhsuk-border-colour;
     }
 
@@ -562,10 +562,6 @@ $nhsuk-header-reverse-item-active-colour: $nhsuk-reverse-text-colour;
     align-self: center;
     padding: 0 math.div($nhsuk-gutter-half, 2);
 
-    &[hidden] {
-      display: none;
-    }
-
     @include nhsuk-media-query($from: tablet) {
       padding: 0 $nhsuk-gutter-half;
     }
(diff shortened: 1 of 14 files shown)


packages/nhsuk-frontend/src/nhsuk/core/objects/_button-group.scss
```diff
@@ -3,6 +3,12 @@
 @use "../tools" as *;
 
 @include nhsuk-exports("nhsuk/core/objects/button-group") {
+
+  // Hide hidden button groups
+  .nhsuk-button-group[hidden] {
+    display: none;
+  }
```
frankieroberto: This would ideally be a `:not([hidden])` suffix in the css below, but that doesn't work for some Sass reason to do with the other nested selectors.

colinrotherham:
Thanks @frankieroberto, this is great 🙌

We need to take care when it comes to specificity as `:not([hidden])` is like adding another class

* `.nhsuk-back-link:not([hidden])` now overrides `.nhsuk-back-link--reverse`
* `.nhsuk-button:not([hidden])` now overrides `.nhsuk-button--icon`

The latter is only on `main` but you'd see ~`display: flex`~ `display: inline-block` for icon buttons

From our Slack chat, I wonder if your `[hidden] { display: none !important; }` idea is better?

Backstop report showing back link specificity issue

frankieroberto:
@colinrotherham ah! 

I’m now thinking that a global `*[hidden] { } ` selector would be better - less likely that we forget to exclude hidden elements in other css.

frankieroberto:
Switched to a global `!important` rule in d7c362383e5d9387617522f3bc9f5883f6b36255.

packages/nhsuk-frontend/src/nhsuk/components/header/_index.scss
```diff
@@ -79,7 +79,7 @@ $nhsuk-header-reverse-item-active-colour: $nhsuk-reverse-text-colour;
     background-color: $nhsuk-brand-colour;
 
     // Add shadow to bottom of header when open
-    &:has(.nhsuk-header__menu-list:not([hidden])) {
+    &:has(.nhsuk-header__menu-list) {
```
colinrotherham, line 82: This is drawing a grey border below the menu even when it's closed, sorry 😣

Image

frankieroberto: Good spot!

PR #1892, opened 2026-04-16
[v10] Refactor character count method to reduce repeated updates

### LINKED ISSUE #1619: Character count component counts code points, not characters

colinrotherham:
## Description

This PR uplifts https://github.com/alphagov/govuk-frontend/pull/6449 from GOV.UK Frontend v6.0.0 to improve the character count

Our JavaScript API is public by default (we publish types) so I've added 7b8cd787df2f2f66f917c11994a6f1c908666d28 and d66c040a84c35de9546e720e9b54edfb4aba251f for compatibility

Credit to @querkmachine for the original changes

Unblocks https://github.com/nhsuk/nhsuk-frontend/issues/1619

## Checklist

- [x] CHANGELOG entry

### DIFF
--- a/packages/nhsuk-frontend/src/nhsuk/components/character-count/character-count.mjs
+++ b/packages/nhsuk-frontend/src/nhsuk/components/character-count/character-count.mjs
@@ -179,9 +179,11 @@ export class CharacterCount extends ConfigurableComponent {
   /**
    * Count the number of characters (or words, if `config.maxwords` is set)
    * in the given text, and update the component-wide count
+   *
+   * @param {string} [text] - Deprecated
    */
-  updateCount() {
-    const text = this.$textarea.value
+  updateCount(text) {
+    text = text ?? this.$textarea.value
 
     if (this.config.maxwords) {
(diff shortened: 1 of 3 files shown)


packages/nhsuk-frontend/src/nhsuk/components/character-count/character-count.mjs
```diff
@@ -192,6 +194,15 @@ export class CharacterCount extends ConfigurableComponent {
     this.length = text.length
   }
 
+  /**
```
MatMoore, line 197: can we remove the deprecated methods again in this PR? Seems like as long as there's a replacement we could update the tests to use that...

MatMoore (approved):
LGTM, only thing I'm not sure about it is why we need to add deprecated methods in?

packages/nhsuk-frontend/src/nhsuk/components/character-count/character-count.mjs
colinrotherham: We can remove them in v11

It's unlikely, but teams might be using them in v10.x so it would be a breaking change

PR #1895, opened 2026-04-17
[v10] Add character count `countType: "characters"` option using Intl.Segmenter

### LINKED ISSUE #1619: Character count component counts code points, not characters

colinrotherham:
## Description

This PR adds character count support for `countType: "characters"` using Intl.Segmenter

It's a non-breaking change because `countType: "length"` is the default

```patch
  {{ characterCount({
    label: {
      text: "Can you provide more detail?"
    },
    name: "more-detail",
-   maxlength: 200
+   maxlength: 200,
+   countType: "characters"
  }) }}
```

Closes #1619

## Checklist

- [ ] CHANGELOG entry


colinrotherham:
I'm not sure whether to make `countType: "length"` (code points) the default as indicated by:

* https://github.com/nhsuk/nhsuk-frontend/issues/1619

That leaves us free to use `countType: "characters"` as our opt-in for graphemes

I've made a similar comment in https://github.com/alphagov/govuk-frontend/issues/1364#issuecomment-4280763394

**Update:** We did do this

packages/nhsuk-frontend/src/nhsuk/components/character-count/character-count.mjs
```diff
@@ -535,7 +549,7 @@ export function initCharacterCounts(options) {
  * @property {number} [maxwords] - Deprecated. Use `maxlength` and `countType: "words"` instead.
  * @property {number} [threshold=0] - The percentage value of the limit at which point the count message is displayed.
  *   If this attribute is set, the count message will be hidden by default.
- * @property {'characters' | 'words'} countType - The count type (`"characters"` or `"words"`) used to count the text.
+ * @property {'length' | 'characters' | 'words'} countType - The count type (`"length"`, `"characters"` or `"words"`) used to count the text.
```
MatMoore: We could possibly name the default 'codepoints' instead of 'length' to be clearer

MatMoore (approved):

packages/nhsuk-frontend/src/nhsuk/components/character-count/character-count.mjs
colinrotherham: I did actually use `"code-points"` initially but thought it might be more confusing for non-developers

I've picked `"length"` to match Nunjucks `| length` → JavaScript `.length`

We could actually have no default and make it `undefined`?

```suggestion
 * @property {'characters' | 'words'} [countType] - The count type (`"characters"` or `"words"`) used to count the text.
```

MatMoore: Ah fair enough, lets leave it as length then

PR #1747, opened 2025-12-19
[v10] Add support for array values in component configs

colinrotherham:
## Description

This PR adds a new `array` config schema type (alongside the existing `string`, `boolean`, `object` types)

We can currently configure a component by:

- [setting Nunjucks macro options](https://github.com/nhsuk/nhsuk-frontend/blob/main/docs/configuration/README.md#setting-nunjucks-macro-options)
- [passing JavaScript configuration](https://github.com/nhsuk/nhsuk-frontend/blob/main/docs/configuration/README.md#passing-javascript-configuration)
- [adding HTML data attributes](https://github.com/nhsuk/nhsuk-frontend/blob/main/docs/configuration/README.md#adding-html-data-attributes)

Once this PR is merged, arrays will be supported in the following configurations:

### Nunjucks macro options

Config values added via `nhsukAttributes()` using an item `type: "array"` option

```njk
{{- nhsukAttributes({
  "data-attributes": {
    value: ["value1", "value2"],
    type: "array"
  }
}) -}}
```

### JavaScript configuration

Config array values via `initAll()` and `createAll()`, for example to configure additional classes:

```mjs
import { FileUpload, createAll } from 'nhsuk-frontend'

createAll(FileUpload, {
  chooseFilesButtonClassList: [
    'nhsuk-button--secondary',
    'nhsuk-button--small'
  ],
  i18n: {
    chooseFilesButton: 'Choose files'
  }
})
```

**Note:** Otherwise the [`classList` browser API](https://developer.mozilla.org/en-US/docs/Web/API/Element/classList) throws when space-separated classes are provided:

```console
InvalidCharacterError: The token provided contains HTML space characters, which are not valid in tokens.
```

### HTML data attributes

Config values added via HTML `data-*` attributes:

```html
<div data-module="app-component" data-attributes='["value1", "value2"]'>
```

## Checklist

- [ ] Tested against our [testing policy](https://github.com/nhsuk/nhsuk-frontend/blob/master/docs/contributing/testing.md) (Resolution, Browser & Accessibility)
- [x] Follows our [coding standards and style guide](https://github.com/nhsuk/nhsuk-frontend/blob/master/docs/contributing/coding-standards.md)
- [ ] CHANGELOG entry

PR #1896, opened 2026-04-20
[v10] Add support for functions in component configs

### LINKED ISSUE #1619: Character count component counts code points, not characters

colinrotherham:
## Description

This PR adds a new `function` config schema type (alongside `string`, `boolean`, `object` and `array` types)

Support for `array` types was added recently in:

* https://github.com/nhsuk/nhsuk-frontend/pull/1747

### JavaScript configuration

Config function values via `initAll()` and `createAll()`, for example to configure the character count function:

```mjs
import { CharacterCount, createAll } from 'nhsuk-frontend'

createAll(CharacterCount, {
  countFunction: (string) => string.length
})
```

## Checklist

- [ ] CHANGELOG entry

### DIFF
--- a/packages/nhsuk-frontend/src/nhsuk/common/configuration/normalise-string.mjs
+++ b/packages/nhsuk-frontend/src/nhsuk/common/configuration/normalise-string.mjs
@@ -14,17 +14,16 @@
  * @returns Normalised data
  */
 export function normaliseString(value, property) {
+  let outputType = property?.type
+
   if (
     !isValid(value) ||
-    (property?.type && !['string', 'number', 'boolean'].includes(property.type))
+    (outputType && !['string', 'number', 'boolean'].includes(outputType))
   ) {
     return
   }
 
   const trimmedValue = value.toString().trim()
-
-  let output
-  let outputType = property?.type
 
   // No schema type set? Determine automatically
   if (!outputType) {
@@ -38,6 +37,8 @@ export function normaliseString(value, property) {
       outputType = 'number'
     }
   }
+
+  let output
 
   switch (outputType) {
     case 'boolean':
--- a/packages/nhsuk-frontend/src/nhsuk/common/configuration/normalise-string.mjs
+++ b/packages/nhsuk-frontend/src/nhsuk/common/configuration/normalise-string.mjs
@@ -14,7 +14,10 @@
  * @returns Normalised data
  */
 export function normaliseString(value, property) {
-  if (!isValid(value)) {
+  if (
+    !isValid(value) ||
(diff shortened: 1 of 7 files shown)


packages/nhsuk-frontend/src/nhsuk/common/configuration/normalise-string.mjs
MatMoore: maybe move this to the top so you can use it in the precondition, as it appears in that expression twice.

MatMoore (commented):
The changes themselves look fine, but I am just wondering if there are any security considerations here, so just wanted to flag that.

It seems like the function type is effectively a way to embed a form of inline script, and if you are otherwise blocking inline scripts via a CSP, is this circumventing that protection?

colinrotherham:
> It seems like the function type is effectively a way to embed a form of inline script, and if you are otherwise blocking inline scripts via a CSP, is this circumventing that protection?

Thankfully, if this was configured via inline `<script>`, it would still be blocked by `unsafe-inline`

You'll also see that when it comes to HTML:

1. [String `data-*` attributes do not allow arrays, functions or objects](https://github.com/nhsuk/nhsuk-frontend/blob/component-config-functions/packages/nhsuk-frontend/src/nhsuk/common/configuration/normalise-string.jsdom.test.mjs#L144-L159)
2. [Function `data-*` attributes are automatically dropped](https://github.com/nhsuk/nhsuk-frontend/blob/component-config-functions/packages/nhsuk-frontend/src/nhsuk/common/configuration/normalise-dataset.jsdom.test.mjs#L45)

Config functions are only allowed to be set via `initAll()`, `createAll()` or `new Component()`

Although with JavaScript being JavaScript, you can always continue to overwrite things at will anyway:

```mjs
CharacterCount.prototype.updateCount = () => {
  // Overwrite count function
}
```

```mjs
class CharacterCountDanger extends CharacterCount {
 updateCount() {
   // Extend count function
 }
}
```

i.e. Adding a config option hasn't introduced any new ways to override functions

packages/nhsuk-frontend/src/nhsuk/common/configuration/normalise-string.mjs
colinrotherham: Thanks @MatMoore, done in https://github.com/nhsuk/nhsuk-frontend/pull/1896/commits/3617004a291917ee245d5a2225bdf948fdffbe74

MatMoore:
Ah, that makes sense. I was thinking they would be configurable by data attributes.

MatMoore (approved):

PR #1897, opened 2026-04-20
[v10] Add character count `countFunction` option

### LINKED ISSUE #1619: Character count component counts code points, not characters

colinrotherham:
## Description

This PR adds a new `countFunction` option to character counts to cater for server-side differences in:

- New lines that vary due to `\n` versus `\r\n`
- Word counts that vary based on empty space and punctuation
- How empty space is trimmed before counting
- Support for multi-byte strings

For example, services might already count multi-byte strings server-side (e.g. [`len()` in Python](https://docs.python.org/3.9/library/functions.html?highlight=len#len)) resulting in client-side count mismatches, yet [upcoming support for grapheme `countType`](https://github.com/nhsuk/nhsuk-frontend/pull/1895) is blocked by a [3rd party library integration](https://grapheme.readthedocs.io/en/latest/grapheme.html)

Support for a custom `countFunction` means teams can close this support gap

## Checklist

- [ ] CHANGELOG entry

### DIFF
--- a/packages/nhsuk-frontend/src/nhsuk/components/character-count/character-count.mjs
+++ b/packages/nhsuk-frontend/src/nhsuk/components/character-count/character-count.mjs
@@ -61,6 +61,7 @@ export class CharacterCount extends ConfigurableComponent {
     const {
       i18n,
       maxlength,
+      countFunction,
       countType,
       screenReaderCountMessageClass,
       textareaDescriptionClass,
@@ -72,7 +73,7 @@ export class CharacterCount extends ConfigurableComponent {
       locale: closestAttributeValue(this.$root, 'lang')
     })
 
-    if (countType === 'characters') {
+    if (countType === 'characters' || !!countFunction) {
       if (!('Segmenter' in Intl)) {
         throw new SupportError(
           formatErrorMessage(
@@ -83,7 +84,7 @@ export class CharacterCount extends ConfigurableComponent {
       }
 
       this.segmenter = new Intl.Segmenter(this.i18n.locale, {
-        granularity: 'grapheme'
+        granularity: countType === 'words' ? 'word' : 'grapheme'
       })
     }
 
@@ -221,10 +222,11 @@ export class CharacterCount extends ConfigurableComponent {
    */
   updateCount(text) {
     const { $textarea, countFunctions } = this
-    const { countType } = this.config
+    let { countType, countFunction } = this.config
 
     text ??= $textarea.value
-    this.length = countFunctions[countType].call(this, text)
+    countFunction ??= countFunctions[countType]
+    this.length = countFunction.call(this, text)
   }
 
   /**
@@ -451,7 +453,7 @@ export class CharacterCount extends ConfigurableComponent {
    * Character count functions
    *
    * @constant
-   * @satisfies {Record<string, (this: CharacterCount, text: string) => number>}
+   * @satisfies {Record<string, CharacterCountConfig['countFunction']>}
    */
   countFunctions = Object.freeze({
     /**
@@ -544,6 +546,7 @@ export class CharacterCount extends ConfigurableComponent {
       maxlength: { type: 'number' },
       threshold: { type: 'number' },
       countType: { type: 'string' },
+      countFunction: { type: 'function' },
       textareaDescriptionClass: { type: 'string' },
       visibleCountMessageClass: { type: 'string' },
       screenReaderCountMessageClass: { type: 'string' },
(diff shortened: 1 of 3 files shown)


packages/nhsuk-frontend/src/nhsuk/components/character-count/character-count.mjs
```diff
@@ -492,6 +489,24 @@ export class CharacterCount extends ConfigurableComponent {
...
   })
 
+  /**
+   * Character count functions
+   *
+   * @constant
+   * @satisfies {Record<string, (text: string, segmenter?: Intl.Segmenter | null) => number>}
+   */
+  static countFunctions = Object.freeze({
+    characters(text, segmenter) {
```
MatMoore: Some thoughts on the interface here...

Currently there are two things that influence the choice of algorithm: the selection of `countFunction`, and whether segmenter is set or not. Wondering if it would be simpler to have explicit functions for graphemes as well as characters and words? In that case passing `countType: "graphemes"` would be syntactic sugar for passing `countFunction: countFunctions.graphemes` and the same goes for "characters" and "words".

It also seems slightly inconsistent that if you pass `countType: "words"` with a custom function, we pass through a segmenter that can do word segmentation, but if you don't pass the custom function, it doesn't use the segmenter.

Maybe if the interface is simplified to `fn(text)` the segmenter could be an internal implementation detail?

colinrotherham: Yeah, I partly did this because [**Intl.Segmenter** also supports `granularity: "word"`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Segmenter) 😮

So instead of `countType: "words"` we could keep `maxwords` but set `granularity: "word"`

I'll see what feedback comes from the GOV.UK Design System team

>Maybe if the interface is simplified to fn(text) the segmenter could be an internal implementation detail?

Yeah, I'm glad you said. I reckon for now let's ditch the static property and keep it all internal, especially with a potential performance cost to creating a new segmenter on every keystroke (hence injecting it in)

colinrotherham: Feedback applied, thanks @MatMoore

Regarding word counts via the segmenter, I've messaged @romaricpascal [and added a comment](https://github.com/alphagov/govuk-frontend/issues/1364#issuecomment-4280763394) as the [Unicode **Default Word Boundary Specification**](https://unicode.org/reports/tr29/#Default_Word_Boundaries) varies quite significantly from `/\S+/g`

For example, consider this phrase:

>My mother-in-law—Wait, what?

It matches only 3 words currently:

```js
["My", "mother-in-law—Wait,", "what?"]
```

Yet using the segmenter with `granularity: "word"` it matches 6 words:

```js
["My", "mother", "in", "law", "Wait", "what"]
```

MatMoore: > Yeah, I'm glad you said. I reckon for now let's ditch the static property and keep it all internal, especially with a potential performance cost to creating a new segmenter on every keystroke (hence injecting it in)

Yeah I figured we wouldn't want to instantiate one each call. The way you've done it is what I was imagining.

That unicode word boundary spec seems like a cleaner way to do word counting if we were implementing from scratch, but I guess it's a potentially breaking change so maybe it's not worth it if we're keeping `maxwords`. Are we going to wait for the govuk implementation before we merge this?

colinrotherham: Thanks. Yeah it'll be discussed at their next dev catch up

What we know is that they've [proposed for `maxwords` to be deprecated](https://github.com/alphagov/govuk-frontend/issues/1364#issuecomment-4161213597)

We could take the `countType` option as an opt-in to use segmenter. But as [this comment mentions](https://github.com/alphagov/govuk-frontend/issues/1364#issuecomment-494042062) that might mean upgrading `/\S+/g` to a better word boundary regex for [older browsers that lack support](https://caniuse.com/wf-intl-segmenter)

colinrotherham: They've not had chance to discuss anything character count related other than this comment:

* https://github.com/alphagov/govuk-frontend/pull/6995#pullrequestreview-4235206737

But that applies to https://github.com/nhsuk/nhsuk-frontend/pull/1899 primarily

For this PR all count functions now use the same interface so it's ready for another review

MatMoore (approved):

colinrotherham:
@anandamaryon1 I'll be adding a changelog entry into #1899 to wrap up this stack of work:

* https://github.com/nhsuk/nhsuk-frontend/pull/1893
* https://github.com/nhsuk/nhsuk-frontend/pull/1895
* https://github.com/nhsuk/nhsuk-frontend/pull/1897
* https://github.com/nhsuk/nhsuk-frontend/pull/1899

With this initial PR used to uplift some more recent GOV.UK Frontend changes:

* https://github.com/nhsuk/nhsuk-frontend/pull/1892

Plus this PR to allow components to configure functions within their options:

* https://github.com/nhsuk/nhsuk-frontend/pull/1896

---

We've contributed our improvements back to GOV.UK Frontend via:

* https://github.com/alphagov/govuk-frontend/pull/6995

PR #1911, opened 2026-04-29
Fix clean task `Error: ENOTEMPTY: directory not empty, rmdir`

colinrotherham:
## Description

This PR is a workaround to an issue with [Node.js `fsPromises.rm()`](https://nodejs.org/docs/latest-v24.x/api/fs.html#fspromisesrmpath-options) where deleting large directories such as `packages/nhsuk-frontend/dist` throws `ENOTEMPTY` or other errors [when the file system is busy](https://github.com/nodejs/node/issues/54561#issuecomment-2449076853)

```console
[Error: ENOTEMPTY: directory not empty, rmdir '/path/to/nhsuk-frontend/packages/nhsuk-frontend/dist'] {
  errno: -66,
  code: 'ENOTEMPTY',
  syscall: 'rmdir',
  path: '/path/to/nhsuk-frontend/packages/nhsuk-frontend/dist'
}
```

## Checklist

- [ ] CHANGELOG entry

### DIFF
--- a/packages/nhsuk-frontend/package.json
+++ b/packages/nhsuk-frontend/package.json
@@ -53,7 +53,7 @@
     "!**/*.test.*"
   ],
   "scripts": {
-    "clean": "del-cli dist",
+    "clean": "del-cli dist/nhsuk/*",
     "postclean": "tsc --build --clean",
     "build": "gulp build --color",
     "watch": "gulp watch --color",
--- a/shared/lib/components.mjs
+++ b/shared/lib/components.mjs
@@ -1,5 +1,5 @@
 import { readFileSync } from 'node:fs'
-import { basename, join } from 'node:path'
+import { basename, dirname, join } from 'node:path'
 
 import { paths } from '@nhsuk/frontend-config'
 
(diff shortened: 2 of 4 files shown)


packages/nhsuk-frontend/package.json
```diff
@@ -53,7 +53,7 @@
     "!**/*.test.*"
   ],
   "scripts": {
-    "clean": "del-cli dist",
+    "clean": "del-cli dist/nhsuk/*",
```
colinrotherham, line 56: The [`del` package](https://www.npmjs.com/package/del) doesn't set [`maxRetries`](https://nodejs.org/docs/latest-v24.x/api/fs.html#fspromisesrmpath-options) when [removing files or directories](https://github.com/sindresorhus/del/blob/v8.0.1/index.js#L69)

But updating the glob `dist` → `dist/nhsuk/*` [splits out multiple paths to delete](https://github.com/sindresorhus/del/blob/main/index.js#L48-L49) which appears to solve the problem (i.e. smaller batched deletes rather than one large delete)

shared/lib/components.mjs
```diff
@@ -76,12 +76,12 @@ export async function loadAll() {
  * Get component names
  */
 export function getNames() {
-  const listing = files.getDirectories('nhsuk/components', {
+  const listing = files.getListing('nhsuk/components/*/template.njk', {
```
colinrotherham, line 79: This change fixes an issue where switching branches would leave empty directories behind, which then causes errors when trying to run file operations or build tasks on them

E.g. `packages/nhsuk-frontend/src/nhsuk/components/search-input`

frankieroberto (approved):

PR #1921, opened 2026-05-08
[v10] Add a template with all components imported

frankieroberto:
This adds a `template-with-imports.njk` to the `nhsuk-frontend`, so that users can choose to use that instead of `template.njk` and import all components rather than add individual component imports per page or on their own layout page.

Also updates the review app to use this new template.


packages/nhsuk-frontend/src/nhsuk/template-with-imports.njk
```diff
@@ -0,0 +1,43 @@
+{% extends "template.njk" %}
+
+{% from "nhsuk/macros/attributes.njk" import nhsukAttributes -%}
+{% from "nhsuk/macros/icon.njk" import nhsukIcon %}
```
colinrotherham: Mind if we keep the new line to keep the components separate?

There's a missing `%}` → `-%}` too

```suggestion
{% from "nhsuk/macros/attributes.njk" import nhsukAttributes -%}
{% from "nhsuk/macros/icon.njk" import nhsukIcon -%}

```

packages/nhsuk-frontend/src/nhsuk/template-with-imports.njk
```diff
@@ -0,0 +1,43 @@
+{% extends "template.njk" %}
```
colinrotherham: Need to trim the new line here too?

```suggestion
{% extends "template.njk" -%}
```

packages/nhsuk-frontend/src/nhsuk/template-with-imports.njk
```diff
@@ -0,0 +1,43 @@
...
+{% from "nhsuk/components/fieldset/macro.njk" import fieldset -%}
+{% from "nhsuk/components/file-upload/macro.njk" import fileUpload -%}
+{% from "nhsuk/components/footer/macro.njk" import footer -%}
+{% from "nhsuk/components/header/macro.njk" import header -%}
+{% from "nhsuk/components/hero/macro.njk" import hero -%}
+{% from "nhsuk/components/hint/macro.njk" import hint -%}
+{% from "nhsuk/components/images/macro.njk" import image -%}
+{% from "nhsuk/components/input/macro.njk" import input -%}
+{% from "nhsuk/components/inset-text/macro.njk" import insetText -%}
+{% from "nhsuk/components/label/macro.njk" import label -%}
```
colinrotherham, line 29: We've lost a component

```suggestion
{% from "nhsuk/components/label/macro.njk" import label -%}
{% from "nhsuk/components/legend/macro.njk" import legend -%}
```

CHANGELOG.md
```diff
@@ -148,6 +148,10 @@ If you use Sass and you've changed the default `/assets/` public path (for examp
 
 This was added in [pull request #1915: Render Nunjucks icon macro into SVG files](https://github.com/nhsuk/nhsuk-frontend/pull/1915).
 
+#### Add a template with all components imported
+
+You can now choose to use `template-with-imports.njk` as a base template, which will import all components.
```
colinrotherham: We typically assume the changelog caters for both Nunjucks and HTML users

Shall we clarify this a bit too?

```suggestion
If you are using the Nunjucks macros, you can now extend the `template-with-imports.njk` base template instead of `template.njk`, which will import all components.
```

packages/nhsuk-frontend/src/nhsuk/template-with-imports.njk
frankieroberto: Ahh good spot.

packages/nhsuk-frontend/src/nhsuk/template-with-imports.njk
frankieroberto: Good idea.

packages/nhsuk-frontend/src/nhsuk/template-with-imports.njk
frankieroberto: Can do!

packages/nhsuk-frontend/src/nhsuk/template-with-imports.njk
```diff
@@ -0,0 +1,45 @@
...
+{% from "nhsuk/components/contents-list/macro.njk" import contentsList -%}
+{% from "nhsuk/components/date-input/macro.njk" import dateInput -%}
+{% from "nhsuk/components/details/macro.njk" import details -%}
+{% from "nhsuk/components/do-dont-list/macro.njk" import list -%}
+{% from "nhsuk/components/error-message/macro.njk" import errorMessage -%}
+{% from "nhsuk/components/error-summary/macro.njk" import errorSummary -%}
+{% from "nhsuk/components/fieldset/macro.njk" import fieldset -%}
+{% from "nhsuk/components/file-upload/macro.njk" import fileUpload -%}
+{% from "nhsuk/components/footer/macro.njk" import footer -%}
+{% from "nhsuk/components/header/macro.njk" import header -%}
```
frankieroberto, line 23: Technically header, footer, skip link and attributes don't need to be imported here, as this template extends `template.njk` which already imports those, as they're needed for that template itself.

But not sure if omitting them would be confusing? (And importing them twice doesn’t have any huge downsides)

anandamaryon1:
Would this be for use in the prototype kit or useful for prod services too? I'd assume there are downsides to importing everything upfront?

colinrotherham:
@anandamaryon1 For teams that use the same component names as us [we've seen issues where our component takes precedence over their custom (previously overridden) version](https://github.com/NHSDigital/manage-vaccinations-in-schools-prototype/pull/178#issuecomment-4378966032)

We might have solved this in the prototype kit by [introducing the `nhsuk/components` namespace](https://github.com/nhsuk/nhsuk-prototype-kit-package/pull/325)

Just in case, the changes in this PR are opt-in via **template-with-imports.njk** leaving **template.njk** as-is

>Would this be for use in the prototype kit or useful for prod services too?…

But we'd make it the default for the kit

frankieroberto:
@anandamaryon1 I must admit I was mainly thinking of the prototype kit. But it could also be useful for static-site builders like Eleventy (used by the prototype kit website and the design history sites)?

I guess in theory there's a slight performance penalty in importing Nunjucks macros that you're not using, but I’d guess that it is fairly minimal (and not relevant for static site builders).

I don’t know whether any production services using Nunjucks would want to automatically import all macros, or would do it selectively - but then I don’t think we have many production services using Nunjucks?

anandamaryon1:
Ok fair enough. It's opt-in so low/no risk 👍

Just wanted to understand it. 

Will it require manually updating when we add new components? Is there a way to avoid that?

colinrotherham:
>Will it require manually updating when we add new components? Is there a way to avoid that?

Yeah we'll still need `nhsuk-frontend` to add the import

We could make add a new review app page that loops the component directory names and attempts to render the default component? That way any without an import would fail the review app build

frankieroberto:
I'll admit that half my motivation for this is that in the past I've forgotten to add new components to the [prototype kit template](https://github.com/nhsuk/nhsuk-prototype-kit-package/blob/main/lib/views/prototype-kit-template.njk), so this kinda pushes the problem upstream... 😂

CHANGELOG.md
```diff
@@ -148,6 +148,10 @@ If you use Sass and you've changed the default `/assets/` public path (for examp
 
 This was added in [pull request #1915: Render Nunjucks icon macro into SVG files](https://github.com/nhsuk/nhsuk-frontend/pull/1915).
 
+#### Add a template with all components imported
+
+If you are using the Nunjucks macros, you can now extend the `template-with-imports.njk` base template instead of `template.njk`, which will import all components.
```
colinrotherham: Noticed we usually say "our Nunjucks page template" since it was only added in v9.4.0

Typically with the guidance page linked up too

Shall we make an edit and add the usual PR link?

```suggestion
If you are using our Nunjucks [page template](https://service-manual.nhs.uk/design-system/styles/page-template), you can now extend `template-with-imports.njk` instead of `template.njk` to automatically import all components.

This was was added in [pull request #1921: Add a template with all components imported](https://github.com/nhsuk/nhsuk-frontend/pull/1921).
```

colinrotherham (approved):

PR #1938, opened 2026-05-29
[v10] Add back link `direction: "forward"` option

colinrotherham:
## Description

This PR adds optional support for `direction: "forward"` on the back link component:

Back link with forward direction

## Checklist

- [ ] CHANGELOG entry


anandamaryon1:
Interesting one, seems to be suitable only for a quite specific use-case. Do you have any background context to this? 

My concern with this is that it's very subtle, particularly if used top-left in place of a normal back link. I can imagine users clicking out of habit expecting to go back but in fact going forward. 

I think we'd need some usage guidance and testing on this.

PR #1946, opened 2026-06-01
[v10] Updates to link classes and mixins, support for action link as a button

colinrotherham:
## Description

This PR adds the `element` and `type` options to action links:

```njk
{{ actionLink({
  text: "Find your nearest A&E",
  element: "button"
}) }}
```

I've also ported various link classes and mixins from:

* https://github.com/nhsuk/nhsuk-frontend/pull/1367

**Note:** Any breaking changes (e.g. link hover, underlines) have been omitted

### Element overrides

Typically these element overrides are used by the NHS.UK React components where `a` and `button` elements allow flexibility with Single Page Application (SPA) routers where links aren't always used

To align with the back link component, action links now have styles to render as a button:

Action link as a button

## Checklist

- [x] CHANGELOG entry


frankieroberto (approved):
I think we could consider deprecating this option for the back links and action links in future? There ought to be no reason why these need to be buttons?

But given we already support it for the back link I guess it’s ok to add support for the action link for now, for consistency if nothing else?

edwardhorsford:
> I think we could consider deprecating this option for the back links and action links in future? There ought to be no reason why these need to be buttons?
> 
> But given we already support it for the back link I guess it’s ok to add support for the action link for now, for consistency if nothing else?

In my team's usage we'd used it in place of a button - but the action it's doing needs to be POST, hence wanting to use a button element behind the scenes.

colinrotherham:
I've (temporarily) set Backstop to `misMatchThreshold: 0` and updated various screenshots in https://github.com/nhsuk/nhsuk-frontend/pull/1946/commits/3b1098a124fbe1e1d8271708e52ead736a08ccfc

PR #1969, opened 2026-06-11
[v10] Add option to make tables scroll

anandamaryon1:
## Description
Tables on small screens can be tricky, but we have the responsive table to try and help. For our sortable table, we cannot include a responsive (stacked cards) option because the sorting relies on hitting the table headings, which don't exist on our responsive table. 

The sortable table header icons make the table take up slightly more horizontal space. And sorting is most helpful on data-heavy tables, which often have many columns. 

Therefore, I propose we add a scrolling container for tables.

### Shadows as scroll indicators

I tried adding a scroll indicator that wasn't a shadow, hoping to fit in better with the design system, but shadows at the left or right edges were visually intuitive and more noticeable.

We're using CSS [`@container` scroll state queries](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Conditional_rules/Container_scroll-state_queries) but a JavaScript polyfill is included for older browsers.

Read more about [CSS `scroll-state()` on the Chrome for Developers blog](https://developer.chrome.com/blog/css-scroll-state-queries).

```css
@container scroll-state(scrollable: inline) {}
@container scroll-state(scrollable: left) {}
@container scroll-state(scrollable: right) {}
```

### Caption wrapping

It also features a specific `max-width: calc(100cqw);` [container query width](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Containment/Container_queries) for table captions, making them wrap at the edge of the visible container width, preventing the captions from being cut off in the scroll overflow.

### Review app preview

* https://nhsuk-frontend-pr-1969.herokuapp.com/nhsuk-frontend/components/tables/scrolling/

<br>

Scrolling table example

## Checklist

- [x] CHANGELOG entry


colinrotherham:
@anandamaryon1 Nice work on the PR 🙌

Was it based on this excellent material? Would be good to double check we've covered everything:

* [Keyboard-Only Scrolling Areas](https://adrianroselli.com/2022/06/keyboard-only-scrolling-areas.html)
* [**Under-Engineered Responsive Tables**](https://adrianroselli.com/2020/11/under-engineered-responsive-tables.html)
* [A Responsive Accessible Table](https://adrianroselli.com/2017/11/a-responsive-accessible-table.html)
* [Hey, It’s Still OK to Use Tables](https://adrianroselli.com/2017/11/hey-its-still-ok-to-use-tables.html#ResponsiveCSS)

What do you want to do about the [existing `.nhsuk-table-container` wrapper](https://github.com/nhsuk/nhsuk-frontend/blob/cecadf472ed7fcf8311d6cf8b96e54b10b6d7101/packages/nhsuk-frontend/src/nhsuk/components/tables/_index.scss#L12-L35)?

It's something we flagged in https://github.com/nhsuk/nhsuk-frontend/pull/1156#discussion_r1993189195 but we only need one

packages/nhsuk-frontend/src/nhsuk/components/tables/_index.scss
```diff
@@ -347,4 +347,24 @@ $sortable-table-hover-colour: color.adjust(nhsuk-colour("grey-5"), $lightness: -
...
   }
+
+  // Remove bottom margin from tables nested inside a scrollable container
+  // to avoid double spacing (the container provides the margin).
+  //
+  // Pin the caption to the left of the visible scroll area so it
+  // remains readable when the table overflows. 
+  // Wrap table caption within visible scroll area with max-inline-size.
+  // (established by container-type: inline-size on .nhsuk-scrollable-container)
+  .nhsuk-scrollable-container {
```
colinrotherham: I'd suggest we use `.nhsuk-table-wrapper` as the name?

We follow the `${component}-wrapper` convention plus we have `.nhsuk-main-wrapper` too:

```console
.nhsuk-label-wrapper
.nhsuk-input-wrapper
```

Then to solve the [double-wrapper problem](https://github.com/nhsuk/nhsuk-frontend/pull/1969#issuecomment-4707337612) we can style them both the same?

```suggestion
  .nhsuk-table-wrapper,
  // Deprecated, to be removed in v11.0
  .nhsuk-table-container {
```

anandamaryon1: Yes `wrapper` is more consistent with our other things. I was thinking `container` because it literally contains it's content? Also, losing the word 'scrollable' makes it less obvious what this class does. 

I went for `scrollable-container` as a more generic thing, assuming it might be helpful in other non-table situations too. You thinking it should be table specific?

anandamaryon1:
Made a few edits:
* use `focus-visible` so that focus doesn't show when using mouse (it's quite distracting)
* switched to `overflow: auto` so that scrollbars don't show when not needed
* straighten indicator 'shadow'

packages/nhsuk-frontend/src/nhsuk/components/tables/_index.scss
colinrotherham: Container is good too. For comparison, here are our current container uses:

```scss
.nhsuk-table-container
.nhsuk-table-container-fluid
.nhsuk-width-container
.nhsuk-card__heading-container
.nhsuk-header__navigation-container
```

In which case I'd keep the existing `.nhsuk-table-container` class

At the moment it's not generic and only scrolls tables

anandamaryon1: updated to move it all into `. nhsuk-table-container`

anandamaryon1:
Updated: move all styles into `nhsuk-table-container`

packages/nhsuk-frontend/src/nhsuk/components/tables/_index.scss
```diff
@@ -17,25 +17,89 @@ $sortable-table-hover-colour: color.adjust(nhsuk-colour("grey-5"), $lightness: -
...
+  // 2. Visual scroll indicators use the background-attachment: local trick from
+  //    Lea Verou (https://lea.verou.me/2012/04/background-attachment-local/).
+  //    Gradients with 'local' attachment scroll with content (hiding shadow at edges).
+  //    Gradients with 'scroll' attachment stay fixed to container.
+  //    Left/right covers hide shadows when scrolled to start/end.
+  //    Left/right indicator lines show scroll availability.
+  // 3. Margin is removed so there isn't double spacing.
 
   .nhsuk-table-container {
+    --nhsuk-table-container-bg: #{$nhsuk-body-background-colour};
```
colinrotherham: We support IE11 yet [CSS custom properties](https://caniuse.com/css-variables) are not supported

Can we use plain old Sass variables here please?

packages/nhsuk-frontend/src/nhsuk/components/tables/_index.scss
```diff
@@ -17,25 +17,89 @@ $sortable-table-hover-colour: color.adjust(nhsuk-colour("grey-5"), $lightness: -
...
+
     overflow-x: auto;
 
-    -webkit-overflow-scrolling: touch;
-    -ms-overflow-style: -ms-autohiding-scrollbar;
+    background: // [2]
+      // Left cover — hides left shadow when scrolled to the start
+      linear-gradient(
+        to right,
+        var(--nhsuk-table-container-bg) nhsuk-spacing(2),
```
colinrotherham: Then the colours can just be written into the CSS as usual:

```suggestion
        #{$nhsuk-body-background-colour} nhsuk-spacing(2),
```

colinrotherham:
@anandamaryon1 This is all conflict-free now and the Prettier issues are resolved

For the linear gradient, it's looking like a semi-transparent solid line to me

Has something gone wrong or shall we just make it a border?

Feels like it needs to extend below the table slightly too, avoiding the clash with the bottom table cell borders

Linear gradient table overflow

anandamaryon1:
Yeah it's behind the table content because it's a background. 

If using a border, can it be made to hide and show when scrolled?

Having it hang below the table content isn't possible with it being a background I don't think, but here's a quick mockup:
image

colinrotherham:
>Having it hang below the table content isn't possible with it being a background I don't think, but here's a quick mockup:

We could add padding bottom of the container but remove it with a negative margin?

That way the gradient would appear to hang below

colinrotherham:
We're going to investigate this approach a little further

Unfortunately using a solid colour to mask a gradient doesn't work on other background colours:

<br>

Shadow on white background

Shadow on grey background

Shadow on blue background

anandamaryon1:
We'll have to go with a JavaScript based solution for this, to enable the visual scroll indicator to work on different background colours.

anandamaryon1:
And change the option to `scroll: true`

colinrotherham:
Should it be `scrolling: true` like the guidance?

anandamaryon1:
I just changed the guidance to `scroll: true` 😉

colinrotherham:
I've managed to offset the scrolling container into the gutter

Realistic shadows look far more "discoverable" but shall I roll them back to the flat style?

Still need to toggle the shadows on/off, but would be great to use [CSS `scroll-state()` one day](https://developer.chrome.com/blog/css-scroll-state-queries#stuck)

Scrolling table shadows

anandamaryon1:
Thanks for this @colinrotherham. A couple of things on review.

### Overflowing into to the gutter
I'm not sure about this. On small screens, the shadow is very hard to see because it's right on the edge, particularly on my phone which has curved edges. 

The table content overflowing into the gutter to signal it can be scrolled is a nice idea and often the best way to make something obviously scrollable, but with our gutter being so thin, and the table columns sometimes lining up such that there's a gap at the overflow point, I'm not sure it's very noticeable. 

I also wonder about other layouts, like a filters side panel, where the table content will touch it when scrolling, or when placed to the left.

### Shadow stying
Yeah I think you're right about this being more visually intuitive and noticeable. I was trying to force it into line with our styling but in this case maybe we should let it be an outlier. I might like to tweak the darkness of the reverse version, it seems less noticeable to me. (I think just bump to `0.25` from `0.15`)

### Table header wrapping
Makes sense in a way, we don't need ugly wrapping now we can make the table as wide as needed, but, in terms of showing the table content, I don't think it always helps. It means for example that columns with small numbers but long headings show very spaced out, making comparing the numbers more difficult. I see scrolling as a last resort. So, I think we can leave wrapping as default, people can still apply their own word wrap tweaks if needed.

colinrotherham:
Thanks @anandamaryon1

Feedback below:

<br>

> ### Overflowing into to the gutter
>…I also wonder about other layouts, like a filters side panel, where the table content will touch it when scrolling, or when placed to the left.

The negative offset (or overflow) was at `$nhsuk-gutter-half` (16px) so columns wouldn't touch, but we'd definitely have to sort out the focus styles so they appear inside the viewport

To test it out your thoughts here's a before and after, reducing the offset from 16px to 6px:

### Before

Full width offset using `$nhsuk-gutter-half` (16px)

Tables_scrolling_mobile (before)

### After

Partial width offset (6px) looks a bit broken

Tables_scrolling_mobile (after)

<br>
<br>

> ### Shadow stying
>…I might like to tweak the darkness of the reverse version, it seems less noticeable to me. (I think just bump to `0.25` from `0.15`)

I've pushed up a changed to use:

* `rgba($nhsuk-text-colour, 0.3)` at `opacity 0.5` (default)
* `rgba($nhsuk-text-colour, 0.3)` at `opacity 0.75` (reverse)

<br>

> ### Table header wrapping
>…So, I think we can leave wrapping as default, people can still apply their own word wrap tweaks if needed.

Wrapping disabled was an idea that came from [MOJ scrollable panel](https://design-patterns.service.justice.gov.uk/components/scrollable-pane/))

But yeah I agree. I've brought the wrapping back again

colinrotherham:
@anandamaryon1 Scroll JavaScript added (for older browsers) plus new test page:

* https://nhsuk-frontend-pr-1969.herokuapp.com/nhsuk-frontend/examples/scrolling-tables/

Modern browsers use CSS `scroll-state()` to show/hide scroll shadows

Needs some tests

anandamaryon1:
Nice, thanks @colinrotherham it's looking good.

Doing some testing and found that we've not yet accounted for tables in cards. We'll need to adjust the padding offsets to sit flush inside the cards. 

For example: 
image

Or even, keeping the table margin bottom but removing the card contend padding bottom:
image

anandamaryon1:
Found some bugs on older Safari versions.

iPhone 11, iOS, v13.3
Scrolling + sortable tables make the page overflow. (but shadow works correctly, and no overflow on non-sortable)
image

iPhone 15, iOS, v17.5
Scroll tables (any), the shadow doesn't scroll.
image

Both fixed in v18.

Slipping between the CSS/JS fallbacks maybe?

Not found anything in other browsers 👍

colinrotherham:
@anandamaryon1 Thanks for the browser testing

I've added support for both `params.card` and `params.scroll` at the same time

Can you check the margins look alright? The scrolling tables example page has been updated too:

* https://nhsuk-frontend-pr-1969.herokuapp.com/nhsuk-frontend/examples/scrolling-tables/

colinrotherham:
Changelog entry added

@anandamaryon1 Speaking to @edwardhorsford, he was keen to try flat semi-transparent shadows (again)

I've given them a try:

* https://github.com/nhsuk/nhsuk-frontend/pull/2036

With the usual review app example pages

* [**Scrolling tables full page example** (after)](https://nhsuk-frontend-pr-2036.herokuapp.com/nhsuk-frontend/examples/scrolling-tables/) [(versus before)](https://nhsuk-frontend-pr-1969.herokuapp.com/nhsuk-frontend/examples/scrolling-tables/)
* [**All table examples** (after)](https://nhsuk-frontend-pr-2036.herokuapp.com/nhsuk-frontend/components/tables/) [(versus before)](https://nhsuk-frontend-pr-1969.herokuapp.com/nhsuk-frontend/components/tables/)

Full width table with flat shadows

Full width table (reverse) with flat shadows

anandamaryon1:
> @anandamaryon1 Speaking to @edwardhorsford, he was keen to try flat semi-transparent shadows (again)


I like it, thanks both. Commented on PR.

colinrotherham (approved):

PR #1970, opened 2026-06-16
Scrollbar container with scrollbars

frankieroberto:
An alternative to #1969 - this one uses scrollbars instead of a shadow to indicate scrolling being available.

Could do both though?

Screenshot 2026-06-16 at 12 31 17

### DIFF
--- a/packages/nhsuk-frontend/src/nhsuk/components/tables/_index.scss
+++ b/packages/nhsuk-frontend/src/nhsuk/components/tables/_index.scss
@@ -34,6 +34,19 @@
     }
   }
 
+  .nhsuk-table-container::-webkit-scrollbar {
+    height: 8px;
+  }
+
+  .nhsuk-table-container::-webkit-scrollbar-track {
+    background: nhsuk-colour("grey-5");
+  }
+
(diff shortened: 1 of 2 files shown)


anandamaryon1:
Nice, seems to work more consistently than `overflow-x: scroll`, but not showing on iOS:

image


I think we'll need both the scrollbars and an indicator. For browsers that don't show the scrollbar and for cases where the table content is cut off between columns.

frankieroberto:
One idea for mobile: should the table bust out of the page container and go full-bleed to the edges of the viewpoint? That'd buy you a few more pixels of space to see the table in, and might make it clearer there's more content to see? 🤔 

(On the other hand it might make it harder to scroll vertically past the table)

anandamaryon1:
Reading this has put me off customising the scrollbar: https://ericwbailey.website/published/dont-use-custom-css-scrollbars/

Overflowing in the margin: interesting, could work in some instances but seems risky and our gutter is very small (16px)

colinrotherham:
Thanks for this @frankieroberto

I'm going to close it in favour of https://github.com/nhsuk/nhsuk-frontend/pull/1969

PR #1971, opened 2026-06-17
[v10] Update hero to support content columns

frankieroberto:
This PR adds a new `content` option for the Hero component, mainly intended to support [product pages](https://github.com/nhsuk/nhsuk-service-manual/pull/2310)

Like the footer component the `content` option supports an array of columns with:

* `heading` — Optional heading for content column
* `text` – Optional text to use within each content column
* `html` – Optional HTML to use within each content column
* `image` – Optional image to use within each content column
* `width` – Width of each content column
* `classes` – Classes to add to the content column

Thoughts welcome. Still tbc is deciding whether or not the second column (usually containing an illustration) should be hidden on mobile or stacked beneath - or if both should be options. (existing product pages across NHS and GOVUK do a mix)


anandamaryon1:
Secondary works, but in terms of it being a layout thing, I'd usually call that an `aside` but then that is also an HTML tag which maybe isn't the right one. 

I think we'll need to allow both hiding and stacking. Eg. in the current design system homepage, we'd certainly not hide the 'whats new' box on mobile. 
image

frankieroberto:
@anandamaryon1 yeah good point - we should allow both hiding and stacking. Even if it's _just_ an illustration on the right, there’s an argument that it's still nice to show it mobile - although it does then take up more space.

'Aside' could work as an alternarive to secondary. Another option @colinrotherham suggested was to follow the convention from the footer and use an array so you can have as many columns as you like. Something like this:

```njk
{{ hero({
  content: [
    {
      width: "two-thirds",
      heading: "My heading",
      html: firstColumnHtml
    },
    {
      width: "one-third",
      html: secondColumnHtml
    }  
  ]
}) }}
```

More flexible - but also more complicated and perhaps harder to make backwards-compatible?

anandamaryon1:
Array of columns is logical and consistent, but I can't see us using more than two columns in a hero. 

I'd vote for `asideHTML` just because secondary sounds to me like backup html, for an error message or something.

colinrotherham:
Don't forget the array of columns is automatically split into rows in the footer

E.g. You can pass 6 columns into the footer but 2 rows of 3 columns will be generated

anandamaryon1:
Ahh didn't realise this. @colinrotherham 

Is it possible to make this non-breaking if going for an array of columns?

colinrotherham:
@frankieroberto I've added support for `content` with multi-column support

How's this? Not sure on `headingSize` versus `heading.size` etc

```njk
{{ hero({
  content: [
    {
      heading: 'This is a header for the product or service',
      headingSize: 'l',
      html: primaryHtml,
      classes: 'nhsuk-u-margin-top-5'
    },
    {
      html: secondaryHtml
    }
  ]
}) }}
```

anandamaryon1:
Nice 👍 

I suppose `heading.size` (and then `heading.text`) is more consistent with other components?

colinrotherham:
>I suppose `heading.size` (and then `heading.text`) is more consistent with other components?

@anandamaryon1 That would be brilliant but unfortunately it needs another GitHub issue

Talking to @paulrobertlloyd we think an internal `heading()` component would be best

Perhaps we could unify all the components to use `heading.text`, `heading.html` etc?

### Do and don't list

```console
title
```

### Error summary

```console
titleText
titleHtml
```

### Footer

```console
navigation.title
meta.visuallyHiddenTitle
```

### Hero

```console
heading
headingLevel
headingSize
headingClasses
```

### Notification banner

```console
titleId
titleText
titleHtml
titleHeadingLevel
```

### Panel

```console
titleText
titleSize
headingLevel
```

### Tabs

```console
title
```

### Task list

```console
items[].title.text
items[].title.html
```

anandamaryon1:
Oooh yeah we have a few different approaches there! 

Yes OK seems like it'd need to be a wider thing and in a new v11 PR.

colinrotherham:
>Yes OK seems like it'd need to be a wider thing and in a new v11 PR.

If we're going to rename these we should deprecate them first in v10?

They can all be supported in a backwards compatible way

I've split out a heading and caption component here as a start:

* https://github.com/nhsuk/nhsuk-frontend/pull/1985

frankieroberto:
> @frankieroberto I've added support for `content` with multi-column support
> 
> How's this? Not sure on `headingSize` versus `heading.size` etc
> 
> ```
> {{ hero({
>   content: [
>     {
>       heading: 'This is a header for the product or service',
>       headingSize: 'l',
>       html: primaryHtml,
>       classes: 'nhsuk-u-margin-top-5'
>     },
>     {
>       html: secondaryHtml
>     }
>   ]
> }) }}
> ```

This seems fine but how would you specify grid widths of each column? We don’t have a pattern for that, I don’t think?

colinrotherham:
>This seems fine but how would you specify grid widths of each column? We don’t have a pattern for that, I don’t think?

Same as the footer, stick `width: "two-thirds"` etc in the column options

(But unlike the footer I've locked the columns-per-row to 2)

colinrotherham:
@frankieroberto Thoughts on switching to `heading.text`, `heading.html` and `heading.level` etc?

We can either deprecate the current `heading` and `headingLevel` or always support both for convenience?

frankieroberto:
@colinrotherham I like the idea - might be good to standardise on using objects for anything that can accept `html`, `text` and `classes`? There’s a few other components this applies to too (eg Panel).

Don't have a strong view on deprecation vs always supporting both though.

colinrotherham (approved):

colinrotherham:
@frankieroberto Forgot to say, @anandamaryon1 noticed the SVG overflows the hero width

Should we:

* Add support for `content.image` to populate the image component?
* Add support for images with `one-third`, `two-thirds` max-width etc ([see review comments](https://github.com/nhsuk/nhsuk-service-manual/pull/2565/changes/BASE..2cb1ff2d853d222739debade65c20960cfaddea8#diff-ea1c137aca18ab6983dd778df623c8f1ef96032749e02d28853fcf327c2e724c))

Teams can customise the hero how they like, but would be nice to [support our home page](https://github.com/nhsuk/nhsuk-service-manual/pull/2565) out of the box

<br>

Image overflow

packages/nhsuk-frontend/src/nhsuk/components/hero/fixtures.mjs
```diff
@@ -73,19 +104,34 @@ const fixtures = {
...
+          `,
+          classes: 'nhsuk-u-margin-top-5'
+        },
+        {
+          html: outdent`
+            <svg xmlns="http://www.w3.org/2000/svg" width="680" height="300">
+              <circle cx="200" cy="150" r="80" fill="white"/>
+              <circle cx="280" cy="80" r="40" fill="white"/>
+              <circle cx="120" cy="80" r="40" fill="white"/>
+            </svg>
```
colinrotherham: Can we use a less well known example? 🙏

frankieroberto: 🐭 😂

anandamaryon1:
I've added some basic styling for images or SVGs within `.nhsuk-hero__content`. Now reading your comment above again @colinrotherham and the way I've added them probably isn't right and should be more specific and via Nunjucks, but not sure how this would work with SVGs/ supporting both `img` and `svg` tags?

In terms of sizes, I'm not sure how we can do this nicely. It feels a bit image-specific. And then I start thinking about using `figure` and supplying different images for art direction and my head spins :D 

PS. I'll make a separate PR for our image component for a frameless option, which removes the border and background.

packages/nhsuk-frontend/src/nhsuk/components/hero/_index.scss
```diff
@@ -27,80 +42,131 @@
...
+      display: block;
+
+      max-width: 100%;
+      height: auto;
+      margin-right: auto;
+      margin-left: auto;
+      padding-top: 0;
+    }
+
+    &:has(svg, img) {
```
colinrotherham: Are we alright with these [styles being optional due to `:has()` browser support](https://caniuse.com/css-has)?

packages/nhsuk-frontend/src/nhsuk/components/hero/_index.scss
```diff
@@ -27,80 +42,131 @@
...
+      padding-top: 0; // [17]
+
+      @media #{nhsuk-from-breakpoint(tablet)} {
+        @include nhsuk-responsive-padding(8, ("top", "bottom"));
+      }
```
colinrotherham: Rather than duplicate the default padding, can we temporarily suppress it?

```suggestion
      @media #{nhsuk-until-breakpoint(tablet)} {
        padding-top: 0; // [17]
      }
```

packages/nhsuk-frontend/src/nhsuk/components/hero/_index.scss
colinrotherham: It might be safer to use negative margins on the image to offset the parent padding?

(We added support for negative values recently)

```scss
@include nhsuk-responsive-margin(-8, ("top", "bottom"))

packages/nhsuk-frontend/src/nhsuk/components/hero/_index.scss
anandamaryon1: Forgot we had `until` 😅

colinrotherham (approved):

anandamaryon1 (approved):

PR #1984, opened 2026-06-23
[v10] Extend global inline code styles

frankieroberto:
Discovered an issue in https://github.com/nhsuk/nhsuk-service-manual/pull/2585 that `<code>` elements within a summary list don't get the red-on-white colour.

This fixes that. Also extends the inline code style to where they are within `<th>` elements.

| Before | After |
| ------ | ------|
| Screenshot 2026-06-23 at 14 54 03 | Screenshot 2026-06-23 at 14 53 31 |

## Checklist

- [x] CHANGELOG entry


colinrotherham (commented):
Thanks @frankieroberto 

Cache retrieval is failing again (even though it was just saved). Might have to run the actions a few more times—they must be having a slow day again

Am I right in thinking the code block page examples are duplicated?

Once for normal, once for reverse

frankieroberto:
@colinrotherham duplicated the example into the reverse section now:

Screenshot 2026-06-23 at 15 08 03

packages/nhsuk-frontend-review/src/examples/code-blocks.njk
```diff
@@ -161,6 +161,19 @@
...
+          rows: [
+            {
+              key: { html: "<code>nhsuk-u-margin-0</code>", classes: "nhsuk-u-width-one-half" },
+              value: { text: "Responsive margin on all sides" }
+            },
+            {
+              key: { html: "<code>nhsuk-u-margin-top-0</code>" },
+              value: { text: "Responsive margin on the top" }
+            }
+          ] }) }}
```
colinrotherham: Let's keep our usual code style?

```suggestion
          {{ summaryList({
            rows: [
              {
                key: {
                  html: "<code>nhsuk-u-margin-0</code>",
                  classes: "nhsuk-u-width-one-half"
                },
                value: {
                  text: "Responsive margin on all sides"
                }
              },
              {
                key: {
                  html: "<code>nhsuk-u-margin-top-0</code>"
                },
                value: {
                  text: "Responsive margin on the top"
                }
              }
            ]
          }) }}
```

anandamaryon1:
Nice thanks for this @frankieroberto 

Should the reverse styles be using `nhsuk-code--reverse`? 

eg. top one here:
Screenshot 2026-06-23 at 15 45 17

frankieroberto:
@anandamaryon1 

> Nice thanks for this @frankieroberto
> 
> Should the reverse styles be using `nhsuk-code--reverse`?

Ah good spot. Done in f942a1796c73b5164140b51b96038a8ffb854225. It needs the `nhsuk-code--inline` class adding too for some reason.

Arguably the inverse style could be applied automatically, but that’s a whole other can of worms so maybe leave that for now? (The govuk design system team are looking at this too)

packages/nhsuk-frontend-review/src/examples/code-blocks.njk
```diff
@@ -312,6 +335,19 @@
...
+          rows: [
+            {
+              key: { html: "<code class=\"nhsuk-code--inline nhsuk-code--reverse\">nhsuk-u-margin-0</code>", classes: "nhsuk-u-width-one-half" },
+              value: { text: "Responsive margin on all sides" }
+            },
+            {
+              key: { html: "<code class=\"nhsuk-code--inline nhsuk-code--reverse\">nhsuk-u-margin-top-0</code>" },
+              value: { text: "Responsive margin on the top" }
+            }
+          ] }) }}
```
colinrotherham: Sorry I only fixed the indentation on one

frankieroberto: fixed

packages/nhsuk-frontend-review/src/examples/code-blocks.njk
```diff
@@ -312,6 +335,29 @@
 
           <p class="nhsuk-body">This change was introduced in <a class="nhsuk-link nhsuk-link--reverse nhsuk-link--no-visited-state" href="https://github.com/nhsuk/nhsuk-frontend/pull/1556">pull request #1556: Uplift GOV.UK Frontend file upload component</a>.</p>
 
+          <h2>Override classes</h2>
+          {{ summaryList({
+            rows: [
+              {
+                key: {
+                  html: "<code class=\"nhsuk-code--inline nhsuk-code--inverse\">nhsuk-u-margin-0</code>",
```
colinrotherham: Super minor, but we use single quotes when double quotes are within

```suggestion
                  html: '<code class="nhsuk-code--inline nhsuk-code--inverse">nhsuk-u-margin-0</code>',
```

packages/nhsuk-frontend-review/src/examples/code-blocks.njk
```diff
@@ -312,6 +335,29 @@
...
+                  classes: "nhsuk-u-width-one-half"
+                },
+                value: {
+                  text: "Responsive margin on all sides"
+                }
+              },
+              {
+                key: {
+                  html: "<code class=\"nhsuk-code--inline nhsuk-code--inverse\">nhsuk-u-margin-top-0</code>"
```
colinrotherham:
```suggestion
                  html: '<code class="nhsuk-code--inline nhsuk-code--inverse">nhsuk-u-margin-top-0</code>'
```

colinrotherham (commented):
Thanks @frankieroberto 🙌

Just needs a changelog entry?

Probably adjust the PR title to say we're adding global styles (I think we say this?) for inline code elements within table heading, description term and description details elements?

i.e. Whilst it does apply to the summary list these are global styles

packages/nhsuk-frontend-review/src/examples/code-blocks.njk
frankieroberto: Ah damn I initially did this but then changed it to use the backslash as I thought that was the style! 🤦‍♂️

colinrotherham:
>Ah good spot. Done in https://github.com/nhsuk/nhsuk-frontend/commit/f942a1796c73b5164140b51b96038a8ffb854225. It needs the `nhsuk-code--inline` class adding too for some reason.

Can I have your thoughts on this?

i.e. Should we make `<code>` global (no parent restrictions) and remove `.nhsuk-code--inline`?

colinrotherham (approved):

frankieroberto:
@colinrotherham:

> Can I have your thoughts on this?
> 
> i.e. Should we make `<code>` global (no parent restrictions) and remove `.nhsuk-code--inline`?

Feels like it could be one for a follow-up PR as we consider global styles more generally? 🤷‍♂️

PR #1985, opened 2026-06-25
[v10] Add heading and caption components

colinrotherham:
## Description

This PR adds heading and caption components and rolls them out to all components

* Captions can be placed relative to the heading – `"before"`, `"after"`, `"start"` or `"end"`
* Captions inherit heading size or custom `size` can be set – `"m"`, `"l"` or `"xl"`

**Note:** NHS.UK frontend supports `.nhsuk-caption--bottom` but it needed fixing for `"after"` placement

### Heading with caption

```njk
{{ heading({
  text: "What is your home address?",
  caption: "About you",
  size: "l"
}) }}
```

Heading component with caption

### Heading with caption (custom placement)

```njk
{{ heading({
  text: "What is your home address?",
  caption: {
    text: "About you",
    placement: "after"
  },
  size: "l"
}) }}
```

Heading component with caption placed after

## Checklist

- [ ] CHANGELOG entry


packages/nhsuk-frontend/src/nhsuk/components/caption/README.md
```diff
@@ -0,0 +1,9 @@
+# Caption
+
+## Installation
+
+See the [main README quick start guide](https://github.com/nhsuk/nhsuk-frontend#quick-start) for how to install this component.
+
+## Guidance and examples
+
+The caption component is used by the heading component and is not currently documented by the [design system in the NHS digital service manual](https://service-manual.nhs.uk/design-system).
```
anandamaryon1: https://service-manual.nhs.uk/design-system/styles/typography#headings-with-captions

Is related documentation for it.

packages/nhsuk-frontend/src/nhsuk/components/heading/macro-options.mjs
```diff
@@ -0,0 +1,102 @@
...
+    isComponent: true
+  },
+  size: {
+    type: 'string',
+    required: false,
+    description:
+      'Size of the heading – `"xxs"`, `"xs"`, `"s"`, `"m"`, `"l"` or `"xl"`.',
+    released: '10.6.0'
+  },
+  sizes: {
```
anandamaryon1, line 50: Is this needed for internal use only? Might be a bit confusing if output in design system?

packages/nhsuk-frontend/src/nhsuk/components/heading/README.md
```diff
@@ -0,0 +1,9 @@
+# Heading
+
+## Installation
+
+See the [main README quick start guide](https://github.com/nhsuk/nhsuk-frontend#quick-start) for how to install this component.
+
+## Guidance and examples
+
+The heading component is used by other components and is not currently documented by the [design system in the NHS digital service manual](https://service-manual.nhs.uk/design-system).
```
anandamaryon1: Related guidance: https://service-manual.nhs.uk/design-system/styles/typography#headings

anandamaryon1:
Nice. Will we need to remember to swap over examples to use this in the design system at or after v11?

packages/nhsuk-frontend/src/nhsuk/components/heading/macro-options.mjs
colinrotherham: Same as the `code` and `legend` components, it's just there for our own documentation

But let's say we upgrade the hero to have a documentation page, it'll be needed then

### Components without documentation pages

We have a special case for label and legend components though

Their Nunjucks macro options are displayed for the component they belong to. You'll see this line in the service manual code to make sure we don't link to a separate page for them:

```mjs
const hasComponentPage = !['label', 'legend'].includes(option.slug)
```

If we need to one day, we can add `heading` to that list

packages/nhsuk-frontend/src/nhsuk/components/heading/README.md
colinrotherham: Thanks, both README files updated ([see diff](https://github.com/nhsuk/nhsuk-frontend/compare/59aa29a62393fc7f33be08840d51f575d8dd2309..bc80ff1b3dc113f0e3f107fce7ec758cfe8cd742))

packages/nhsuk-frontend/src/nhsuk/components/caption/README.md
colinrotherham: Thanks, both README files updated ([see diff](https://github.com/nhsuk/nhsuk-frontend/compare/59aa29a62393fc7f33be08840d51f575d8dd2309..bc80ff1b3dc113f0e3f107fce7ec758cfe8cd742))

packages/nhsuk-frontend/src/nhsuk/components/panel/template.njk
```diff
@@ -11,35 +12,30 @@
...
+  {{ heading({
+    text: params.titleText,
+    html: params.titleHtml,
+    level: params.headingLevel,
+    size: params.titleSize,
+    sizes: ["m", "l", "xl"],
+    className: classNamesTitle,
+    classPrefix: classNamesTitle ~ "--",
+    classes: params.titleClasses
+  }) | trim | indent(2) }}
```
colinrotherham, line 29: Here's an example of how we're using `heading()` to:

* Override `nhsuk-heading` to `nhsuk-panel__title`
* Override `nhsuk-heading-` prefix to `nhsuk-panel__title--`
* Limit sizes to `"m"`, `"l"` and `"xl"`

paulrobertlloyd:
Should `before` and `after` placements use paragraphs, not spans, as they are block-level, not inline?

paulrobertlloyd:
Should `start` and `end` include `: ` in visibly hidden text? Otherwise when read aloud, spans of text are going to make no sense:

🗣️ about you what is your home address?

versus:

🗣️ about you [slight pause] what is your home address?

colinrotherham:
>Should before and after placements use paragraphs, not spans, as they are block-level, not inline?

@paulrobertlloyd Consider this PR "as-is" according to our current guidance:

* [Styles → Typography: **Headings**](https://service-manual.nhs.uk/design-system/styles/typography#headings)
* [Styles → Typography: **Headings with captions**](https://service-manual.nhs.uk/design-system/styles/typography#headings-with-captions)

We copy GOV.UK Frontend which uses `<span>` https://github.com/alphagov/govuk-frontend/issues/1505

I do agree though, and perhaps a `visuallyHiddenText` option would help?

For a fair comparison, HMRC switched to:

* [`<h2>` captions before the `<h1>`](https://design.tax.service.gov.uk/hmrc-design-patterns/page-heading/#section-heading) in the guidance
* [`<p>` captions when after the `<h1>`](https://github.com/hmrc/hmrc-frontend/blob/e32e03cd2e2cf461d3b50752d0ec611966f9015e/src/components/page-heading/template.njk#L4) in their page heading component

Notice how the "before" examples add visually hidden text?

```html
<h2 class="govuk-caption-l hmrc-caption-l">
  <span class="govuk-visually-hidden">This section is </span>Your details
</h2>
```

---

**Update:** I've included support for `caption.element` should `<p>` be needed

paulrobertlloyd:
If this PR is just componetising of existing markup, happy to approve.

paulrobertlloyd (approved):

colinrotherham:
>If this PR is just componetising of existing markup, happy to approve.

It was until https://github.com/nhsuk/nhsuk-frontend/pull/1985/changes/16c8231e0d93c77ad3402e5f7d949614653fb74b for your suggestion to deprecate `nhsuk-caption--bottom`

We're now requiring `<span>` around the text (any element will do) for the auto-positioning to work

Can you check you're happy with the approach or we can continue with `--bottom`?

**Update:** Now using ~`<strong>`~ `<span>`

---

From the changelog entry:

> ### Remove heading caption bottom class modifier
> We've updated the heading caption to automatically adjust its margin so that the `nhsuk-caption--bottom` modifier class can be removed. To support this, the heading text must be wrapped in `<span> </span>` tags if the caption is nested within the heading.
> 
> Update your HTML markup using the [headings with captions examples in the NHS digital service manual](https://service-manual.nhs.uk/design-system/styles/typography#headings-with-captions) as follows:
> 
> - wrap the heading text in `<span> </span>` tags if the caption is nested
> - remove the caption `nhsuk-caption--bottom` modifier class
> 
> ```patch
>   <h1 class="nhsuk-heading-l">
> -   What is your home address?
> +   <span>What is your home address?</span>
> -   <span class="nhsuk-caption-l nhsuk-caption--bottom">
> +   <span class="nhsuk-caption-l">
>       About you
>     </span>
>   </h1>
> ```
> 
> The class name `nhsuk-caption--bottom` is deprecated and will be removed in a future release.
> 
> This change was introduced in [pull request #1985: Add heading and caption components](https://github.com/nhsuk/nhsuk-frontend/pull/1985).

anandamaryon1:
Thanks for this @colinrotherham.

I feel that caption classes should be able to be mixed and matched with headings, eg. `nhsuk-heading-l` paired with `nhsuk-caption-l` to give designers freedom to alter based on the page's visual hierarchy and the content of the caption. 

I prefer `<span>` to `<strong>`, because strong makes me assume the caption would be bold text, which it isn't. Allowing this element to be swapped out by users should allow for cases where people need it as a tagged as heading etc. 

I do think that it'd be helpful to align our wording on heading/title and how it's used, eg. `headingSize' vs `heading.size`. I see we use `title` a lot but given this is a heading component, and in HTML these should always(?) be heading elements, we should go for `heading.`. 

This example makes a good case for using heading over title:

**Panel:** 
```
titleText
titleSize
headingLevel
```

I think we should deprecate names rather than support both longer term, to avoid the mess and confusion of people having a mix of both in their projects.

colinrotherham:
@anandamaryon1 Thanks for confirming

Are we correct in supporting all four caption positions?

>Placement of the caption relative to the heading – `"before"`, `"after"`, `"start"` or `"end"`. Defaults to `"start"`.

Bearing in mind that `"after"` didn't actually work (incorrect margins) before this PR fixed it

anandamaryon1:
> Are we correct in supporting all four caption positions?

👍

colinrotherham:
Just to follow up:

* Headings with different caption sizes are now supported `"before"`, `"after"`, `"start"` or `"end"`
* We've kept the heading HTML unchanged for now (no `<strong>` or `<span>` text wrapper)
* We've kept the heading caption `--bottom` modifier

### Heading caption `--bottom` modifier

Why is the modifier necessary?

When multiple headings are stacked together, it identifies which heading the caption belongs to

Without it, this happens:

Heading caption bottom modifier class

Issue #1153, opened 2025-02-21
REQ: Utility classes scoped to breakpoints


paulmsmith:
Utility classes are very useful, they save us needing to write new scss/css selectors to tweak spacing but by default they apply across all 'breakpoints'.

Often as components going from being stacked on small viewports to larger viewports, it would be helpful to be able to declaratively adjust that spacing. 

For example:  `.nhsuk-u-margin-bottom-0--from-desktop` or `.nhsuk-u-margin-bottom-0--until-desktop`

paulmsmith:
As they'd generate a bunch of css perhaps they could be optionally configured in settings? So they are only outputted to css when needed.

edwardhorsford:
I've certainly come across this _need_ before - though have typically made a custom class then and used the media query mixins to do what I needed.

A risk with these might be that they don't get used?

paulmsmith:
> A risk with these might be that they don't get used?

Yeh, I thought about that. Maybe a Sass setting variable `$nhsuk-responsive_utilites` or something would prevent them being outputted to the compiled css by default.

frankieroberto:
Also the `-from-xxx` suffixes have the potential for ambiguity and confusion (or is that just me?). Would `-on-xxx` be clearer? or `-xxx-only` or `-xxx-and-yyy`?

edwardhorsford:
I think it would probably be better to have grids working from tablet rather than desktop - but it's also potentially an annoying breaking change. As I understand it it was likely done as it suited what the NHS.UK website needed.

paulmsmith:
I was originally referring to some of the utility classes and making them scoped to breakpoints such as `nhsuk-u-margin-bottom-0--from-desktop`

edwardhorsford:
Good point @paulmsmith. @frankieroberto perhaps make a new issue for the point you raise?

PR #1987, opened 2026-06-29
Add `nhsuk-u-display-none` override classes for breakpoints

frankieroberto:
This PR adds new `nhsuk-u-display-none-*` override classes for breakpoints

The changes are related to https://github.com/nhsuk/nhsuk-frontend/issues/1153 and were upstreamed from NHS app frontend at [`utility/_display.scss`](https://github.com/nhsuk/nhsapp-frontend/blob/main/src/utilities/_display.scss)

### Mobile

* `nhsuk-u-display-none-until-mobile` – hide content until mobile breakpoint
* `nhsuk-u-display-none-from-mobile` – hide content from mobile breakpoint (and above)

### Tablet

* `nhsuk-u-display-none-until-tablet` – hide content until tablet breakpoint
* `nhsuk-u-display-none-from-tablet` – hide content from tablet breakpoint (and above)

### Desktop

* `nhsuk-u-display-none-until-desktop` – hide content until desktop breakpoint
* `nhsuk-u-display-none-from-desktop` – hide content from desktop breakpoint (and above)

### Large desktop

* `nhsuk-u-display-none-until-large-desktop` – hide content until large desktop breakpoint
* `nhsuk-u-display-none-from-large-desktop` – hide content from large desktop breakpoint (and above)

## Checklist

- [x] CHANGELOG entry

### DIFF
--- a/packages/nhsuk-frontend/src/nhsuk/core/utilities/_display.scss
+++ b/packages/nhsuk-frontend/src/nhsuk/core/utilities/_display.scss
@@ -1,3 +1,4 @@
+@use "../../core/settings" as *;
 @use "../../core/tools" as *;
 
 ////
@@ -24,15 +25,17 @@
   display: none !important;
 }
 
-.nhsuk-u-display-none-from-tablet {
-  @include nhsuk-media-query($from: tablet) {
-    display: none !important;
+@each $name, $value in $nhsuk-breakpoints {
+  @include nhsuk-media-query($until: $name) {
+    .nhsuk-u-display-none-until-#{$name} {
+      display: none !important;
+    }
   }
-}
 
-.nhsuk-u-display-none-on-mobile {
-  @include nhsuk-media-query($until: tablet) {
-    display: none !important;
+  @include nhsuk-media-query($from: $name) {
+    .nhsuk-u-display-none-from-#{$name} {
+      display: none !important;
+    }
   }
 }
 
--- a/packages/nhsuk-frontend/src/nhsuk/core/utilities/_display.scss
+++ b/packages/nhsuk-frontend/src/nhsuk/core/utilities/_display.scss
@@ -24,6 +24,18 @@
   display: none !important;
 }
 
+.nhsuk-u-display-none-from-tablet {
+  @include nhsuk-media-query($from: tablet) {
+    display: none !important;
+  }
+}
+
+.nhsuk-u-display-none-on-mobile {
+  @include nhsuk-media-query($until: tablet) {
+    display: none !important;
+  }
+}
+
 .nhsuk-u-display-none-print {
   @media print {
     display: none !important;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -335,6 +335,7 @@ This change was introduced in [pull request #1998: Add `compact` option for tabl
 
 - [#1942: Update `nhsuk-spacing()` to add missing `$adjustment` and `$important` params](https://github.com/nhsuk/nhsuk-frontend/pull/1942)
 - [#1986: Update `nhsuk-font-monospace` font size to better match body text](https://github.com/nhsuk/nhsuk-frontend/pull/1986/changes)
+- [#1987: Add `nhsuk-u-display-none` override classes for breakpoints](https://github.com/nhsuk/nhsuk-frontend/pull/1987)
 - [#1996: Update hover colour for task list and pagination](https://github.com/nhsuk/nhsuk-frontend/pull/1996)
 - [#1997: Preserve conditionally revealed content margins when JavaScript is not supported](https://github.com/nhsuk/nhsuk-frontend/pull/1997)
 - [#2000: Update responsive table row margin and border width](https://github.com/nhsuk/nhsuk-frontend/pull/2000)


davidhunter08:
@frankieroberto they are only used on the [NHS App web header](https://design-system.nhsapp.service.nhs.uk/components/header/).

> We extend the [NHS design system header](https://service-manual.nhs.uk/design-system/components/header) with added `nhsapp-u-hide-from-tablet` and `nhsapp-u-hide-until-tablet` modifier classes to show and hide nav items.

packages/nhsuk-frontend/src/nhsuk/core/utilities/_display.scss
```diff
@@ -24,6 +24,18 @@
   display: none !important;
 }
 
+.nhsuk-u-display-none-from-tablet {
+  @include nhsuk-media-query($from: tablet) {
+    display: none !important;
+  }
+}
+
+.nhsuk-u-display-none-on-mobile {
```
colinrotherham: Thanks for adding these @frankieroberto 🙌

Does the suffix `-on-mobile` mean it's visible below mobile width?

```css
.nhsuk-u-display-none-on-mobile {
```

This is what the `-until` convention solves, as our mobile breakpoint only kicks in at 320px

### Until mobile

We stack the prefix/suffix until mobile width is reached

Text input with prefix and suffix at 224px watch breakpoint

### From mobile

We move the prefix/suffix inline from mobile width and above

Text input with prefix and suffix at 320px mobile breakpoint

<br>
<br>

I'd much prefer to use `-from-` (in v11) or its counterpart `-until` like:

```suggestion
.nhsuk-u-display-none-until-tablet {
```

---

I've dug out these [original comments from v11 regarding `-until-XXX`](https://github.com/nhsuk/nhsuk-frontend/pull/1296#issuecomment-3606060754)

@anandamaryon1 was open to `-from-` (and maybe `-until-`) but we removed them as redundant

>On the naming, I think naming from mobile-first makes sense (it's been long enough!). Part of me wants to say `-always` should be `-from-mobile` but then we'd have to have all of them being `from` or `until`.

Following our Sass API the matching terminology would be:

* `-at-XXX` for exact breakpoint definitions (mobile only, tablet only etc)
* `-from-XXX` for given breakpoint width and up (mobile width and up, tablet width and up etc)
* `-until-XXX` when lower than the given breakpoint (e.g. lower than tablet width, AKA until tablet)

Some examples are shown in our GOV.UK Frontend roots:

https://github.com/nhsuk/nhsuk-frontend/blob/d49bb29e997b93aa52c8911f04b6df24b04191a6/packages/nhsuk-frontend/src/nhsuk/core/tools/_grid.scss#L47-L50

https://github.com/nhsuk/nhsuk-frontend/blob/d49bb29e997b93aa52c8911f04b6df24b04191a6/CHANGELOG.md#L2072-L2079

frankieroberto: @colinrotherham ah ok - I forget that there's a size class _below_ mobile! 🤦 

I guess that means this should be `nhsuk-u-display-none-until-tablet`? Although it's a bit of a double negative. Could be `nhsuk-u-display-from-tablet-only`? 🤷 

(Or we could not add these at all, if they're only used in the NHS app header - maybe there's a way to avoid needing them there?)

frankieroberto:
> @frankieroberto they are only used on the [NHS App web header](https://design-system.nhsapp.service.nhs.uk/components/header/).
> 
> > We extend the [NHS design system header](https://service-manual.nhs.uk/design-system/components/header) with added `nhsapp-u-hide-from-tablet` and `nhsapp-u-hide-until-tablet` modifier classes to show and hide nav items.

@davidhunter08 Aha, thanks! I'll have a dig into how it's used there.

packages/nhsuk-frontend/src/nhsuk/core/utilities/_display.scss
colinrotherham: It's hard isn't it 😩

The problem with using `-from-` is knowing what `display` value to restore:

```css
.nhsuk-u-display-block-from-tablet
.nhsuk-u-display-flex-from-tablet
.nhsuk-u-display-inline-from-tablet
.nhsuk-u-display-inline-block-from-tablet
.nhsuk-u-display-inline-flex-from-tablet
.nhsuk-u-display-table-from-tablet
.nhsuk-u-display-table-row-from-tablet
.nhsuk-u-display-table-cell-from-tablet
```

But since we have `[not-]supported-hidden` and `visually-hidden`, how about simply `hidden`?

```css
.nhsuk-u-hidden-until-tablet
.nhsuk-u-hidden-from-tablet
```

colinrotherham: @frankieroberto @anandamaryon1 What shall we do with this PR?

### Add new utility classes

```scss
.nhsuk-u-hidden-until-tablet
.nhsuk-u-hidden-from-tablet
```

### Prefer teams to use the Sass API

For example, using the new `nhsuk-until-breakpoint()` and `nhsuk-from-breakpoint()` mixins:

```scss
.nhsuk-u-hidden-until-tablet {
  @media #{nhsuk-until-breakpoint(tablet)} {
    display: none !important;
  }
}

.nhsuk-u-hidden-from-tablet {
  @media #{nhsuk-from-breakpoint(tablet)} {
    display: none !important;
  }
}
```

Or our existing `nhsuk-media-query()` mixin:

```scss
.nhsuk-u-hidden-until-tablet {
  @include nhsuk-media-query($until: tablet)
    display: none !important;
  }
}

.nhsuk-u-hidden-from-tablet {
  @include nhsuk-media-query($from: tablet)
    display: none !important;
  }
}
```

anandamaryon1: We have visibility override classes already, so this pairs with that I suppose. One small advantage of having a class rather than just using a media query is the class is visible in the markup.

@frankieroberto did you find any more from app side on usage of this? 

I guess we'd want to caution this in the override page too, eg. "try not to hide things from users just because design responsive is hard"

colinrotherham: I've added a loop to include a from/until `display-none` override class per breakpoint

These will be great for sortable tables

E.g. Conditionally shortening a table header from "Current charge" to just "Current" on mobile

Screenshot 2026-07-17 at 17 11 08

colinrotherham (approved):

PR #1989, opened 2026-07-01
[v10] Add support for card and summary list actions as buttons

### LINKED ISSUE #1665: Feature suggestions: support call with summary list and rendering individual summary list rows
# Proposal

Extend the summary list component to support `call`, letting users provide html for individual rows, possibly including a new summaryListRow macro for making individual rows.

Syntax example: 

```nunjucks
{% call summaryList() %}
  {{ fullNameRow }}
  {{ dateOfBirthRow }}
{% endcall %}
```

This would then allow each row to be defined (and possibly stored) separately:

```nunjucks
{% set fullNameRow %}
  {{ summaryListRow({
    key: {
      text: "Full name"
    },
    value: {
      text: data.participant.fullName
    }
  }) }}
{% endset %}

{% set dateOfBirthRow %}

  {% set dob %}
    {{ data.participant.dateOfBirth | formatDate }}<br>
    <span class="nhsuk-hint">({{ data.participant.dateOfBirth | formatRelativeDate(true) }} old)</span>
  {% endset %}
  
  {{ summaryListRow({
    key: {
      text: "Date of birth"
    },
    value: {
      html: dob
    },
    actions: {
      items: [{
        text: "Change",
        href: "#",
        visuallyHiddenText: "date of birth"
      }]
    }
  }) }}
{% endset %}
```

## What is the issue?

In my service, we've got multiple summary lists that share individual rows. Several summary lists include someone's date of birth or name, whilst other rows are unique. Ideally you would use a common include for shared bits, but with the current summary list you can't do this. There's no easy way to abstract each row so that it can be shared. They could be one large summary list with dynamic logic, but this then means distinct summary lists are forced to be in one template.

Rows may often be quite simple - just a key and a value. But if they have more complex logic for display (such as my date of birth example), there can be quite a bit of template code / Nunjucks alongside used to produce the value. 

One option would be for this data to be provided by routes - but you can't provide all the data for the row, you  have to provide the key, value, and actions separately. It's also often not ideal for routes to be generating html - this is something Nunjucks is good at.

## Example implementation

I made a patched [summary list component](https://github.com/NHSDigital/manage-breast-screening-prototype/blob/main/app/views/_components/summary-list/summary-list-template.njk) - that supports call, and a [separate component for producing rows](https://github.com/NHSDigital/manage-breast-screening-prototype/blob/main/app/views/_components/summary-list/summary-list-row-template.njk).

One catch is that the current css/markup and how it deals with actions. The current component iterates through the rows array to work out if any row has actions, and if not, applies `nhsuk-summary-list__row--no-actions` to each row. This isn't compatible with being provided the entire html for the row, and it's not ideal for the markup of each row to depend on the markup of other rows.

In my prototype, I worked around this by just assuming other rows have actions unless told otherwise - but ideally we might refactor the css to not need this class at this level.

One solution might to move this class to the parent wrapper and not the individual rows - and then have the summaryList component inspect the the html of the provided rows. Or perhaps just make it a param you need to set if using call.

### LINKED ISSUE #1988: Feature suggestion: support buttons within summary list actions
For our service, we sometimes want to take direct action from a summary list. In the screenshot below, 'Undo deferral' reverts a status on our service. 

Image

Because it's making a change it needs to POST to the server. This is not possible with the summary list macro, as it forces that each action item is wrapped in an anchor. Ideally this would be more flexible so it could have buttons or arbitrary html.

For us to think about:
* could this work like the button component where you can specify an element?
* would we need new button-link styles?

colinrotherham:
## Description

This PR adds support for card and summary list actions as button elements and closes:

* https://github.com/nhsuk/nhsuk-frontend/issues/1988
* https://github.com/nhsuk/nhsuk-frontend/issues/1665

The following new Nunjucks macro options are now supported:

- `item.element` to use `"button"` or `"a"` elements
- `item.id` for the element `id` attribute
- `item.type` for the button `type` attribute
- `item.name` for the button `name` attribute
- `item.value` for the button `value` attribute

Action items using `element: "button"` will be visually styled as links.

```patch
+ <form method="post" novalidate>
    {{ card({
      heading: "Regional Manager",
      actions: {
        items: [
          {
            text: "Delete",
-           href: "/delete"
+           element: "button",
+           name: "action",
+           value: "delete"
          },
          {
            text: "Withdraw",
-           href: "/withdraw",
+           element: "button",
+           name: "action",
+           value: "withdraw"
          }
        ]
```

## Checklist

- [x] CHANGELOG entry


colinrotherham:
@edwardhorsford Internally I've split out a separate `summaryListRow()` component

You'll notice it has both a `params` and a 2nd `parent` parameter

```njk
{% macro summaryListRow(params, parent) %}
```

This is so that each row can access its `parent.params`, `parent.hasActions` etc

```njk
{% call summaryList() %}
  {{ summaryListRow({
    key: {
      text: "Name"
    },
    value: {
      text: "Sarah Philips"
    },
    actions: {
      items: [
        {
          href: "#",
          text: "Change",
          visuallyHiddenText: "name"
        }
      ]
    }
  }) }}

  {{ summaryListRow({
    key: {
      text: "Contact information"
    },
    value: {
      html: "72 Guild Street<br>London<br>SE23 6FH"
    }
  }, {
    hasActions: true
  }) }}
{% endcall %}
```

edwardhorsford:
@colinrotherham love this! great to see call getting supported.

Probably not for now - but it felt awkward that each row needs to know if other rows has actions. It would be much nicer if the 'has-actions' class stuff could just be contained on the parent wrapper.

colinrotherham:
Thanks @edwardhorsford

>Probably not for now - but it felt awkward that each row needs to know if other rows has actions. It would be much nicer if the 'has-actions' class stuff could just be contained on the parent wrapper.

Yeah it's a shame, but `caller()` content can't access its parent context 😬

Similarly if we passed in a context object, [Nunjucks can't update it to set `hasActions`](https://github.com/mozilla/nunjucks/issues/313)

packages/nhsuk-frontend/src/nhsuk/components/card/action/template.njk
```diff
@@ -0,0 +1,38 @@
...
+{#- Set classes for this component #}
+{%- set classNames = "nhsuk-link" -%}
+
+{%- if params.classes %}
+  {% set classNames = classNames ~ " " ~ params.classes %}
+{% endif %}
+
+{%- set attributesHtml %}
+  {{- nhsukAttributes({
+    class: classNames,
```
MatMoore, line 23: could you use the quoted keys here, to make it easier to port to jinja? Or are we not doing that elsewhere?

packages/nhsuk-frontend/src/nhsuk/components/card/action/template.njk
```diff
@@ -7,23 +7,51 @@
...
       value: params.id,
       optional: true
-    },
-    href: {
-      value: params.href if params.href else "#",
-      optional: true
     }
   }) -}}
 
+  {% if element == "button" %}
```
MatMoore, line 30: In order to use buttons for form submission, it seems like we need to externally handle wrapping the button in a `form`. That seems like it would make it more difficult to have more than one button per row. 

Could we add support for a `form` attribute? Or perhaps just allow items to take an extra `attributes` mapping?

I'm also wondering how you would handle the different button variants, like the warning button?

packages/nhsuk-frontend/src/nhsuk/components/summary-list/template.njk
```diff
@@ -1,5 +1,6 @@
 {% from "nhsuk/macros/attributes.njk" import nhsukAttributes %}
 {% from "nhsuk/components/card/macro.njk" import card %}
+{% from "nhsuk/components/card/components/action/macro.njk" import cardAction %}
```
MatMoore: The imports for these subcomponents are a bit unwieldy if we want to use it in application code as well. What do you think about dropping the second `/components` so it's shortened to `nhsuk/components/summary-list/row/macro.njk`?

packages/nhsuk-frontend/src/nhsuk/components/summary-list/template.njk
```diff
@@ -16,46 +16,28 @@
...
-      {% endfor %}
-      </ul>
-      {% endif %}
-    </dd>
-    {% endif %}
-  </div>
-{% endfor %}
+{% if caller %}
+  {{- caller() }}
+{% elif params.html %}
```
MatMoore, line 30: do we need `params.html` as well as `caller()` support or is caller support enough? Is this just for consistency with other components?

packages/nhsuk-frontend/src/nhsuk/components/summary-list/row/macro.njk
```diff
@@ -0,0 +1,3 @@
+{% macro summaryListRow(params, parent) %}
```
MatMoore, line 1: Are we still aiming to support a usage like:

```nunjucks
{% call summaryList(...) }}
  {{ summaryListRow(..., ...) }}
  {{ summaryListRow(..., ...) }}
  {{ summaryListRow(..., ...) }}
{% endcall %}
```

If so, we would have to repeat the `parent` params in each call, and we would also need to make up a value for `parent.loop`.

I feel like it's possible to get a cleaner API by looking at what information you actually need from `parent` and merging it into `params` at the call site.

- `hasActions` could be passed in from the parent component
- the parent component can determine if the row is the last row border, and override the border param accordingly
- the `card` param could be copied to `params.card` of summaryListRow

Either way, I don't think we have quite achieved the proposal from https://github.com/nhsuk/nhsuk-frontend/issues/1665 because setting a variable for a particular row (to be reused amongst multiple summary lists) would erase the context we are currently getting from `loop` about whether it is the last row or not. If you can eliminate the need for that and still apply the correct styles that respect lastRowBorder, then I think this would meet that use case.

packages/nhsuk-frontend/src/nhsuk/components/card/action/template.njk
colinrotherham: Sadly we're only using quoted keys for:

* Hyphened keys such as `"aria-describedby"`
* Non-hyphenated keys when mixed with hyphenated keys (sometimes)

We couldn't reach consensus unfortunately

packages/nhsuk-frontend/src/nhsuk/components/card/action/template.njk
colinrotherham: I'd assumed (like all the form components) that manually wrapping a `<form>` was alright?

Action items already support `classes` and `attributes`

For example, given:

* Manually added `<form>` wrapper
* Set the same `name` attribute `<button name="action">`
* Set a unique `value` attribute `<button name="action" value="delete">`

In the POST body you'd see `{ action: 'delete' }` or whatever the clicked button value was

colinrotherham: >I'm also wondering how you would handle the different button variants, like the warning button?

The action "links" just add the `nhsuk-link` class to the given element

So for `element: "button"` you'd get a button styled as a text link

packages/nhsuk-frontend/src/nhsuk/components/summary-list/template.njk
colinrotherham: Yeah just for consistency

packages/nhsuk-frontend/src/nhsuk/components/summary-list/row/macro.njk
colinrotherham: Yeah for that usage I've included a review app example in https://github.com/nhsuk/nhsuk-frontend/commit/24adfa5f339b83112f43afe0999d1e9b6c1343ee

>* `hasActions` could be passed it in from the parent component

[We already do this](https://github.com/nhsuk/nhsuk-frontend/blob/v10.5.2/packages/nhsuk-frontend/src/nhsuk/components/summary-list/template.njk#L30-L34) ([now passed in](https://github.com/nhsuk/nhsuk-frontend/blob/component-actions/packages/nhsuk-frontend/src/nhsuk/components/summary-list/template.njk#L37)) but it's with `call` blocks that the magic is revealed

For example, how would we solve:

* With only `caller()` available, how does the parent locate the rows?
* With only a call to `summaryListRow()`, how does the row access its parent context?

I chatted to @edwardhorsford earlier and `hasActions` is unnecessary in modern browsers

For example, we can use [`:has()`](https://caniuse.com/css-has) at the parent level to check for rows with actions:

```patch
- .nhsuk-summary-list__row--no-actions {
+ .nhsuk-summary-list__row--no-actions,
+ .nhsuk-summary-list:has(.nhsuk-summary-list__actions)
+   .nhsuk-summary-list__row:not(:has(.nhsuk-summary-list__actions)) {
```

It's a shame that `lastRowBorder` wouldn't work at the parent level

But having `summaryListRow()` gives you direct access to `border: false` at row level

colinrotherham: You can now omit `hasActions: true` when composing the rows manually

It'll be handled automatically in modern browsers via https://github.com/nhsuk/nhsuk-frontend/pull/1989/commits/aa7f63af7fd4bc446c70ecdb38eef32c8ba3e347

I'm still not sure how a row can access `card` or its own `loop.index` from the parent

packages/nhsuk-frontend/src/nhsuk/components/summary-list/template.njk
colinrotherham: Sounds good, I've done this in https://github.com/nhsuk/nhsuk-frontend/pull/1989/commits/20e974d1a151c0c84b15372a89ff4a513899933b

packages/nhsuk-frontend/src/nhsuk/components/card/action/template.njk
MatMoore: I think this is ok then 👍🏻

packages/nhsuk-frontend/src/nhsuk/components/summary-list/row/template.njk
```diff
@@ -0,0 +1,56 @@
+{% from "nhsuk/components/card/action/macro.njk" import cardAction %}
+
+{#- Set classes for this component #}
+{%- set classNames = "nhsuk-summary-list__row" -%}
+
+{%- if params.classes %}
+  {% set classNames = classNames ~ " " ~ params.classes %}
+{% endif %}
+
+{%- set rows = parent.rows | select("mapping") if parent.rows | length else [] %}
```
MatMoore: I think this `rows` is no longer needed now

colinrotherham: Well spotted, thanks

We're still checking these from the parent:

* `parent.params.lastRowBorder` to skip `border: false` on the last row
* `parent.classNames` to also skip `border: false` on the last row
* `parent.loop.last` to check if we're currently looping the last row

packages/nhsuk-frontend/src/nhsuk/components/summary-list/row/macro.njk
colinrotherham: Just an update following a discussion this @edwardhorsford on Slack last week

Setting `lastRowBorder: false` now adds `nhsuk-summary-list--no-last-row-border` at the top level

It's no longer necessary to share border state from parent to child

```njk
{% call summaryList({
  lastRowBorder: false
}) %}
  {{ summaryListRow({
    key: {
      text: "Name"
    },
    value: {
      text: "Sarah Philips"
    },
    actions: {
      items: [
        {
          href: "#",
          text: "Change",
          visuallyHiddenText: "name"
        }
      ]
    }
  }) }}

  {{ summaryListRow({
    key: {
      text: "Contact information"
    },
    value: {
      html: "72 Guild Street<br>London<br>SE23 6FH"
    }
  }) }}
{% endcall %}
```

packages/nhsuk-frontend/src/nhsuk/components/summary-list/row/template.njk
colinrotherham: ☝️ We're now only using these options to detect and prevent duplicate classes being added

Setting `lastRowBorder: false` adds a [new `--no-last-row-border` modifier at parent level only](https://github.com/nhsuk/nhsuk-frontend/pull/1989#discussion_r3529231307)

MatMoore (approved):
👍🏻 Happy to merge when you are happy with it. 

That said, if `parent` is only used to prevent duplicate classes being added, you could move all that logic into the call site in the parent component, and have a single params hash for the row component.

That would make it more intuitive to use externally, but if it's intended to be internal only we don't need to worry about it too much.

colinrotherham:
Thanks @MatMoore

I've just pushed a typo fix in a commit message, no actual changes

>That said, if `parent` is only used to prevent duplicate classes being added, you could move all that logic into the call site in the parent component, and have a single params hash for the row component.

We tend to let components manage their own classes

For example, each component checks boolean flags or variants, and omits classes when already set

Moving the logic would mean we lose this nice compact approach:

```njk
{{- _exampleItem(params, parent, loop.index) -}}
```

Because sadly [we can't merge objects](https://github.com/mozilla/nunjucks/issues/313), parents would have to expand and add every child property:

```njk
{{- _exampleItem({
  id: item.id,
  name: item.name,
  divider: item.divider,
  checked: item.checked,
  hint: item.hint,
  // …
  classes: "nhsuk-example--modifier" if reason
}, parent, loop.index) -}}
```

I reckon we keep it as it is

colinrotherham:
Added fixtures and screenshot tests for both `lastRowBorder: false` and `border: false` on a specific row

PR #1997, opened 2026-07-06
[v10] Preserve conditionally revealed content margins when JavaScript is not supported

colinrotherham:
## Description

This PR preserves our default conditionally revealed content margins when JavaScript is not supported

Previously we'd set smaller margins assuming the content was hidden

Checkboxes with JavaScript disabled showing tighter spacing

## Checklist

- [x] CHANGELOG entry


frankieroberto (approved):
Checked in a browser with JavaScript enabled and disabled, and looks good to me.

PR #2001, opened 2026-07-09
[v10] Add copy icon with code component support

### LINKED ISSUE #469: Copy to share link

colinrotherham:
## Description

This PR adds a copy icon and updates the code component to support icon-only buttons

Code with icon-only copy button

Code with icon-only copy button, reverse

## Checklist

- [ ] CHANGELOG entry


frankieroberto:
@colinrotherham @edwardhorsford I can see that having the option for a smaller icon-based button might be useful for some contexts, but in the "Copy code" examples above (ie for code snippets in documentation) it doesn't really save any space as there's still the same amount of vertical padding added to the top of the code?

colinrotherham:
>@colinrotherham @edwardhorsford I can see that having the option for a smaller icon-based button might be useful for some contexts, but in the "Copy code" examples above (ie for code snippets in documentation) it doesn't really save any space as there's still the same amount of vertical padding added to the top of the code?

If you're happy with the icon I'd like to add an `inline: true` option with the code + button on one line

edwardhorsford:
@colinrotherham not entirely sure what you're suggesting. But if it's about being able to use in a span/with content then yes - I suspect a 'button' might feel too strong on the page and something more subtle would be appropriate.

I reckon after copying we should temporarily swap the icon to a tick - the same way the wordy version swaps to 'Code copied'.

On the icon design: I'd be tempted to try to simplify these, but I defer to @anandamaryon1 on thoughts.

Agree with Frankie that I doubt we'd want to use this on the design system.

colinrotherham:
Like this @edwardhorsford

Screenshot 2026-07-13 at 16 00 02

<br>

Screenshot 2026-07-13 at 16 02 29

PR #2002, opened 2026-07-13
[v10] Add image component `background`, `border` and `width` options

anandamaryon1:
## Description

This PR updates the image component to add new Nunjucks options:

* `background` – Background colour for the image component – `"card"` or `false`
* `border` – If set to `false`, removes the border-bottom from the image component
* `width` – Width of the image component. You can pass any design system grid width here

### Before

Images use `$nhsuk-card-background-colour` by default

Screenshot 2026-07-13 at 17 12 49

### After

Setting Nunjucks `background: false` and `border: false`

Screenshot 2026-07-13 at 17 13 25

## Checklist

- [x] CHANGELOG entry


packages/nhsuk-frontend/src/nhsuk/components/images/macro-options.mjs
```diff
@@ -66,6 +66,13 @@ const options = {
       'A list of image source URLs and their respective sizes. Separate each image with a comma.',
     released: '3.0.0'
   },
+  frameless: {
```
colinrotherham: Thoughts on swapping this for what we do elsewhere?

For example, in the code component you'd set:

* `border: false` to add `nhsuk-code--no-border`
* `variant: "reverse"` to add `nhsuk-code--reverse`

Instead of `frameless`?

```
{{ code({
  html: fixture.html | highlight("html"),
  border: false,
  variant: "reverse"
}) }}
```

anandamaryon1: And reverse would change the background colour from white to transparent? 

Maybe `border: false` and `background: false`?

colinrotherham: We do support `background` in the code component, but it's a string

```
  background: {
    type: 'string',
    required: false,
    description:
      'Optional background colour for the code element – `"body"` or `"template"`.',
    released: '10.4.0'
  },
```

In that case it's optional, but yeah I think `background: false` (or any empty value) would work

Maybe we can save a reverse version for another day

packages/nhsuk-frontend/src/nhsuk/components/images/macro-options.mjs
```diff
@@ -66,11 +66,18 @@ const options = {
...
+  border: {
     type: 'boolean',
     required: false,
     description:
-      'If set to `true`, removes the border and background colour from the image container.',
+      'If set to `false`, removes the border-bottom from the image container.',
+    released: '10.6.0'
+  },
+  background: {
+    type: 'boolean',
```
colinrotherham: Since the default is the card background colour, this should be a `string`

But like other components, overriding this with `background: false` would remove it

packages/nhsuk-frontend/src/nhsuk/components/images/macro-options.mjs
```diff
@@ -66,11 +66,18 @@ const options = {
...
+    required: false,
+    description:
+      'If set to `false`, sets the background colour to transparent for the image container.',
```
colinrotherham: Maybe something like this?

```suggestion
      'Background colour for the image component – `"card"` or `false`. Defaults to `"card"`. To remove the background colour, set `background` to `false`.',
```

packages/nhsuk-frontend/src/nhsuk/components/images/template.njk
```diff
@@ -3,8 +3,12 @@
 {#- Set classes for this component #}
 {%- set classNames = "nhsuk-image" -%}
 
-{%- if params.frameless %}
-  {% set classNames = classNames ~ " nhsuk-image--frameless" %}
+{%- if params.border === false %}
```
colinrotherham: We support both `classes` and `border: true`

So we'll need to check for both here, i.e. to skip the class if already set

```suggestion
{%- if params.border === false and (not params.classes or not "nhsuk-image--no-border" in params.classes) %}
```

packages/nhsuk-frontend/src/nhsuk/components/images/template.njk
```diff
@@ -3,8 +3,12 @@
...
+  {% set classNames = classNames ~ " nhsuk-image--no-border" %}
+{% endif -%}
+
+{%- if params.background === false %}
```
colinrotherham:
```suggestion
{%- if params.background === false and (not params.classes or not "nhsuk-image--no-background" in params.classes) %}
```

packages/nhsuk-frontend/src/nhsuk/components/images/macro-options.mjs
anandamaryon1: Ah ok, will this mean we can add other colours in future, eg. page-background or NHS blue?

colinrotherham: Yeah if necessary, but also because we do this in the code component already 😊

anandamaryon1:
Thanks for the review @colinrotherham I've had a go at amending based on your comments.

frankieroberto:
@anandamaryon1 @colinrotherham do we know what the background colour on the image is even useful for? I wonder if it could just become transparent by default, without needing any options?

colinrotherham:
So you get a [white (card) background on the caption](https://nhsuk-frontend-pr-2002.herokuapp.com/nhsuk-frontend/components/images/default/)?

frankieroberto:
> So you get a [white (card) background on the caption](https://nhsuk-frontend-pr-2002.herokuapp.com/nhsuk-frontend/components/images/default/)?

Ah! Could we move the background colour to the `figcaption` element instead?

colinrotherham:
Yeah could do, then we don't need to worry about a background option?

colinrotherham:
Sorry to hijack this PR @anandamaryon1 but @frankieroberto had a great suggestion above

If we apply the border and background to the `<figcaption>` element only then no changes are necessary

I'm assuming it's a bug that we ever added a 1px bottom border to images without captions anyway?

anandamaryon1:
Ah, I assumed it was to give images a default white background if people drop an image in that has transparency, which then gives it a visible square on our grey background. But admittedly, a total guess.

I'll ask around and see whether anybody knows the reasoning.

anandamaryon1:
Not got a conclusive answer but to avoid a breaking change, let's go with keeping this as adding new background and border options.

We will also add a `width` option to allow widths to be set that can override the built-in width (of 2/3 at desktop and 1/2 on print)

colinrotherham:
@anandamaryon1 Added the `background` and `border` options again

Plus the new image `width` option:

* [**Image** component examples](http://nhsuk-frontend-pr-2002.herokuapp.com/nhsuk-frontend/components/images/)

colinrotherham (approved):

Issue #1554, opened 2025-08-29
Tables have a background colour when hovering over a row


frankieroberto:
Tables currently have a background colour change when you hover over the a row.

However the background colour is the same colour as the regular page colour, so it's only noticeable when the table is on a white background, for instance when it is inside a Tab component or a card.

See examples:

* [Table](https://service-manual.nhs.uk/design-system/components/table) - no background hover effect noticeable
* [Table in Tabs](https://service-manual.nhs.uk/design-system/components/tabs) - background hover effect noticeable

Is this feature intentional, or should it be removed? Normally I'd expect a hover effect to indicate that something is clickable (eg in the Task list or a clickable card), however this doesn't apply for tables.

Possibly it was added to assist readability with dense tables - ie the the hover effect helping users to decipher which heading applies to the row?

colinrotherham:
I've proposed that we add a `--no-hover` modifier and `hover: false` option via

* https://github.com/nhsuk/nhsuk-frontend/pull/1801

anandamaryon1:
I don't know the history to this but it seems to be an error since it's invisible in tables on the standard page background, unexpectedly showing only on non-standard background colours. And as you mention, we tend to use a row hover effect to indicate a row is clickable, like in the task list. 

For this reason, we will remove it.

If anybody has a good reason to add it back in, please let us know and we can discuss adding it back in a considered way, or explore alternatives, eg. zebra striping.

colinrotherham:
Table hover styles were removed in v10.4.2 as part of https://github.com/nhsuk/nhsuk-frontend/pull/1801

PR #2003, opened 2026-07-14
[v10] Add `striped` option for tables

colinrotherham:
## Description

This PR adds support for zebra stripes on tables

Striped table rows

Striped table rows, reverse

## Checklist

- [x] CHANGELOG entry


colinrotherham:
Notice how the table row borders extend beyond their container?

**Note:** This screenshot was taken from a previous push with 4px (not 2px) negative margin

Tables with negative margin

AliMahmoudDev:
This looks solid! I like the approach of keeping changes minimal and focused. Small, targeted PRs like this are much easier to review and less likely to introduce regressions.

AliMahmoudDev:
Striped tables are a classic readability enhancement! One thing to consider: the stripes should have enough contrast to be visible but not so strong that they're distracting. NHS design standards would have specific guidance on this.

frankieroberto:
@colinrotherham interesting! I did a quick search and found this backlog issue relating to zebra striping: https://github.com/nhsuk/nhsuk-service-manual-community-backlog/issues/465

I’ve not read any particular research on if it helps usability in some contexts, but maybe others have (feels like the kind of thing people have done studies on?).

We previously had a row hover background colour change, which might serve a similar purpose (help users follow very wide rows?) but that was removed as it was only apparently on tables on white backgrounds: https://github.com/nhsuk/nhsuk-frontend/issues/1554

One thought though: if you have striping on tables, does that remove the need to have horizontal borders between rows? 🤔 The border somehow looks less visible with the striping, as it has less contrast with one side.

ie could you do this?

Screenshot 2026-07-14 at 23 00 02

or the darker grey (but maybe too dark?):

Screenshot 2026-07-14 at 23 01 35

colinrotherham:
@frankieroberto Thanks!

Ahh we have to keep it lighter otherwise striped rows would be more obvious than sortable table headers on hover

Try them next to each other?

But we could shade the border colour (slightly darker) on every striped row to compensate?

Mainly after your thoughts on margin and padding adjustments 😎

Demotive:
Just a quick one to say that we'll want to use these!
One example to check out might be default-striped-in-a-card (screenshot attached)
Screenshot 2026-07-15 at 10 30 44

Demotive:
Personal preference is I _like_ the borders, aesthetically.

frankieroberto:
@Demotive ooh good point, the stripes have way more contrast when on a white background (and look a bit better to me).

frankieroberto:
@colinrotherham:

> Mainly after your thoughts on margin and padding adjustments 😎

Split the difference and make it 2px?

Is it also worth considering only adding the padding adjustments on tables that need it (ie sortable + striped) rather than all of them?

Another option might be to add negative 2px margin on either side of the table so that it goes into the gutter slightly and the text still lines up with the rest of the page.

colinrotherham:
>Is it also worth considering only adding the padding adjustments on tables that need it (ie sortable + striped) rather than all of them?

Maybe. The text will line up, but row borders will visibly extend 2px further into the gutter (for sortable + striped)

>Another option might be to add negative 2px margin on either side of the table so that it goes into the gutter slightly and the text still lines up with the rest of the page.

Yeah that's what's in this PR (at 4px) 🙌

colinrotherham:
>Split the difference and make it 2px?

@frankieroberto 4px now changed to 2px

Try these examples for both striped and sortable:

* [**Tables striped with numeric data, sortable**](https://nhsuk-frontend-pr-2003.herokuapp.com/nhsuk-frontend/components/tables/striped-with-numeric-data-sortable/)
* [**Tables reverse striped with numeric data, sortable**](https://nhsuk-frontend-pr-2003.herokuapp.com/nhsuk-frontend/components/tables/reverse-striped-with-numeric-data-sortable/)

>Just a quick one to say that we'll want to use these!
>One example to check out might be default-striped-in-a-card (screenshot attached)

@Demotive Table as a card examples added. For comparison:

* [Table as a card](https://nhsuk-frontend-pr-2003.herokuapp.com/nhsuk-frontend/components/tables/as-a-card/)
* [**Table striped as a card**](https://nhsuk-frontend-pr-2003.herokuapp.com/nhsuk-frontend/components/tables/striped-as-a-card/) ← New
* [Table as a card (feature)](https://nhsuk-frontend-pr-2003.herokuapp.com/nhsuk-frontend/components/tables/as-a-card-feature/)
* [**Table striped as a card (feature)**](https://nhsuk-frontend-pr-2003.herokuapp.com/nhsuk-frontend/components/tables/striped-as-a-card-feature/) ← New
* [Tables as a card (feature) responsive](https://nhsuk-frontend-pr-2003.herokuapp.com/nhsuk-frontend/components/tables/as-a-card-feature-responsive/)
* [**Tables striped as a card (feature) responsive**](https://nhsuk-frontend-pr-2003.herokuapp.com/nhsuk-frontend/components/tables/striped-as-a-card-feature-responsive/) ← New

anandamaryon1:
Thanks for this, looks good to me! Two of the things I would have commented on (lessen the 4px padding to 2px, and use a semitransparent shade colour) have been done already 👍 

I had a look for research on effectiveness and the best I can find is this: https://alistapart.com/article/zebrastripingmoredataforthecase Which suggest they can be useful, but for data tables. I think we'll want to pair this with guidance that they should be used on wide data tables only. 

One question, zebra striping on responsive view tables: 
image

This looks odd and unnecessary to me. It no longer serves its purpose of helping a user identify and scan a long horizontal line of information, so I think therefore it should not be shown in the stacked view.

colinrotherham:
>This looks odd and unnecessary to me. It no longer serves its purpose of helping a user identify and scan a long horizontal line of information, so I think therefore it should not be shown in the stacked view.

@anandamaryon1 That gap does look a little odd

I've removed the stripes from responsive tables below desktop

colinrotherham:
This PR is now solely for striped tables

Changes to table width (using negative margins) and cell padding are now in:

* https://github.com/nhsuk/nhsuk-frontend/pull/1654

frankieroberto (approved):
Nice!

PR #2012, opened 2026-07-24
Experimental: add auto-tabbing behaviour to date inputs

frankieroberto:
This is an attempt to add auto-tabbing to the date input components in a way that doesn't introduce additional issues.

It aims to solve the core problem, observed in some user research sessions, where:

1. users expect auto-tabbing but spot that it doesn't happen, and find it mildly annoying to have to press tab or click the second and third inputs (impact: low)
2. users expect auto-tabbing and try trying 3 digits into the Day box, expecting it to auto-tab. This can then take a short while to spot and correct (impact: medium)

Auto-tabbing aims to solve this by:

1. auto tabbing between day and month, and month and year after you enter the 2nd character
2. if you weren't expecting auto-tabbing, and enter 2 digits and then press tab with 500ms, the tab keypress is suppressed to avoid accidental tabbing to the next field
3. if you weren't expecting auto-tabbing, and enter 2 digits but then realise you've made a mistake and press backspace within 500ms, then the backspace is applied to the previous input instead
4. after 500ms, or after pressing any key, the tab-suppression and backspace behaviour is turned off, and normal behaviour applies

To additional prevent auto-tabbing mis-firing when you're not expecting it, auto-tabbing does not happen if:

* the next field is disabled
* the next field is not blank (in this case, it's assumed you're editing a field)
* the date field components are not day, month or year but something custom like 'week number' or 'quarter' or 'financial year'

The auto-tabbing does also not currently apply if:

* you enter a single digit in the month field which is between 4 and 9 - although it could, as there are no 2 digit months which start with those


colinrotherham:
Nice to see a demo of this. I've fixed the Nunjucks and TypeScript issue and rebased

I've always wondered if we should support:

* Copying and pasting a date string
* Moving forward as inputs are filled (this PR)
* Moving backwards and inputs are emptied

Whilst only related for authentication purposes, preventing **copy and paste** is discussed here:

>Understanding [SC 3.3.8](https://w3.org/TR/WCAG22#accessible-authentication-minimum)
>[**Accessible Authentication (Minimum) (Level AA)**](https://www.w3.org/WAI/WCAG22/Understanding/accessible-authentication-minimum.html)

accessible-authentication-totp-multiple-fields

edwardhorsford:
Autotabbing is an obvious thing for 2FA code entry, though I've long wondered if a better approach for that specific scenario is to have a single text input behind the scenes and treat the visual treatment of 6 separate boxes as entirely a visual enhancement.

PR #2036, opened 2026-08-07
[v10] Use flat indicator style for scrolling tables

colinrotherham:
## Description

Quick PR to try changing the scrolling table indicators from shadows to flat overlays

**For context:** Buttons, clickable cards and images use flat colours as shadows

Full width table with flat shadows

Full width table (reverse) with flat shadows

## Checklist

- [ ] CHANGELOG entry


anandamaryon1:
Nice compromise, when I tried the flat shadow I had only one thin line but the double shadow improves how noticeable it is, particularly when flush the the edge of a screen, I like it thanks!

I've pushed a slight tweak to make the lighter shadow a bit less wide, and to make the transition smoother when the shadow is applied, to draw your eye less to the shadow being animated in, but then still fast to disappear.

anandamaryon1 (approved):

PR #2067, opened 2026-08-24
Remove redundant landmark roles from HTML5 elements

olitreadwell:
## Description

The header, footer, contents list and pagination components set explicit role attributes on native HTML5 elements that already map to those roles. The attributes (`banner`, `contentinfo`, `navigation`) were added as an Internet Explorer workaround and are redundant in every browser the design system supports.

Removed:

- `role="banner"` from the header component's `<header>` element
- `role="contentinfo"` from the footer component's `<footer>` element
- `role="navigation"` from the contents list and pagination components' `<nav>` elements

The matching allowances in the review app's HTML validation config are removed too, so the attributes can't come back unnoticed.

Left in place: the card `role="text"`, the do and don't list `role="list"` and the responsive table roles. The first two are VoiceOver workarounds that still need research; the table roles are part of the responsive table accessibility discussion.

One thing to know before reviewing: per HTML-AAM, `header` and `footer` only map to `banner` and `contentinfo` when they are not nested inside `section`, `article`, `main`, `aside` or `nav`. On real pages the header and footer are top-level, so the landmarks are unchanged. In the review app's example galleries the components are wrapped in sections, so they now get section-scoped roles instead of the forced landmarks. That matches the spec.

## Checklist

- [x] CHANGELOG entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1515,6 +1515,10 @@ This change was introduced in [pull request #2002: Add image component `backgrou
 - [#2048: Fix margin and line height issues for labels and legends as page headings](https://github.com/nhsuk/nhsuk-frontend/pull/2048)
 - [#2049: Fix radio button and check mark alignment on Windows 125% scaling](https://github.com/nhsuk/nhsuk-frontend/pull/2049)
 
+### :wrench: **Fixes**
+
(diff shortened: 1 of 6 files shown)


colinrotherham:
Closing this whilst we consider our approach to contributions from AI bots in https://github.com/nhsuk/nhsuk-frontend/issues/2079

PR #2071, opened 2026-08-25
[WIP] Add Outline modifiers to checkboxes and radios

brianmchugh-nhs:
## Description

This PR extends the checkboxes and radios adding an `Outline` modifier to apply inline button-style controls, matching NHS button patterns.

## Checklist

- [ ] CHANGELOG entry


colinrotherham:
Sorry wrong button. Approved the workflow run again

colinrotherham:
If you'd like to run the screenshot tests for checkboxes (whilst http://localhost:3000 is available)

```console
npm run test:visual --ignore-scripts -- --filter="Checkboxes"
```

Then to approve the new or updated screenshots:

```console
npm run test:visual:approve
```

packages/nhsuk-frontend/src/nhsuk/components/checkboxes/_outline.scss
```diff
@@ -0,0 +1,235 @@
...
+///
+/// Styles standard nhsuk-checkboxes as inline button-style controls,
+/// matching NHS button interaction patterns.
+////
+
+.nhsuk-checkboxes.nhsuk-checkboxes--outline {
+  display: flex;
+  flex-wrap: wrap;
+  column-gap: nhsuk-spacing(2);
+  row-gap: nhsuk-spacing(2) + $nhsuk-border-width-form-element;
```
colinrotherham, line 14: For reference, we only got [`column-gap` in Safari v14.1](https://caniuse.com/mdn-css_properties_column-gap_flex_context)

For our supported browsers on NHS.UK frontend we'd need to find an alternative (or check it looks alright without). E.g. We're using `margin-right: nhsuk-spacing(4);` for the `--inline` modifier

You'll see we use `row-gap` already but it's an enhancement rather than mandatory

packages/nhsuk-frontend/src/nhsuk/components/checkboxes/_outline.scss
```diff
@@ -0,0 +1,235 @@
...
+    &::before {
+      bottom: -($nhsuk-border-width-form-element + $nhsuk-button-shadow-size);
+      border: $nhsuk-border-width-form-element solid transparent;
+      background-color: transparent;
+    }
+
+    &:active::before {
+      top: -($nhsuk-border-width-form-element + $nhsuk-button-shadow-size);
+      bottom: -$nhsuk-border-width-form-element;
+    }
```
colinrotherham, line 89: These are copied as-is from buttons

But we've missed the high contrast styles

It's worth setting up a Sass `nhsuk-button-style-default()` mixin for them

```scss
    // Override high-contrast link colours to match buttons
    @media screen and (forced-colors: active), (-ms-high-contrast: active) {
      color: ButtonText;

      &,
      &::before {
        border-color: ButtonBorder;
      }

      &:not(:focus):not(:active):hover,
      &:not(:focus):not(:active):hover::before,
      &:not(:focus):not(:active):hover::after {
        border-color: Highlight;
        color: Highlight;
      }
    }
```

packages/nhsuk-frontend/src/nhsuk/components/checkboxes/_outline.scss
```diff
@@ -0,0 +1,235 @@
+@use "../../core" as *;
+
+////
+/// Checkboxes outline modifier
+///
+/// Styles standard nhsuk-checkboxes as inline button-style controls,
+/// matching NHS button interaction patterns.
+////
+
+.nhsuk-checkboxes.nhsuk-checkboxes--outline {
```
colinrotherham, line 10: You'll see us wrap our main component styles in `nhsuk-exports()`

This is to log whether Sass `@import` has already imported it (doesn't affect Sass modules)

Teams can accidentally import styles twice, so this stops the output being duplicated

```suggestion
@include nhsuk-exports("nhsuk/components/checkboxes/outline") {
  .nhsuk-checkboxes.nhsuk-checkboxes--outline {
```

packages/nhsuk-frontend/src/nhsuk/components/checkboxes/_outline.scss
```diff
@@ -0,0 +1,235 @@
...
+
+    border-radius: nhsuk-px-to-rem($nhsuk-button-border-radius);
+
+    color: $nhsuk-secondary-button-text-colour;
+
+    font-weight: $nhsuk-font-weight-bold;
+    line-height: 1.25;
+    user-select: none;
+
+    gap: nhsuk-spacing(2);
```
colinrotherham: Can we stick with `column-gap` over `gap`?

Support for plain old `gap` (for flex box) took a bit longer for full support

**Edit:** Although bear in mind we'd consider this an enhancement that older browsers won't apply

packages/nhsuk-frontend/src/nhsuk/components/checkboxes/_outline.scss
```diff
@@ -0,0 +1,235 @@
...
+
+    opacity: 1;
+
+    background-color: transparent;
+
+    cursor: pointer;
+
+    appearance: none;
+
+    inset: 0;
```
colinrotherham, line 56: Same with this one really, we'd need top/right/bottom/left separately for browser support

packages/nhsuk-frontend/src/nhsuk/components/radios/_outline.scss
```diff
@@ -0,0 +1,237 @@
+@use "../../core" as *;
+
+////
+/// Radios outline modifier
+///
+/// Styles standard nhsuk-radios as inline button-style controls,
+/// matching NHS button patterns.
+////
+
+.nhsuk-radios.nhsuk-radios--outline {
```
colinrotherham, line 10: Can we figure out which rules need the extra specificity here?

It probably won't be many. Seems a shame to chain extra classes for the CSS weight

```suggestion
.nhsuk-radios--outline {
```

colinrotherham:
Is this an existing bug with conditional content?

Notice the lack of text input bottom margin (good) here:

Checkboxes_outline_with_conditional_content_error_message_nested_focus_desktop

<br>
<br>

But on some of the tablet screenshots, the red form-group left border continues down (bad) further:

Checkboxes_outline_with_conditional_content_error_message_focus_tablet

packages/nhsuk-frontend/src/nhsuk/components/checkboxes/_outline.scss
colinrotherham: What did you want to do about `column-gap`, leave it a bit squashed?

Appreciate we won't be actively supporting older browsers on Rubie

packages/nhsuk-frontend/src/nhsuk/components/checkboxes/_outline.scss
colinrotherham: Shall we still swap `inset: 0` with the separate properties?

```patch
- inset: 0;
+ top: 0;
+ right: 0;
+ bottom: 0;
+ left: 0;
```

packages/nhsuk-frontend/src/nhsuk/components/checkboxes/_outline.scss
```diff
@@ -0,0 +1,292 @@
...
+
+$nhsuk-checkboxes-outline-row-gap: nhsuk-spacing(2) + $nhsuk-border-width-form-element + $nhsuk-button-shadow-size;
+
+@include nhsuk-exports("nhsuk/components/checkboxes/outline") {
+  .nhsuk-checkboxes--outline {
+    display: flex;
+    flex-wrap: wrap;
+    row-gap: $nhsuk-checkboxes-outline-row-gap;
+    margin: 0;
+    margin-bottom: $nhsuk-button-shadow-size;
```
colinrotherham, line 18: You reckon we need to apply the outline styles 2nd, not first?

I'm assuming these margins immediately get overridden by the default checkbox styles

Same for radios

packages/nhsuk-frontend/src/nhsuk/components/checkboxes/_outline.scss
brianmchugh-nhs: Sorry this was changed and overwritten. Just checking this now.

packages/nhsuk-frontend/src/nhsuk/components/checkboxes/_outline.scss
```diff
@@ -0,0 +1,292 @@
...
+
+    &.nhsuk-checkboxes {
+      margin: 0;
+      margin-bottom: $nhsuk-button-shadow-size;
+    }
```
colinrotherham: Oh yeah, you've had to put them here too 😅

Making sure the styles are ordered default → outline would fix that?

colinrotherham: E.g. Like this in **packages/nhsuk-frontend/src/nhsuk/components/_index.scss**

```patch
  // Form components
  @forward "checkboxes";
+ @forward "checkboxes/outline";
```

Then we get to see what it'll be like in Rubie (with the overrides hitting last, not first)

Cheers

packages/nhsuk-frontend/src/nhsuk/components/checkboxes/fixtures.mjs
```diff
@@ -1114,6 +1114,196 @@ const fixtures = {
...
+        }
+      },
+      items: [
+        { value: "calc", text: "Calcification" },
+        { value: "asym", text: "Asymmetric density" },
+        {
+          value: "other",
+          text: "Other",
+          conditional: {
+            html: '<div class="nhsuk-form-group"><label class="nhsuk-label" for="other-details">Provide details</label><input class="nhsuk-input nhsuk-u-width-two-thirds" id="other-details" name="other-details" type="text"></div>'
```
colinrotherham: Mind using `components.render("input", { …` like you've done in other examples?

Just makes sure this HTML doesn't drift out of date as we make changes to the input component

packages/nhsuk-frontend/src/nhsuk/components/checkboxes/_outline.scss
```diff
@@ -0,0 +1,292 @@
...
+        left: nhsuk-px-to-rem(16px);
+        width: nhsuk-px-to-rem(10px);
+        height: nhsuk-px-to-rem(5px);
+        border-width: 0 0 $nhsuk-border-width-form-element $nhsuk-border-width-form-element;
+      }
+
+      .nhsuk-checkboxes__conditional {
+        margin-left: nhsuk-spacing(1);
+        padding-left: nhsuk-px-to-rem(20px);
+      }
```
colinrotherham: Once we've flipped the CSS order, I think all of these can go?

Same for radios

```suggestion
```

packages/nhsuk-frontend/src/nhsuk/components/checkboxes/_outline.scss
brianmchugh-nhs: Fixed

packages/nhsuk-frontend/src/nhsuk/components/radios/_outline.scss
```diff
@@ -0,0 +1,297 @@
...
+
+      &:active::before {
+        top: -($nhsuk-border-width-form-element + $nhsuk-button-shadow-size);
+        bottom: -$nhsuk-border-width-form-element;
+      }
+
+      // Override high-contrast link colours to match buttons
+      @media screen and (forced-colors: active), (-ms-high-contrast: active) {
+        border-color: ButtonBorder;
+        color: ButtonText;
```
colinrotherham: Looks like these two are already coming from `@include nhsuk-button-style()`

```suggestion
```

<br>

Can we set that `color: ButtonText;` on the label instead though?

It's currently got `color: #005eb8` but `forced-color-adjust` is (thankfully) fixing it in Edge

I'll check Firefox but it probably looks like this:

Label high contrast colour not set

colinrotherham:
Thanks @brianmchugh-nhs it's looking great

Checkboxes outline

Radios outline

<br>
<br>

Do you reckon we need to keep some sense of the bold blue checked state?

Perhaps with `background-color: HighlightText` on the label?

```css
  color: CanvasText;
  background-color: HighlightText;
  forced-color-adjust: none;
```

Radios outline with background colour when selected

packages/nhsuk-frontend/src/nhsuk/components/radios/_outline.scss
```diff
@@ -0,0 +1,297 @@
...
+      @media screen and (forced-colors: active), (-ms-high-contrast: active) {
+        &::after {
+          forced-color-adjust: none;
+        }
+      }
+    }
+
+    .nhsuk-radios__input:checked + .nhsuk-radios__label {
+      color: nhsuk-colour("white");
+      background-color: nhsuk-colour("blue");
```
colinrotherham, line 261: These two colours aren't reset back to [system colours](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/system-color) in high contrast mode

They're automatically being forced in latest Chrome/Edge/Firefox but in older versions (Firefox especially) this wasn't so reliable

Same as https://github.com/nhsuk/nhsuk-frontend/pull/2071#discussion_r4060196161 really, if we can find every colour and double check they're put back?

packages/nhsuk-frontend/src/nhsuk/components/checkboxes/_outline.scss
```diff
@@ -0,0 +1,290 @@
...
+      line-height: 1.25;
+      user-select: none;
+      @include nhsuk-font-size(16);
+
+      &::before {
+        content: "";
+
+        position: static;
+
+        flex: 0 0 nhsuk-px-to-rem(18px);
```
colinrotherham: Firefox still appears to honour the `width: 2.5rem` despite the new `flex-basis`

You can see how it's reserved too much space (right hand side)

Image

<br>
<br>

Overriding the width instead seems to work

```suggestion
        width: nhsuk-px-to-rem(18px);
```

packages/nhsuk-frontend/src/nhsuk/components/radios/_outline.scss
colinrotherham: Here's another example

High contrast mode is currently ignoring this `box-shadow` colour, but one day it might pick an alternative colour on our behalf so we're better setting our own system colour

Custom colours not overridden to system colours

packages/nhsuk-frontend/src/nhsuk/components/radios/_outline.scss
colinrotherham: False alarm, it doesn't look like that in Firefox

It's ignoring the label colour because contrast is too low

PR #2086, opened 2026-09-08
Card: Remove top margin on care and feature variants

paulrobertlloyd:
## Description

Card’s don’t have a top margin, only a bottom margin like most other components.

However, care cards and feature cards do. Care cards have the same overall metrics as normal cards, so shouldn’t need a top margin.

Feature cards have an offset heading, so a top margin _is_ needed, but only to account for this offset.

Right now, without this change, you need to remove the top margins when these two card variants the first child. For example, if they are sat within the top of a grid column, which itself has top spacing, you end up with extra space where there shouldn’t be.

You can see such wrangling in https://github.com/NHSDigital/manage-vaccinations-in-schools-prototype/blob/cb7e94df279945433595ba550db1baf44b235209/app/assets/stylesheets/core/objects/_grid.scss#L30

## Checklist

- [ ] CHANGELOG entry


colinrotherham (commented):
Makes sense to me 🙌

Thoughts @anandamaryon1?

Probably needs a changelog entry under **Changes** (non-breaking)

anandamaryon1:
Yep makes sense. 

I assume the top-margins were added to give some extra space between care cards and warning callouts and a preceding paragraph, on content pages. They're used very often on nhs.uk so my concern is how they'd adapt to this change.

colinrotherham:
@anandamaryon1 Can we set up some full page examples showing the typical setup?

If you can add some links in here for now that would be great (thanks)

anandamaryon1:
Sure, scroll down these pages: 
* https://www.nhs.uk/medicines/aspirin/
* https://www.nhs.uk/conditions/anaphylaxis/
* https://www.nhs.uk/conditions/migraine/

Also applies to do don't lists.

colinrotherham:
@anandamaryon1 @paulrobertlloyd I've added the Health and Medicines A to Z examples

* [Health A to Z](https://nhsuk-frontend-pr-2086.herokuapp.com/nhsuk-frontend/examples/full-page/health-a-to-z/)
* [Medicines A to Z](https://nhsuk-frontend-pr-2086.herokuapp.com/nhsuk-frontend/examples/full-page/medicines/)
* [Medicines A to Z: Aspirin](https://nhsuk-frontend-pr-2086.herokuapp.com/nhsuk-frontend/examples/full-page/medicines/aspirin/)

Could do with the Conditions A to Z ones too

> I assume the top-margins were added to give some extra space between care cards and warning callouts and a preceding paragraph, on content pages. They're used very often on nhs.uk so my concern is how they'd adapt to this change.

Unfortunately yes, styles for `.nhsuk-back-to-top` are missing from www.nhs.uk so the card top margin is currently necessary to add some margin back again.

I've swapped out `.nhsuk-back-to-top` for a paragraph in the examples above:

```patch
- <div class="nhsuk-back-to-top">
-   <a class="nhsuk-back-to-top__link" href="#nhsuk-nav-a-z" data-nhsappnohistory="">
+ <p class="nhsuk-body-s">
+   <a href="#nhsuk-nav-a-z">
      Back to top
    </a>
- </div>
+ </p>
```

PR #2098, opened 2026-09-16
Replace deprecated command in publish workflow

frankieroberto:
`::set-output` was [deprecated](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/) in 2022, although [postponed from removal](https://github.blog/changelog/2023-07-24-github-actions-update-on-save-state-and-set-output-commands/) (indefinitely?) later that year.

It's no longer mentioned in the docs though, so seems like a good idea to switch to the `>> $GITHUB_OUTPUT` pattern instead. (See [docs](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#setting-an-output-parameter))

(Spotted whilst working on https://github.com/nhsuk/nhsapp-frontend/pull/590)


colinrotherham (commented):
Thanks, it's been on my to-do list every time I see it in the output:
https://github.com/nhsuk/nhsuk-frontend/actions/runs/32865053846

Any thoughts on the Node.js deprecation warning too?

frankieroberto:
@colinrotherham ah yeah I also spotted that the `actions/create-release` and `actions/upload-release-asset@v1` actions have been archived and are no longer maintained.

Wanna copy what `nhsapp-frontend` does and:

* change the process so that you create the release in the GitHub UI, and then that triggers the publish action, rather than the publish action creating the release based on a tag being pushed
* change from `actions/upload-release-asset` to use the GitHub CLI tool `gh release upload`

